PR #7, opened 2026-07-25
Add editor block types, find and replace, zoom, and deep links

westerosweb:
Implements a batch of self-contained editor and shell improvements, each covered by unit tests (106 new tests, 143 total passing).

Editor
- Mermaid diagram and LaTeX math blocks, durable in markdown via ```mermaid fences and $$ fences. Math needs a pre-pass because cmark parses $$ as a paragraph and collapses internal newlines.
- ==highlight== inline mark (Cmd+Shift+M), shared between the editor and the serializer so the two cannot drift.
- Find and replace (Cmd+F), applying ranges last-to-first so offsets stay valid when the replacement length differs.
- Move blocks (Cmd+Shift+Up/Down). Non-contiguous selections are rejected rather than reordered arbitrarily.
- Per-document width (normal/wide) with a toolbar toggle.
- Editor zoom (Cmd +/-/0) as a multiplier over the base font-size preference, clamped in one place.
- Paste without formatting (Cmd+Shift+V).

Shell
- logue:// deep links for documents, meetings, and spaces. Parsing is side-effect free and rejects unknown hosts, malformed UUIDs, traversal paths, and extra components; routing is a separate step.
- Sanitized diagnostics attached to bug reports. Excludes filesystem paths, account name, URLs, and API keys; reports external providers as a count only. Every value is stripped of control characters so a malformed model name cannot forge diagnostic lines.

Notes
- WritingDocument gains optional backing storage for new fields. Swift's synthesized Codable throws keyNotFound for a missing non-optional key even when a default exists, so a non-optional property would have made every existing document undecodable.
- DocumentIcon validates against Unicode category Cc only, not CharacterSet.controlCharacters, which also covers Cf and would reject the zero-width joiner in multi-scalar emoji.
- syncMarkdownFromBlocks moved to BlockEditorView+Reorder to stay under the 800-line file limit without a lint suppression.

Not yet surfaced in the UI (models built and tested only): document icons, layout modes, quick-open palette mode.

## How was this verified?

- [ ] Builds cleanly (`xcodebuild build -project Logue.xcodeproj -scheme Logue -destination 'platform=macOS'`)
- [ ] Lints cleanly (`make lint`) — no new `swiftlint:disable` without justification
- [ ] Added/updated tests where it makes sense
- [ ] Ran `xcodegen generate` if I added/removed `.swift` files or changed `project.yml`

## Checklist

- [ ] My change keeps all AI inference and data processing **on-device** (no new network calls to third-party services without discussion).
- [ ] I did not commit secrets, API keys, or personal data.
- [ ] I read [CONTRIBUTING.md](../CONTRIBUTING.md) and followed the security/concurrency standards.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ memory/
 # Python
 __pycache__/
 *.pyc
+
+# Internal planning docs (not for public repo)
+docs/CAPABILITY-ROADMAP.md
--- a/Logue.xcodeproj/project.pbxproj
+++ b/Logue.xcodeproj/project.pbxproj
@@ -52,6 +52,7 @@
 		127BFC4FD2D2CA4E68169EC5 /* BackupSettingsTab.swift in Sources */ = {isa = PBXBuildFile; fileRef = 431C845CF5A3DC729ED17CC0 /* BackupSettingsTab.swift */; };
 		12AD1B6E3CB248360845B684 /* WritingGoalsPanelView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21ED60A38CACB407A6BEDD62 /* WritingGoalsPanelView.swift */; };
 		13373229911D8C07B23E3591 /* GrammarClarityLLMTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E20D8AB097118E507520863A /* GrammarClarityLLMTests.swift */; };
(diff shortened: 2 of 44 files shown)


shanforge (approved):
LGTM thanks!

PR #23, opened 2026-07-27
fix: disable App Sandbox so Logue appears in Accessibility settings (#22)

### LINKED ISSUE #22: Logue doesn't show in accessibility
### What happened?

When setup, permissions are granted, but the accessibility fails. This is because the app is not displayed in accessibility. All other permissions work.

### Steps to reproduce

Launch setup, grant permissions.

### Logue version

1.0.0.

### macOS version

macOS 27

### Mac model

MacBook Pro m4

### Screenshots or logs

_No response_

westerosweb:
Fixes #22.

## Root cause

Release builds shipped with `com.apple.security.app-sandbox = true`. **A sandboxed process cannot be an Accessibility API client** — there is no entitlement that grants it. `AXIsProcessTrustedWithOptions(prompt: true)` in `AccessibilityService.checkAccessibilityPermission` returns `false`, no prompt is shown, and TCC never creates an entry. So Logue never appears in System Settings → Privacy & Security → Accessibility, and the user cannot add it manually either.

Mic / Speech / Calendar were unaffected because those *are* sandbox-compatible, which is why the reporter saw only Accessibility fail. Debug builds use `Logue.entitlements` with the sandbox **off**, so this never reproduced in development. `git log -S` confirms the sandbox has been on since the initial commit — no release has ever had working cross-app features.

**Verified on the shipped 1.0.0** in `/Applications`:
```
$ codesign -d --entitlements - /Applications/Logue.app
    [Key] com.apple.security.app-sandbox
        [Bool] true
```

### Blast radius is wider than the issue reports

The sandbox silently disabled everything else that needs the same trust:

| Broken | Where |
| --- | --- |
| ⌘⌃I inline assistant, global hotkeys | `NSEvent.addGlobalMonitorForEvents` in `ShortcutManager.swift:52`, `AppDelegate.swift:420`, `MenuBarCompanion.swift:28` |
| Command Center | `CommandCenterController.swift:385` |
| Text replacement / insertion | synthetic `CGEvent` Cmd+C / Cmd+V, `AccessibilityService.swift:369-423` |

## Fix

Logue is distributed Developer ID + notarized via GitHub Releases and Sparkle — **not the Mac App Store** — so the sandbox is not required. **Hardened Runtime stays ON** and every exception remains individually justified.

- **`Logue.release.entitlements`** — `app-sandbox` → `false`, with a comment explaining why it must stay off. Drops two keys that are inert without the sandbox: `files.user-selected.read-write` and the Sparkle `mach-lookup` temporary exception.
- **`project.yml`** — drops `SUEnableInstallerLauncherService`.

### Sparkle / auto-update is unaffected

Per [Sparkle's sandboxing guide](https://sparkle-project.org/documentation/sandboxing/), the XPC services and `SUEnableInstallerLauncherService` exist *only* to work around the sandbox — "if you do not sandbox your application, you should skip this guide." Unsandboxed Sparkle installs directly via `Autoupdate`/`Updater.app`.

Critically, the 1.0.0 → next upgrade is performed by the **already-installed sandboxed 1.0.0** using its own `Installer.xpc`. The new bundle's entitlements have no bearing on that, so the in-flight update path is untouched.

### Data migration — the part that would have bitten us

Unsandboxing re-points `URL.applicationSupportDirectory`, `NSHomeDirectory()` and `UserDefaults` from the container at the real home directory. Shipping the entitlement change alone would have made every upgrading user's data vanish. Measured in the container on a real 1.0.0 install:

| Path | Size |
| --- | --- |
| `Application Support/Logue` (meetings, documents, templates, spaces) | 464K |
| `Application Support/com.bitwize.logue` (vector store, memory index) | 372K |
| `Application Support/FluidAudio` (diarization models) | 697M |
| `.localllmclient` (MLX models) | 2.1G |
| `.cache/huggingface` | 2.1G |

New `SandboxContainerMigrator` runs once in `LogueApp.init()` — before any store resolves its directory — and:

- no-ops on a clean install, once migrated, and if somehow running sandboxed
- moves same-volume, so it is a rename: instant, and needs no extra free space for the ~3 GB of models
- **never overwrites an existing destination** (matters on machines that ran both a release and a local Debug build)
- merges the container's preference domain without clobbering existing keys, skipping `com.apple.*`
- logs every step through `do/catch`, no silent `try?`

Keychain is unaffected: `KeychainHelper` uses the legacy file keychain, whose ACLs are bound to the signing identity, which does not change.

## Regression guards

Nothing fails at build time if someone flips the sandbox back on, so this is checked explicitly in two places:

- `LogueTests/ReleaseEntitlementsTests.swift` (4 tests)
- a CI step in the lint job, since CI does not run tests

**Both were confirmed to actually fail** when `app-sandbox` is set back to `true`:
```
✘ Test "Release is not sandboxed, so it can be an Accessibility API client" failed:
  Expectation failed: (sandbox → true) == false
::error file=Logue/Resources/Logue.release.entitlements::app-sandbox is 'true', expected 'false'
```

## Verification

- `xcodebuild build -configuration Release` → **BUILD SUCCEEDED**
- Signed with the release entitlements → `app-sandbox = false`, `mach-lookup` gone
- `SUEnableInstallerLauncherService` absent from the built `Info.plist`; `SUFeedURL` intact
- `swiftformat --lint` (0/380) and `swiftlint --strict` (0 violations in 408 files) both clean
- 14 tests pass across `ReleaseEntitlementsTests` + `DocumentReplaceTests`

**Not verified end-to-end:** the actual TCC grant on a notarized build — that needs a real signed release. Worth confirming on the first 1.0.1 build before closing the issue.

## Unrelated drive-by

The test target did not compile: `DocumentReplaceTests.swift:158` broke when `textContent` became `String?` in 1e83485, plus a `docComments` lint violation in the same file. CI runs swiftformat only inside `Logue/` and does not run tests, so both went unnoticed. Fixed so the suite runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,25 @@ jobs:
       - name: Run SwiftLint
         run: swiftlint lint --strict Logue/
 
+      # Logue must not be sandboxed: a sandboxed process cannot be an Accessibility
+      # API client, so macOS never lists it under Privacy & Security → Accessibility
+      # and every Logue/CrossApp feature (⌘⌃I, global hotkeys, Command Center, text
+      # replacement) silently stops working. That shipped in 1.0.0 as issue #22.
+      # Nothing fails at build time if this regresses, so it is checked explicitly.
+      - name: Verify app sandbox stays disabled
+        run: |
+          rc=0
+          for plist in Logue/Resources/Logue.entitlements Logue/Resources/Logue.release.entitlements; do
+            value=$(/usr/libexec/PlistBuddy -c "Print :com.apple.security.app-sandbox" "$plist" 2>/dev/null || echo "missing")
+            if [ "$value" != "false" ]; then
+              echo "::error file=$plist::app-sandbox is '$value', expected 'false' (see issue #22)"
+              rc=1
+            else
+              echo "OK: $plist has app-sandbox = false"
+            fi
+          done
+          exit $rc
+
   build:
     name: Build
     runs-on: macos-15
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -210,9 +210,15 @@ jobs:
             codesign --force --options runtime --sign "$SIGNING_IDENTITY" --timestamp "$item"
           done
 
-          # 3. Sign XPC services with proper entitlements for sandboxed Sparkle updates.
-          #    - Downloader.xpc: MUST be sandboxed with network access (downloads update ZIP)
-          #    - Installer.xpc: MUST NOT be sandboxed (needs to write to /Applications)
+          # 3. Sign Sparkle's XPC services.
+          #    Logue is no longer sandboxed, so Sparkle does not use these at all — it
+          #    installs directly via Autoupdate/Updater.app. They still ship inside
+          #    Sparkle.framework, so they must carry a valid signature or the outer
+          #    bundle fails verification. The sandboxed/non-sandboxed split below is
+          #    kept as Sparkle documents it, so re-enabling the sandbox does not
+          #    silently produce a broken updater.
+          #    - Downloader.xpc: sandboxed with network access (downloads update ZIP)
+          #    - Installer.xpc: NOT sandboxed (needs to write to /Applications)
           DOWNLOADER_ENT=$(mktemp)
           cat > "$DOWNLOADER_ENT" << 'ENTEOF'
           <?xml version="1.0" encoding="UTF-8"?>
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -75,8 +75,10 @@ macOS app (macOS 26+ / Tahoe): AI-powered meeting notes + document editing. Priv
 - **Sparkle EdDSA private key** lives only in the GitHub Actions secret `SPARKLE_PRIVATE_KEY` — never store it in the repo. The public key (`SUPublicEDKey`) is embedded in `Info.plist` via `project.yml`. Updates are served entirely from GitHub (appcast in-repo, assets on GitHub Releases) — there is no backend.
 - **Use typed constants for notification userInfo keys** — never string literals like `"success"` in `userInfo` dictionaries.
 - **Entitlements:**
-  - Dev builds: sandbox OFF for development convenience (documented in comments)
-  - Release builds: sandbox ON via `Logue.release.entitlements`
+  - **App Sandbox is OFF in both Debug and Release — do not re-enable it.** A sandboxed process cannot be an Accessibility API client, so macOS never lists the app under Privacy & Security → Accessibility and the user cannot add it manually. That silently kills every `Logue/CrossApp` feature (⌘⌃I inline assistant, global hotkeys, Command Center, text replacement). This shipped in 1.0.0 as issue #22. Guarded by `LogueTests/ReleaseEntitlementsTests.swift` and a CI step.
+  - Distribution is Developer ID + notarization via GitHub Releases and Sparkle — not the Mac App Store — so the sandbox is not required. **Hardened Runtime stays ON.**
+  - Sparkle needs no XPC services or `SUEnableInstallerLauncherService` when unsandboxed; those exist solely to work around the sandbox.
+  - Debug (`Logue.entitlements`) and Release (`Logue.release.entitlements`) still differ: Debug carries the iCloud entitlements, Release does not.
   - Every security exception (`allow-jit`, `disable-library-validation`, etc.) must have a comment explaining why it's needed
 
 ### Concurrency and Thread Safety
@@ -165,4 +167,5 @@ macOS app (macOS 26+ / Tahoe): AI-powered meeting notes + document editing. Priv
 | `Engine/LLMClient.swift` | External LLM provider clients (OpenAI/Anthropic-compatible) — optional, user-configured |
 | `Services/RecordingSessionManager.swift` | Recording lifecycle (`RecordingState` enum) |
 | `Services/EncryptionManager.swift` | AES-256-GCM encryption at rest (7-day migration window) |
+| `Services/SandboxContainerMigrator.swift` | One-time move of user data out of the pre-1.0.1 sandbox container (runs in `LogueApp.init()`) |
 | `App/AppConstants.swift` | All constants: `LLMDefaults`, `Audio`, `Diarization`, `Delays` |
--- a/Logue.xcodeproj/project.pbxproj
+++ b/Logue.xcodeproj/project.pbxproj
@@ -432,6 +432,7 @@
 		FA0EBF1706531E854E711B5E /* PIICategoryCard.swift in Sources */ = {isa = PBXBuildFile; fileRef = 68896D3492B91AC825FEC46E /* PIICategoryCard.swift */; };
 		FB2DCEFD67348622524BEB81 /* RecentContentPane.swift in Sources */ = {isa = PBXBuildFile; fileRef = 63AC1DBC374C162B470E69D4 /* RecentContentPane.swift */; };
 		FB715269552DA5B0A80404E1 /* OverviewPane.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6042105E2C13EF05700916B6 /* OverviewPane.swift */; };
+		FD7360D41D7027945F5BE947 /* SandboxContainerMigratorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87A5F89112EFC50A714AFD79 /* SandboxContainerMigratorTests.swift */; };
 		FDA9B040A07F32E153F7CFA7 /* InlineRewritePopover.swift in Sources */ = {isa = PBXBuildFile; fileRef = D5A27F72324468131F61DCF8 /* InlineRewritePopover.swift */; };
 		FE2793A4972BDC2CEB93C018 /* MenuBarIcon.png in Resources */ = {isa = PBXBuildFile; fileRef = EF7B96F52DB280D23D637F33 /* MenuBarIcon.png */; };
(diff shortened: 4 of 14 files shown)


shanforge (changes requested):
# Review: PR #23 — Disable App Sandbox so Logue appears in Accessibility settings

**Verdict: request one change before approving.** The concept is correct, the migration is well-built, and the code quality is high — but persisted meeting-audio paths are not rewritten, which silently breaks audio playback for every existing user who has recordings.

---

## 1. Is the concept true? Yes, fully

| Claim | Verdict |
| --- | --- |
| A sandboxed process cannot be an Accessibility API client | **True.** No entitlement grants it; `AXIsProcessTrustedWithOptions` returns `false`, TCC never creates an entry, so the app cannot appear under Privacy & Security → Accessibility and the user cannot add it manually. Matches issue #22 exactly. |
| Shipped 1.0.0 is sandboxed | **Verified independently:** `codesign -d --entitlements -` on `/Applications/Logue.app` → `com.apple.security.app-sandbox = true`. |
| Hardened Runtime stays ON | **Verified:** `flags=0x10000(runtime)`, `TeamIdentifier=HC5R66SXM5`. |
| Sandbox is not required for this distribution | **Correct.** The sandbox is only mandatory for the Mac App Store; Developer ID + notarization does not require it. |
| Sparkle's XPC services and `SUEnableInstallerLauncherService` are sandbox-only workarounds | **Correct** per Sparkle's own sandboxing guide — correct to drop them. |
| The 1.0.0 → 1.0.1 update is performed by the already-installed sandboxed 1.0.0, so the new bundle's entitlements do not affect the in-flight update | **Correct reasoning.** The updater that runs is the installed one. |

The blast-radius analysis is also right: global event monitors, synthetic `CGEvent`s, and Command Center all require the same trust, so **no release has ever had working cross-app features**.

---

## 2. Migration — strong, with one real gap

The risk is material and I confirmed it: the container on a real 1.0.0 install holds **4.3 GB** of user data at `~/Library/Containers/com.bitwize.logue`. Shipping the entitlement flip alone would have orphaned all of it.

### Verified correct in `SandboxContainerMigrator`

- **Ordering is safe.** Swift initializes stored properties *before* `init()`'s body, and `@NSApplicationDelegateAdaptor` constructs `AppDelegate()` during that phase — so this needed checking. AppDelegate's only substantive stored property is `SPUStandardUpdaterController`; every store (`MeetingStore.shared`, `ModelManager.shared`, `ScheduledTaskManager.shared`) is touched in `applicationDidFinishLaunching`, which runs **after** `LogueApp.init()`. Nothing creates Application Support early, so the "destination already exists" guard cannot misfire.
- **Crash-safe and idempotent.** The completion flag is set *after* the work, and `migrateItem` skips existing destinations, so an interrupted migration resumes correctly on the next launch.
- **Same-volume `moveItem` is a rename** — instant, and needs no extra free space for the multi-GB model directories.
- **Never clobbers an existing destination**; preferences are merged without overwriting, and `com.apple.*` keys are skipped.
- **`do/catch` + `os.log` throughout**, no silent `try?`. `APP_SANDBOX_CONTAINER_ID` is a reliable runtime sandbox check. Clean installs no-op.
- **Encrypted data stays decryptable.** `KeychainHelper` uses `kSecClassGenericPassword` with no `kSecUseDataProtectionKeychain` and no `keychain-access-groups` entitlement — i.e. the legacy file keychain, whose ACLs bind to the *designated requirement* (team ID + bundle ID), neither of which changes. `EncryptionManager`'s migration flag also lives in the Keychain, so the 7-day unencrypted-fallback window is not accidentally re-opened.

### Blocker: stale audio paths are not rewritten

`MeetingNote.audioFileURL: URL?` is persisted (`decodeIfPresent(URL.self, forKey: .audioFileURL)`), and recordings are written to `Application Support/com.bitwize.logue/Recordings/<meetingID>.m4a` — a path the migrator **does** move. But a `URL` encodes as an **absolute** path, and the migrator rewrites no URLs.

For every upgrading user with recordings:

- the audio file is moved to the real home directory, but the stored URL still points at `…/Library/Containers/com.bitwize.logue/Data/…`
- `AudioPlaybackView.swift:18` sees a non-nil URL and renders a player, but the file is not at that path → **playback silently fails**
- `MeetingStore.swift:256` also targets the stale path when deleting

Transcripts, notes, and summaries are unaffected and the files are recoverable, but this is a visible regression on upgrade.

**Suggested fix:** after moving files, rewrite any `audioFileURL` whose path contains `/Library/Containers/<bundleID>/Data/` by stripping that prefix. Longer term, deriving the path from `meetingID` at load time instead of persisting an absolute URL would eliminate this class of bug entirely.

### Secondary rough edge (not blocking)

The "destination already exists → skip" guard will commonly trigger for `~/.cache/huggingface`, which exists on any machine that has ever used Hugging Face tooling. When it does, up to ~2 GB stays orphaned in the container indefinitely, with only a log line and no user-visible notice, and models may be re-downloaded. Worth either surfacing or cleaning up.

---

## 3. Code quality

High. Stateless `enum` service with clear `MARK` sections, a single source of truth for `relativePaths`, and a `url(_:_:)` helper that avoids repeated component-by-component path building. The new key is declared in `AppConstants` per project convention. Comments explain **why** rather than what — the long entitlements comment is justified for a security decision a future contributor would otherwise "tidy up."

`CLAUDE.md` is updated to reverse the previous "Release builds: sandbox ON" rule, so the documented standard no longer contradicts the code. Regression guards exist in two places (`ReleaseEntitlementsTests.swift` plus a CI step, since CI does not run tests), and both were confirmed to actually fail when the sandbox is flipped back on.

**Nits (non-blocking):**

1. The entitlements comment states the test "fails the build if this flips back to true." Tests do not fail the build, and CI does not run tests — which is precisely why the CI step was added. Worth rewording.
2. `.github/workflows/release.yml` still special-cases Sparkle's XPC services (sandboxed `Downloader.xpc` with network access, non-sandboxed `Installer.xpc`). Those paths are now inert for an unsandboxed app — harmless, but could be simplified or annotated.
3. Keeping the inert sandbox-only entitlements is defensible and documented, though `codesign -d --entitlements` will now advertise capabilities that are not enforced.

---

## 4. Answering the upgrade/new-install question directly

- **New installations:** work cleanly. The migrator finds no container, records the flag, and no-ops.
- **Existing users upgrading from 1.0.0:** work, with the one exception above — meeting **audio playback** breaks because the persisted URLs are stale. Meetings, documents, spaces, templates, the vector store, and all downloaded models migrate correctly.
- **No other migration conflicts.** Keychain state survives, so encryption keys and the encryption-migration flag are preserved; there is no double-migration and no data-loss path.

westerosweb:
Addressed in e1c77cb.

## Blocker — stale audio paths

Confirmed and fixed. Recordings land in `Application Support/com.bitwize.logue/Recordings/<meetingID>.<ext>` (`RecordingSessionManager.moveRecordingFile` / `mixAudioFiles`), which the migrator moves, and a `URL` encodes as an absolute path — so meetings recorded under the sandbox kept pointing into the container.

I also checked whether anything else was affected: `MeetingNote.audioFileURL` is the **only** absolute path Logue persists. Every other path (`meetingsDirectory`, `storageURL`, `dbURL`, `digestCacheURL`, `ModelStorage.rootDirectory`, …) is computed at runtime, and nothing stores absolute paths in `UserDefaults`. So the fix is scoped to this one field.

Fixed at **decode time** in `MeetingNote.init(from:)` via `SandboxContainerMigrator.resolvingLegacyContainerPath`, which strips the `/Library/Containers/<bundleID>/Data` segment. Two properties worth calling out:

- it falls back to the original URL when nothing exists at the rewritten path, so a recording left in the container by a non-clobber skip still plays
- doing it at decode rather than as a one-shot rewrite makes it self-healing, and it also covers restored backups and imports

I confirmed the new decode test fails without the fix rather than trusting it:
```
✘ "Decoding a meeting recorded under the sandbox re-points its audio URL"
  Expectation failed: (decoded.audioFileURL?.path → ".../Library/Containers/com.bitwize.logue/Data/Library/Application Support/...")
                   == (migrated.path            → ".../Library/Application Support/...")
```

On the longer-term suggestion — deriving the path from `meetingID` at load time would eliminate the bug class, and I agree, but recordings use three extensions (`.m4a` mixed, `.wav` mic, `.caf` system), so the lookup isn't a single derived path. Left as a separate change rather than folding it in here.

## Secondary — stranded data on collision

Fixed rather than just surfaced. `migrateItem` skipped an entire tree when the destination existed; `move(_:to:label:)` now recursively merges, moving non-colliding children and only leaving a file behind when one already exists at the destination.

The non-clobber guarantee is unchanged — an existing destination file always wins and the container copy is kept for recovery — but `~/.cache/huggingface` no longer strands ~2 GB just because the directory exists. This also helps the dev-machine case where `Application Support/Logue` exists but holds a different meeting set.

## Nits

1. **Reworded** the entitlements comment. You're right that it was wrong on its own terms: tests don't fail the build and CI doesn't run them, which is exactly why the CI step exists. Both guards are now named accurately, with the CI step identified as the one that actually gates a PR.
2. **Annotated** the `release.yml` XPC block: the services are inert for an unsandboxed app but still ship inside `Sparkle.framework` and must carry valid signatures or outer-bundle verification fails. Kept the sandboxed/non-sandboxed split as Sparkle documents it so re-enabling the sandbox can't silently produce a broken updater.
3. **No change** to the inert sandbox-only entitlements, per your read that it's defensible — the trade-off is documented in the file.

## Verification

- 8 new cases in `SandboxContainerMigratorTests`: path rewrite, not-migrated fallback, pass-through of `nil`/remote/ordinary URLs, decode-time rewrite, and move / merge / never-overwrite / recursive-merge
- 33 tests pass across 5 suites; Release build succeeds; `swiftformat --lint` and `swiftlint --strict` clean (0 violations, 409 files); entitlements guard still reports `app-sandbox = false` for both files

Thanks for catching the audio path — that would have been a visible regression for every upgrading user with recordings.

shanforge (approved):
LGTM, thanks

PR #24, opened 2026-07-27
Support pasting over a multi-block selection (Cmd+V)

westerosweb:
## Summary

Multi-block selection supported Copy, Cut, Select All and Delete — but not Paste. `MultiBlockKeyHandler` had no `"v"` case, so `Cmd+V` did nothing while blocks were selected. You could copy a run of blocks and delete a run, but never replace one.

Adds `BlockEditorDocument.replaceBlocks(ids:withMarkdown:)`: removes the selected blocks, inserts the blocks parsed from the pasted markdown at the first selected position, and returns the last pasted block so the caret can be placed.

A non-contiguous selection collapses to that position. That differs from block *move*, which rejects non-contiguous selections — because a move would scramble document order, whereas a replace has no ordering left to preserve.

Paste inside a focused text block is untouched: `BlockTextView.paste(_:)` still owns the clipboard there, and this path only runs while `multiBlockSelection.isActive`.

## How was this verified?

- **11 new tests** — single and contiguous replacement, multi-block expansion, block-structure preservation, returned focus target, the three no-op cases (empty markdown, empty selection, unknown IDs), the non-empty-document invariant, and non-contiguous selection
- **159 tests pass, 0 fail** across all non-LLM suites — no regressions
- `make lint` clean: 0 violations in 410 files (SwiftFormat + SwiftLint `--strict`)
- Build succeeds with no new warnings

- [x] Ran `make setup` at least once, so the pre-commit hooks are installed
- [x] Builds cleanly
- [x] Lints cleanly (`make lint`)
- [x] No new `swiftlint:disable`
- [x] Added/updated tests
- [x] Ran `xcodegen generate`

## Checklist

- [x] Keeps all AI inference and data processing on-device — no network or inference code touched
- [x] No secrets, API keys, or personal data committed
- [x] Followed the standards in CONTRIBUTING.md / CLAUDE.md

## Notes for reviewers

The test suite includes an **emoji + CJK** case deliberately. `1e83485` recently fixed a whole-word find bug that trapped on multi-UTF-16 input, and it survived review partly because the surrounding tests used only ASCII (`"cat"`, `"dog"`, `"aaaa"`). Any text path in this codebase bridges `NSString` UTF-16 offsets and Swift `String` grapheme indices, so multi-byte coverage should be table stakes here.

Manual check worth doing: select several blocks, `Cmd+C`, select a different run, `Cmd+V`. The keyboard dispatch itself has no automated coverage — only the document mutation underneath does.

### DIFF
--- a/Logue.xcodeproj/project.pbxproj
+++ b/Logue.xcodeproj/project.pbxproj
@@ -214,6 +214,7 @@
 		6D86DD0D26811B19EEE00299 /* AIContentScorer.swift in Sources */ = {isa = PBXBuildFile; fileRef = BCC95C3045D2301DCC960849 /* AIContentScorer.swift */; };
 		6E33CC2D3A37790D824199A3 /* MeetingStore+Diarization.swift in Sources */ = {isa = PBXBuildFile; fileRef = E7FD01A02C6FB9C64A0238CF /* MeetingStore+Diarization.swift */; };
 		6E577F8B48F4A11B36E16B66 /* SelectableOptionCard.swift in Sources */ = {isa = PBXBuildFile; fileRef = 54288BA1E89F95FD74442B71 /* SelectableOptionCard.swift */; };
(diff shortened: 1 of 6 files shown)


westerosweb:
Closing — this work is going straight to main rather than through a PR.

PR #25, opened 2026-07-27
Add wikilink graph, document properties, saved views, and inbox

westerosweb:
## Summary

Builds out linking and organisation on top of the existing document model. 17 commits, ~250 new tests, no changes to recording, transcription, or inference.

**Linking and graph**
- `[[wikilinks]]` with a completion menu (`[[` opens a ranked list of documents and meetings), rendered as real links — brackets hidden, hand cursor, click to follow
- `LinkIndex`: outgoing links, backlinks, unresolved targets, resolving across documents *and* meetings
- Links panel in a new **Connect** sidebar group, showing the graph neighbourhood grouped by how each item connects
- Typed relationship fields (`belongs_to` / `has` / `related_to`) with **computed inverses** — declaring one side is enough
- Rename repair: renaming a document retargets every inbound `[[link]]` and relationship field

**Properties and organisation**
- Typed document properties (`type`, `status`, `url`, `due_date`, …) plus arbitrary custom keys, with an inspector panel
- Document types carrying icon, colour, sidebar order, default properties, and a body template
- Saved views with nested all/any filter groups, relative date expressions, and regex
- Bulk actions (tag, trash, move, organise) and a capture-then-organise inbox

**Editor**
- ⌘V over a multi-block selection (copy/cut/delete already existed; paste was the missing verb)
- Remote-image detection in pasted markdown, with SSRF-style protections

**Navigation**
- `logue://` deep links now actually navigate — the notifications previously had no observer
- Back history, so Back after following a link returns to the document you came from

## How was this verified?

- [x] Ran `make setup` at least once, so the pre-commit hooks are installed
- [x] Builds cleanly
- [x] Lints cleanly (`make lint` — SwiftFormat + SwiftLint `--strict`, 449 files, 0 violations)
- [x] No new `swiftlint:disable`
- [x] Added/updated tests — **427 passing, 0 failing**
- [x] Ran `xcodegen generate`

Wikilink clicking and the Back button were also **verified by hand in the running app**, because unit tests over the models did not catch either being unwired (see below).

## Checklist

- [x] Keeps all AI inference and data processing on-device — no inference, network, or model code touched. The one network-adjacent addition (remote-image detection) only *detects* URLs; nothing is fetched in this PR
- [x] No secrets, API keys, or personal data committed
- [x] Followed the standards in CONTRIBUTING.md / CLAUDE.md

## Notes for reviewers

**Please review commit-by-commit.** The messages carry the reasoning and the trade-offs; the squashed diff loses that.

Known trade-offs, each documented at its call site:

1. **Rename repair is not atomic.** Each affected document is saved individually and the encrypted store has no cross-document transaction, so a mid-way failure leaves some repaired and some stale. Stale links surface as "Unresolved" in the Links panel rather than vanishing, so it degrades visibly — but it is not all-or-nothing.
2. **`applyBulk` re-saves every document in the selection.** `WritingDocument` is not `Equatable`, so there is no cheap unchanged check. Judged better than making the whole model `Equatable`; open to the other call.
3. **`logue://space/<uuid>` parses but does not navigate.** Space selection lives in sidebar view state rather than a store, so there is nothing to set. Documented at the `case` instead of faking it.
4. **`LinkIndex` rebuilds rather than invalidating incrementally.** Simpler and correct at current library sizes; worth revisiting if anyone has tens of thousands of documents.

Back-compatibility detail worth a look: every new persisted field is **optional-backed**. Swift's synthesized `Codable` throws `keyNotFound` for a missing non-optional key *even when it has a default*, so a non-optional would make every existing document fail to decode. `isOrganised` additionally defaults to `false` when constructed but `nil` when decoded, so new documents land in the inbox while an existing library reads as already organised rather than flooding it.

**Where I got it wrong, in case it shapes how you review:** three features shipped "done" while broken, each because I tested the code I wrote and inferred the rest of the chain — wikilink clicks (styling, hit-testing and resolution all passed; AppKit never delivered the click), deep links (parsing and dispatch tested; nothing was listening), and the Back button (fixed `goBack()`, but the visible control calls `handleBreadcrumbClick`). All three are fixed and manually confirmed. The general lesson is that this repo has no UI-test coverage, so anything UI-adjacent needs the running app — a UI-test target would be a worthwhile follow-up.


Logue/Engine/MarkdownStyler.swift
```diff
@@ -214,6 +219,63 @@ final class MarkdownStyler {
...
+            // Hide the `|alias` separator and target when an alias is present, so the
+            // display text is what reads.
+            if link.displayText != nil {
+                let inner = (textStorage.string as NSString).substring(with: innerRange)
+                if let pipe = inner.range(of: "|") {
+                    let pipeOffset = inner.distance(from: inner.startIndex, to: pipe.lowerBound)
+                    bracketRanges.append(
+                        NSRange(location: innerRange.location, length: pipeOffset + 1)
+                    )
+                }
```
shanforge: **CRITICAL — this third append breaks the pairing invariant and corrupts unrelated text.**

`pairedDelimiter` documents and relies on `[opening, closing, opening, closing, …]` and pairs by `idx % 2`. An aliased link appends **three** ranges, so every delimiter added after it has flipped parity. `deleteBackward` trusts the pairing and deletes both ranges.

Repro — type `[[A|a]] [[B]]`, put the caret just after `[[` of `[[B]]`, press Backspace:
```
[[A|a]] [[B]]  ->  [[a]] B]]
```
One keystroke silently rewrites a *different* link, and `didChangeText()` persists it. `deleteBackward` lives on `MarkdownNSTextView`, so table cells are affected too.

Suggested fix: keep exactly two ranges per link by folding `Target|` into the **opening** delimiter (`location: link.range.location, length: 2 + pipeUTF16Offset + 1`), or track hidden ranges in a separate `hiddenRanges` array that is excluded from pairing.

Logue/Engine/MarkdownStyler.swift
```diff
@@ -214,6 +219,63 @@ final class MarkdownStyler {
...
+            )
+
+            // Hide the `|alias` separator and target when an alias is present, so the
+            // display text is what reads.
+            if link.displayText != nil {
+                let inner = (textStorage.string as NSString).substring(with: innerRange)
+                if let pipe = inner.range(of: "|") {
+                    let pipeOffset = inner.distance(from: inner.startIndex, to: pipe.lowerBound)
+                    bracketRanges.append(
+                        NSRange(location: innerRange.location, length: pipeOffset + 1)
```
shanforge: **UTF-16 vs grapheme mismatch.** `String.distance(from:to:)` counts `Character`s, but the result is used as an `NSRange.length` (UTF-16 units).

```
"[[Alpha|first]]"    -> {2,6}  want {2,6}   ok
"[[\u{1F600}A|alias]]"      -> {2,3}  want {2,4}   leaks "|alias"
"[[\u{1F468}\u{200D}\u{1F469}\u{200D}\u{1F467}|Family]]" -> {2,2}  want {2,9}   splits the ZWJ sequence
```
Any astral-plane char, ZWJ sequence, flag, or NFD accent before the `|` leaves the raw target and pipe visible and hands `addAttributes` a range that splits a surrogate pair.

Fix: `let pipe = (textStorage.string as NSString).range(of: "|", range: innerRange)` then `NSMaxRange(pipe) - innerRange.location`. This also removes the per-link full-string `substring` copy.

Note the existing tests only place emoji *before* the link, which is where the arithmetic is already correct — that is why this ships green. `InlineAttributeVisitor.swift:101` has the same pre-existing pattern.

Logue/Views/Editor/BlockEditor/Views/BlockTextView.swift
```diff
@@ -1287,6 +1290,22 @@ final class BlockNSTextView: MarkdownNSTextView {
             return
         }
         super.insertText(string, replacementRange: replacementRange)
+
+        // Typing the second `[` completes a `[[` trigger — offer link targets.
+        // Fired once here rather than on every keystroke because NSMenu is modal
```
shanforge, line 1297: **The completion menu fires inside code blocks and hijacks typing.** There is no `markdownStyleEnabled` guard, and code blocks use this same class (`BlockRowView` constructs `BlockTextView(... markdownStyleEnabled: false ...)`).

Repro: focus a code block and type a shell guard `if [[ -z "$x" ]]` (or `a[[1]]` in Python). On the second `[`, a modal `NSMenu` opens, the following keystrokes go to its type-select instead of your code, and Return inserts `[[Some Doc]]`.

Fix: `guard markdownStyleEnabled else { return }` before triggering (ideally also skip inline-code spans).

Separately: `NSMenu.popUp` spins a nested modal run loop from inside `insertText(_:replacementRange:)`, while `NSTextInputContext` is mid-`handleEvent`. Worth deferring presentation one run-loop turn.

Logue/Models/DocumentStore.swift
```diff
@@ -228,9 +236,41 @@ final class DocumentStore {
...
+    ///
+    /// Each affected document is saved individually. A failure part-way through
+    /// therefore leaves earlier documents repaired and later ones stale rather than
+    /// rolling back — the encrypted store has no cross-document transaction. Stale
+    /// links show up as "Unresolved" in the Links panel rather than being lost, so
+    /// partial repair degrades visibly instead of silently.
+    private func repairInboundLinks(from oldTitle: String, to newTitle: String, renamedID: UUID) {
+        guard oldTitle.caseInsensitiveCompare(newTitle) != .orderedSame else { return }
+
+        for document in documents where document.id != renamedID {
```
shanforge, line 256: **Rename repair misses meeting summaries and trashed documents.**

`LinkIndex.build` treats meeting summaries as first-class link *sources*, but this loop only walks `documents`, so renaming a document silently breaks every `[[link]]` inside every meeting summary. Trashed documents are skipped too, so restoring one later yields permanently broken links.

`LinkRenamer.rewriting(body:from:to:)` already works on any string, so covering meetings is small. I would also repair trashed docs — they are cheap, and the alternative is unrecoverable.

Logue/Models/DocumentStore.swift
```diff
@@ -228,9 +236,41 @@ final class DocumentStore {
...
+
+    /// Retargets `[[wikilinks]]` and relationship fields that pointed at the old
+    /// title, so a rename does not silently break every inbound reference.
+    ///
+    /// Each affected document is saved individually. A failure part-way through
+    /// therefore leaves earlier documents repaired and later ones stale rather than
+    /// rolling back — the encrypted store has no cross-document transaction. Stale
+    /// links show up as "Unresolved" in the Links panel rather than being lost, so
+    /// partial repair degrades visibly instead of silently.
+    private func repairInboundLinks(from oldTitle: String, to newTitle: String, renamedID: UUID) {
```
shanforge: **Four other title-mutation paths bypass this repair.** `repairInboundLinks` has exactly one caller (`renameDocument`), but titles are also mutated by:

- `:346` `generateAITitle`
- `:395` `regenerateAITitle` — **also missing `uniqueTitle`**, so it can mint a duplicate title that then hijacks `[[link]]` resolution via first-writer-wins
- `Logue/Agent/Tools/WriteDocumentTools.swift:124`
- `Logue/Services/MeetingStore.swift:305` `renameMeeting` (meetings are link targets)

These are automatic/AI paths, so a user's links break without them doing anything deliberate. Suggest routing every title change through one `rename` entry point that does `uniqueTitle` + repair.

Logue/Views/Editor/PropertiesPanelView.swift
```diff
@@ -0,0 +1,197 @@
...
+
+        default:
+            TextField("Value", text: $draft)
+                .textFieldStyle(.roundedBorder)
+                .font(.callout)
+                .onAppear { draft = value?.displayString ?? "" }
+                .onChange(of: draft) { _, newValue in
+                    // Keep list values as lists so filters can match individual entries.
+                    if case .list = value {
+                        onEdit(.list(newValue.split(separator: ",").map {
```
shanforge: **Opening the panel silently retypes values and writes to disk on every keystroke.**

`onAppear` mutates `draft` *after* the first body evaluation, which is a state change `onChange` observes — so `onEdit` fires with no user input. Merely opening the panel on a `.number(5)` property emits `onEdit(.text("5"))` and destroys the type. `PropertyValue` has 5 cases; this editor handles 2 and funnels the rest into `.text`, so `.number` is unreachable from the UI entirely.

Compounding: this is wired to `store.updateDocument`, which stamps `modifiedAt` and does an encrypted disk write — **per keystroke, undebounced** — while the rest of the editor debounces via `AppConstants.Delays`. And `modifiedAt` is what saved views sort on.

Fix: switch exhaustively over `value` with a `.number` branch that parses back to `Double`, and commit on `.onSubmit`/focus-loss (or debounce) instead of `onChange`.

Logue/Models/DocumentStore+Organisation.swift
```diff
@@ -0,0 +1,184 @@
...
+                Self.logger.error(
+                    "Failed to save \(url.lastPathComponent, privacy: .public): \(error.localizedDescription, privacy: .public)"
+                )
+            }
+        }
+    }
+
+    // MARK: - Saved views
+
+    func addSavedView(_ view: SavedView) {
```
shanforge, line 98: **This whole CRUD surface is unreachable.** `addSavedView`, `updateSavedView`, `deleteSavedView`, `addDocumentType`, `updateDocumentType`, `deleteDocumentType` each have **zero call sites** outside this file (verified by grep across `Logue/`).

Consequences: nothing ever writes `saved-views.json`, so `savedViews` is permanently `[]`, so the `Section("Views")` in `DocumentListPane` (guarded by `!store.savedViews.isEmpty`) can never render. The 293 lines of `SavedView.swift` — nested filter groups, 9 comparators, regex, relative dates — are reachable only from tests. `documentTypes` is likewise never persisted, so the starter types get fresh UUIDs on every launch.

Shipping an on-disk schema that nothing can produce is a future migration liability. Suggest either landing the editor UI in this PR, or splitting these out and landing them with their consumer.

Logue/Models/DocumentStore+Organisation.swift
```diff
@@ -0,0 +1,184 @@
...
+
+    // MARK: - Bulk actions
+
+    /// Applies a bulk transform to the selected documents and saves each changed one.
+    func applyBulk(
+        to ids: Set<UUID>,
+        transform: ([WritingDocument]) -> [WritingDocument]
+    ) {
+        let selected = documents.filter { ids.contains($0.id) }
+        guard !selected.isEmpty else { return }
```
shanforge, line 176: **`applyBulk` skips the cache invalidation and selection cleanup the single-document paths perform.**

It never calls `invalidateCaches()` / `rebuildIndexMap()`, unlike `addTag`/`removeTag` (`DocumentStore.swift:284-300`) and `deleteDocument` (`:179-189`).

- Stale tag index: after a bulk tag, `store.allTags` still returns the memoised pre-bulk list, so the new tag is missing from tag UI until an unrelated single-doc edit clears it.
- Worse: bulk-trash the currently open document and `selectedDocumentID` is untouched — the editor keeps displaying **and saving into** a trashed document. `BulkAction.trashing` also does not clear `spaceID` (unlike `deleteDocument`).

Fix: after the loop call `invalidateCaches()`, and clear `selectedDocumentID` when the selected doc became trashed.

Logue/Services/ContentNavigator.swift
```diff
@@ -0,0 +1,136 @@
...
+        }
+        select(target)
+    }
+
+    /// Returns to the previous link-navigation target. Returns whether it moved, so
+    /// the caller can fall back to its own back behaviour.
+    @discardableResult
+    static func goBack() -> Bool {
+        guard let previous = NavigationHistory.shared.popPrevious() else {
+            logger.debug("Back requested but history is empty")
```
shanforge, line 81: **`goBack()` reports success even when the navigation did not happen → dead Back button.**

This returns `true` whenever an entry was popped, but `select()` silently returns when the id is unknown. Callers (`MainWindowView+Navigation.swift:56-59`) trust the boolean and early-return, skipping the list fallback.

Repro: in A follow `[[B]]`, permanently delete A from Trash, press Back → the entry pops, selection is refused, nothing happens.

Compounding: `navigateToContentBreadcrumb` does `storeChangeVersion += 1` *before* calling this, so when no selection change follows, the counter stays ahead of `lastSeenStoreVersion` and the **next real sidebar click is swallowed** by the version guard. CLAUDE.md calls this out directly: sync `lastSeenVersion` in the handler that increments the counter.

Also: the guard checks membership but not `isTrashed`, so Back can open a trashed document, contradicting `resolve` which excludes them. And `NavigationHistory.clear()` is never called anywhere, so Back can jump to a stale document instead of the list.

Logue/Models/PropertyValue.swift
```diff
@@ -0,0 +1,160 @@
...
+            try container.encode(value, forKey: .value)
+        case let .list(values):
+            try container.encode(Kind.list, forKey: .kind)
+            try container.encode(values, forKey: .value)
+        }
+    }
+
+    init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let kind = try container.decode(Kind.self, forKey: .kind)
```
shanforge: **An unknown `kind` throws and takes the whole document with it.** `try container.decode(Kind.self, ...)` throws `DataCorrupted` on an unrecognised value; that propagates out of `[String: PropertyValue]` → out of `WritingDocument.init(from:)` → `DocumentStore` logs and **skips the file**, so the document disappears from the library.

This is the exact hazard the sibling design deliberately avoided — `WritingDocument.swift:92-93` stores relationships as raw strings so the format "survives an unknown key from a newer build." `PropertyValue` has no such tolerance.

Concretely: a user on a newer build that adds a case writes `{"kind":"reference",…}`; rolling back to this build makes every document with that property vanish.

Fix: add `case unknown(String)` (or fall back to `.text`) when `Kind(rawValue:)` fails, and re-encode it unchanged so the round trip is lossless. The same fragility applies to `FilterComparator` / `SavedViewSort` / `FilterGroup.Logic`.

Logue/Views/Editor/BlockEditor/Views/MultiBlockKeyHandler.swift
```diff
@@ -98,6 +104,9 @@ final class KeyHandlerNSView: NSView {
         case "a":
             coordinator?.onSelectAll()
             return true
+        case "v":
+            coordinator?.onPaste?()
```
shanforge: **Cmd+V is consumed even when nothing handles it.** `onPaste` is optional-chained (unlike `onCut()`/`onSelectAll()`), so nil is a reachable state — and this still `return true`, swallowing the event instead of passing it down the responder chain. Same when the clipboard holds no `.string` (image-only).

Also, `charactersIgnoringModifiers == "v"` matches **Cmd+Option+V** as well. (Cmd+Shift+V is safe — that yields `"V"`.)

Fix:
```swift
case "v":
    guard let onPaste = coordinator?.onPaste else { return super.performKeyEquivalent(with: event) }
    onPaste()
    return true
```

Logue/Engine/LinkRenamer.swift
```diff
@@ -0,0 +1,105 @@
...
+                "[[\(replacement)|\(display)]]"
+            } else {
+                "[[\(replacement)]]"
+            }
+
+            result = (result as NSString).replacingCharacters(in: link.range, with: rebuilt)
+        }
+
+        // Guard against a caller passing a body that no longer matches the ranges.
+        return nsBody.length == (body as NSString).length ? result : body
```
shanforge: **This guard is tautological — and the comment promises a safety property the code cannot provide.**

`body` is an immutable parameter and the loop only mutates `result`, so both sides measure the same string: the condition is unconditionally `true` and the `: body` branch is unreachable. No test can cover it.

Fix: delete this ternary and the `nsBody` binding on line 25, and `return result` — the ranges come from the same `body` in the same call, so nothing is needed. (Or express the real invariant, if one was intended.)

Logue/Views/Editor/LinksPanelView.swift
```diff
@@ -0,0 +1,191 @@
...
+                }
+            }
+            .padding(16)
+        }
+        .task(id: rebuildKey) { rebuild() }
+    }
+
+    /// Rebuild when the document changes or its body is edited.
+    private var rebuildKey: String {
+        "\(documentID.uuidString)-\(documentBody.count)"
```
shanforge: **Change detection uses body *length*, so the panel goes stale; and the rebuild runs on the main actor.**

Two issues:

1. `documentBody.count` misses any length-preserving edit — retyping `[[Notes]]` as `[[Ideas]]` keeps the key identical, so the panel keeps showing `Notes` as an outgoing link and never shows `Ideas`. The index is a snapshot of the **whole library**, but the key only observes the current document, so another document being renamed/created/trashed while the panel is open leaves phantom backlinks. CLAUDE.md prescribes hash-based change detection — `documentBody.hashValue`, or better a hash of `WikiLinkParser.uniqueTargets(in:)` so prose edits do not rebuild the graph at all, combined with a store mutation counter.

2. `rebuild()` is called from a non-async `.task` closure, i.e. on `@MainActor`, and `LinkIndex.build` runs `NSRegularExpression` over every document body and every meeting summary. The block editor commits `body` on a 300 ms debounce, so with this panel open the entire library is re-scanned on the main thread every ~300 ms of typing. Suggest building off the main actor (`LinkIndex` and its inputs are already `Sendable`).

Logue/Models/SavedView.swift
```diff
@@ -0,0 +1,293 @@
...
+        case .isOnOrAfter: return subject >= boundary
+        case .isBefore: return subject < boundary
+        default: return false
+        }
+    }
+
+    /// An invalid pattern matches nothing. User-authored regexes are expected to be
+    /// wrong sometimes, and a filter is not a place to throw.
+    private func matchesRegex(_ candidates: [String]) -> Bool {
+        guard let regex = try? NSRegularExpression(pattern: value) else { return false }
```
shanforge: `NSRegularExpression` is compiled inside a function called **once per document per evaluation** — filtering 1,000 documents compiles the same pattern 1,000 times, on the main thread, with a user-authored pattern and no backtracking bound. The codebase's own convention is the cached static (`WikiLinkParser.swift:39`, `RemoteImageImporter.swift:25`).

Also, the `try?` silently swallows an invalid pattern, so a user staring at empty filter results gets no signal. Note `RelativeDate` already declares a `Logger` at `SavedView.swift:11` that is never used — logging the compile failure there closes both points at once.

Logue/Engine/RemoteImageImporter.swift
```diff
@@ -0,0 +1,123 @@
...
+        /// UTF-16 range of the whole `![alt](url)` reference.
+        let range: NSRange
+    }
+
+    private static let logger = Logger(subsystem: AppConstants.bundleID, category: "RemoteImageImport")
+
+    /// `![alt](url)` — alt may be empty, the URL may not contain spaces or parens.
+    private static let pattern = "!\\[([^\\]\n]*)\\]\\(([^)\\s]+)\\)"
+
+    private static let regex: NSRegularExpression? = try? NSRegularExpression(pattern: pattern)
```
shanforge: **Privacy: cleared — but this file is dead code.**

To be explicit for the record, I verified three ways that this makes **no network requests**: there is no `URLSession` / `dataTask` / `data(from:)` anywhere in this file or the diff, it has **zero call sites** in `Logue/`, and the validation is stricter than required (HTTP(S) only; blocks loopback, `*.local`, `::1`, `fc/fd/fe80`, `10/*`, `127/*`, `192.168`, `172.16-31`, and `169.254` cloud metadata; filenames are SHA-256 derived, not path derived). The `absoluteString` on line 83 is hash input, not a log. So the PR's "nothing is fetched" claim holds.

The issue is that 123 lines of production code plus 159 lines of tests certify a feature that cannot run — the commit landed detection with no importer and no paste-path wiring. Green tests over unreachable code read as "this works." Suggest wiring it up here or landing it with its consumer.

⚠️ Blocking requirement for whoever adds the fetch: `isImportable` only validates the **original** URL, so a permitted host can 302 to `169.254.169.254`. The follow-up needs redirect re-validation, plus an opt-in gate, `URLSession` timeouts, and a byte cap.

Logue/Views/Sidebar/DocumentListPane.swift
```diff
@@ -64,12 +66,27 @@ struct DocumentListPane: View {
     @State private var renameText = ""
     @FocusState private var isRenameFieldFocused: Bool
     @State private var hasAutoSelected = false
+    /// Active saved view, when one is chosen instead of a built-in filter.
+    @State private var activeSavedViewID: UUID?
```
shanforge: **Three mutually-exclusive filter states kept in sync by hand.** `activeSavedViewID` and `activeTypeName` are added alongside the existing `filterMode`, and every write has to nil the other two manually (`:124-125`, `:142-144`, `:162-163`), with the invariant re-derived a fourth time in a view guard at `:244`.

This is what CLAUDE.md forbids directly: *"Use `RecordingState` enum — never use separate boolean flags for state machines."*

Fix: one enum, e.g.
```swift
enum ListFilter { case builtin(DocFilterMode), savedView(UUID), type(String) }
```
with `filteredDocs` and the inbox-header check both derived from it. That removes the three nil-out sites and makes the fourth impossible to get wrong.

Logue/Services/LinkIndex.swift
```diff
@@ -0,0 +1,238 @@
...
+    /// Every item with at least one outgoing or incoming link.
+    var linkedItemIDs: Set<UUID> {
+        Set(outgoingByID.keys).union(backlinksByID.keys)
+    }
+
+    // MARK: - Private
+
+    /// Titles are matched case-insensitively and ignore surrounding whitespace,
+    /// so `[[  alpha ]]` finds a note titled `Alpha`.
+    private static func normalise(_ title: String) -> String {
```
shanforge: `normalise(_:)` is duplicated character-for-character in three places: here, `ContentNavigator.swift:133`, and `LinkRenamer.swift:102`.

This two-line function *is* the app's title-identity rule — it decides what `[[Foo]]` matches. Three copies means the graph, the navigator, and the renamer can silently disagree after any future tweak. Suggest one definition (`WikiLinkParser.titleKey(_:)`) used by all three.

Related duplication worth folding while you are here: "strip link syntax from a target" exists three times (`LinkRenamer.swift:99`, `LinkIndex.swift:147-152`, `ContentNavigator.swift:32`), and the link-syntax sanitiser twice (`WikiLinkCompletion.insertionText`, `LinkRenamer.sanitisedTitle`) — a divergence in the latter is a link-forgery vector.

Also note there are two competing definitions of link *resolution*: `LinkIndex.resolve` uses first-writer-wins over `idsByNormalisedTitle`, while `ContentNavigator.resolve` re-implements it with an explicit documents-before-meetings rule. They agree today only because `build` happens to append documents first — an accident, not a contract.

shanforge (changes requested):
## Review summary

Reviewed locally with five parallel agents over an isolated worktree, with the highest-severity findings independently re-verified. Detail is in the inline comments; this is the map.

**Requesting changes** — the craft on individual types is high and mechanical rule compliance is spotless, but there are two confirmed corruption/hijack bugs and roughly 1,000 lines of unreachable production code that ship an on-disk schema nothing can create.

### First, the part that matters most: back-compatibility is safe

Existing v1.0.0 documents still decode. All three new persisted fields are `Optional`, so the synthesized decoder uses `decodeIfPresent`; `storedIsOrganised ?? true` is a genuinely nice touch that makes an existing library read as *already organised* instead of flooding the Inbox; documents are one-file-per-doc with a per-file `catch`, so no single bad file can take out the library. The legacy-JSON decode tests pin it.

I also confirmed **nothing was dropped** by the `MainWindowView` −32 lines (pure re-homing; every modifier, `onChange` and observer intact), there is **no data-loss path in multi-block paste** (validation precedes `recordUndo()`/`removeAll`), single-block paste is untouched, plain click/drag-select is unaffected, `AppDelegate` observers are correctly removed, and recording/transcription/inference/diarization are not touched at all.

### Blocking

1. **Backspace corrupts a different wikilink, and persists it** — the third delimiter range breaks the `pairedDelimiter` parity invariant. `MarkdownStyler.swift:272`
2. **`[[` menu hijacks typing inside code blocks** — no `markdownStyleEnabled` guard. `BlockTextView.swift:1295`
3. **UTF-16 vs grapheme bug in the alias hide-range** — leaks the target and splits surrogate pairs. `MarkdownStyler.swift:270`
4. **Rename repair misses meeting summaries and trashed docs**, and four other title paths bypass it entirely (two of them AI-triggered). `DocumentStore.swift:256,259`
5. **Properties panel destroys typed values and writes to disk per keystroke.** `PropertiesPanelView.swift:188`
6. **~1,000 lines unreachable** — the saved-view/document-type CRUD surface and `RemoteImageImporter` have zero call sites. `DocumentStore+Organisation.swift:67`, `RemoteImageImporter.swift:25`
7. **`goBack()` returns `true` when navigation was refused**, and leaves a version counter ahead so the next sidebar click is swallowed. `ContentNavigator.swift:80`
8. **Unknown `PropertyValue.kind` throws and the document vanishes.** `PropertyValue.swift:79`
9. **Cmd+V swallowed when unhandled.** `MultiBlockKeyHandler.swift:108`
10. **Table-cell wikilinks are styled as links but hand `logue-wikilink://` to macOS** ("no application set to open the URL") — the click handler lives only on `BlockNSTextView` while `TableCellNSTextView` extends `MarkdownNSTextView`, and the scheme is not registered. Moving `clicked(onLink:at:)` and `wikiLinkTarget(at:)` up to `MarkdownNSTextView` fixes it.

### Worth fixing, not blocking

Links-panel staleness + main-actor graph rebuild, `applyBulk` cache/selection cleanup, the new `Logue/organisation/` directory being outside both `clearAllData()` and `BackupManager`, the three hand-synced filter states, the triplicated `normalise(_:)`, the tautological guard in `LinkRenamer`, and a doc comment orphaned from `invalidate()` by the new members inserted above it (`MarkdownStyler.swift:219-226`).

### Genuinely well done

Zero force-unwraps, force-casts, over-long lines, oversized functions or new suppressions across 5,870 lines; every delay in `AppConstants.Delays`; `@Observable` throughout. **Zero restate-the-obvious comments** — the `WikiLinkURL`-vs-`DeepLink` rationale, the `.link`-attribute-vs-`mouseDown` dead end, and the `characterIndexForInsertion` rounding note all record things a reviewer would otherwise have to ask about. `WikiLinkClickTests` building a real text view and hit-testing real glyph rects is the right way to test AppKit. `LinkIndex` is designed for testability rather than retrofitted for it.

### One structural note

Your own note says three features shipped "done" while broken because you tested the code you wrote and inferred the rest of the chain. That pattern repeated here — items 2, 6 and 10 are all "logic tested, chain unwired." Given how much is unreachable, this may be worth splitting: land the wikilink graph after 1-5 and 10, and hold saved views / document types / remote images until their UI exists.

Logue/Services/LinkIndex.swift
westerosweb: Fixed the triplicated `normalise(_:)` — it is now `WikiLinkParser.titleKey(_:)` and the graph, navigator and renamer all call it.

The rest of this comment I have **not** done, and I would rather say so than have it look handled:

- "strip link syntax from a target" still exists in three places.
- `WikiLinkCompletion.insertionText` and `LinkRenamer.sanitisedTitle` are still separate sanitisers. Agreed this is the one with teeth — a divergence there is a forgery vector, not just duplication.
- `LinkIndex.resolve` and `ContentNavigator.resolve` still implement resolution twice, agreeing today only because `build` happens to append documents before meetings. You are right that this is an accident rather than a contract.

Those three want one `WikiLinkIdentity` type owning key, sanitisation and resolution, with both call sites reduced to it. That is a wider refactor than the rest of this pass and I did not want to fold it in alongside ten behavioural fixes where a mistake would be hard to spot. Tracking it as a follow-up rather than leaving it implied.

Logue/Views/Sidebar/DocumentListPane.swift
westerosweb: Not done, and it is a fair reading of the rule — three mutually-exclusive states with the invariant maintained by hand at four sites is exactly what the enum guidance exists to prevent.

My reason for deferring rather than doing it: this pass added the write surface for saved views and types on top of these same three properties, so collapsing them into `ListFilter` at the same time would have put a state-machine refactor and a new feature in one diff over the same lines. I would rather land it separately where a mistake is visible.

One thing that changed in the meantime, which makes the current shape slightly less bad: the three nil-out sites are now four, but deleting a saved view or a type also clears the corresponding state (`deleteSavedView`, `deleteType`), so the case where a stale id or name silently widened the filter to everything is gone. That was the failure mode with actual consequences; the hand-sync is now a tidiness problem.

Logue/Models/PropertyValue.swift
westerosweb: Fixed for `PropertyValue`: there is now a `.unknown(kind:value:)` case that keeps the original tag *and* its payload — via a small `PreservedJSON` type — so the round trip is lossless including nested objects and arrays. Tests cover the whole-document case you described (encode, rewrite one property as a future kind, decode, document still there), the round trip, and a missing `value`.

On the sibling point: `FilterComparator`, `SavedViewSort` and `FilterGroup.Logic` are **not** yet tolerant. Their blast radius is smaller — an unknown value fails the `saved-views.json` decode, which `load` turns into `[]`, so the user loses their saved views rather than their documents — but "loses all saved views" is still worse than "ignores one condition". It needs a decision about what an unknown comparator should *mean* when matching (match nothing, or drop the condition and match everything), which changes filter semantics, so I did not want to pick one silently. Following up separately.

westerosweb:
## Review addressed — 4 commits, `efc5355..fc8d12f`

All ten blocking items are fixed. I verified each against the code before changing anything, and every behavioural fix has a test that fails on the previous commit — item 1's test reproduces your exact repro and shows the opening of `[[B]]` pairing to offset 2, inside the first link.

Two of the four decisions were not mine to make, so they went to the repo owner: fixes land on this branch (not on the storage PR that contains these commits), and the unreachable code gets wired up rather than removed — except the image importer, which is deleted.

### Blocking

| # | Item | Fix |
| --- | --- | --- |
| 1 | Backspace corrupts a different wikilink | `Target\|` folds into the **opening** delimiter, so every link contributes exactly two ranges and parity holds |
| 2 | `[[` menu hijacks code blocks | `markdownStyleEnabled` guard; presentation deferred one run-loop turn out of `insertText` |
| 3 | UTF-16 vs grapheme in the alias hide-range | measured with `NSString.range(of:)` |
| 4 | Rename repair misses meetings, trashed docs, and four paths | one `applyTitle(_:to:)` entry point; repair covers meeting summaries and trashed documents; renaming a meeting repairs links to it |
| 5 | Properties panel destroys types and writes per keystroke | exhaustive switch with a `.number` branch; commits on submit/focus-loss |
| 6 | ~1,000 unreachable lines | saved views and types wired up; `RemoteImageImporter` deleted |
| 7 | `goBack()` returns true when refused | `select()` reports what it did; counter only bumped when something changes; trashed excluded; `clear()` now called |
| 8 | Unknown `PropertyValue.kind` throws, document vanishes | `.unknown(kind:value:)` with lossless re-encode |
| 9 | Cmd+V swallowed when unhandled | falls through to the responder chain; Cmd+Option+V no longer matches |
| 10 | Table-cell wikilinks hand the scheme to macOS | handler moved up to `MarkdownNSTextView` |

On **6**, what made them reachable: "Save Current Filter as View…" captures the current filter and sort, with rename/delete for the active view; "New Type from This Document…" builds a type from a document's own properties, with rename/delete. Two supporting changes were needed to make "save what I'm looking at" true — `FilterField` gained `isPinned`/`isOrganised`, since Pinned and Inbox were not expressible, and starter types are persisted on first run instead of getting fresh identifiers every launch.

Your blocking note about redirect re-validation is recorded with the deletion, so whoever adds the fetch inherits it rather than rediscovering it.

### Also fixed

`applyBulk` cache and selection cleanup (bulk-trashing the open document left the editor saving into a trashed document) · Links-panel staleness now keyed on link targets rather than body length, and the graph rebuild moved off the main actor · `normalise(_:)` unified as `WikiLinkParser.titleKey(_:)` · filter regex compiled once and invalid patterns logged · `LinkRenamer`'s tautological guard removed · the orphaned `invalidate()` doc comment reunited with its declaration · the organisation directory added to `clearAllData()` and to backups, with optional fields so v1 backups still decode.

### Not done, deliberately

Replied in-thread on each: the three filter states → one enum, the remaining sanitiser/resolution duplication, and unknown-value tolerance for `FilterComparator`/`SavedViewSort`/`FilterGroup.Logic`. Each is a refactor or a semantic decision I did not want to fold in beside ten behavioural fixes.

### On your structural note

You were right that items 2, 6 and 10 were all "logic tested, chain unwired", and that this is the same pattern I had already written down. What I have taken from it concretely: the two tests that certified `WriteEchoFilter` in the storage PR were removed for the same reason, because `isEcho` turned out to have no caller either — a test naming a mechanism is worth less than a test naming the outcome. 431 tests pass here.

shanforge (approved):
LGTM, thanks!!!

PR #26, opened 2026-07-28
Store documents as plain markdown in ~/Logue, editable outside the app

westerosweb:
## Plain markdown storage

Documents can be stored as ordinary `.md` files in `~/Logue` instead of encrypted JSON in Application Support, so they can be edited in another editor, tracked in git, or read by an agent. Off by default. Meetings, transcripts and audio are unaffected and always stay encrypted.

**The folder is the storage, not a copy of it.** There is no mirror and nothing to reconcile: while the setting is on, `DocumentStore` reads and writes through `DocumentStorage` and the encrypted path is not used for document content at all.

> **Stacked on #25.** This branch was cut from `feat/knowledge-graph-and-document-organisation`, so 17 of its 33 commits belong to that PR. Reviewing #25 first, or reading only the 16 commits after `Add the document content and derived split`, keeps the two apart.

### What it does

| Outside the app | In the app |
| --- | --- |
| File edited | document updated |
| New `.md` file | imported as a new document |
| File renamed | still the same document |
| File moved between folders | document refiled |
| File deleted | document moved to trash |
| New folder, nested included | new space |
| Folder renamed or moved | space follows it |
| Folder deleted | space, nested spaces, and their documents trashed |

And the same in reverse: creating, renaming, moving or deleting a space writes its folder immediately. That is not a nicety — it is what earns a scan the right to read a missing folder as a deletion instead of guessing.

### Decisions worth reviewing

**Turning it on is ordered so failure is never destructive.** Warning sheet with an explicit acknowledgement → create the folder, refusing one that holds unrelated files → write every document → **read each back and compare id, title and body** → only then flip the setting. Anything short of complete success leaves the setting and the originals alone. The encrypted originals are kept even on success: they cost kilobytes and they are what makes this reversible.

**Turning it off moves the folder to the Trash**, as the default. A folder left behind is no longer read or written but still looks exactly like the library, so an edit made there silently does nothing — two apparent sources of truth, which is what this design exists to remove. The Trash rather than deletion, and only after the documents are safely re-encrypted. "Turn Off and Keep Folder" is offered for anyone tracking it in git.

**`~/Logue`, not `~/Documents/Logue`.** iCloud Drive's "Desktop & Documents Folders" option syncs `~/Documents`, so keeping unencrypted notes there meant a setting the user turned on years ago could put them in the cloud today. A folder left at the old path is moved, and never merged into a folder already at the new one.

**Echoes are matched by content, not path or time.** A path filter cannot tell our write from the user's write to the same file a moment later, and a time window silently widens into one where real edits are dropped. Comparing what is on disk against what we wrote stays correct however long the write took.

**A missing root folder means do nothing.** An unmounted volume, a folder the user moved, and an unfinished sync all look identical to "every folder was deleted" — and acting on that would empty the library. Guarded in the scan itself, not only at the call site.

**Trash stays inside the app.** Deleted documents are not written to the folder — nobody wants deleted notes sitting in their home folder — so they stay in encrypted storage and are carried back when the setting is turned off.

### Also here

- **`_space.md`** gives a folder its space identity, so renaming a folder is a rename rather than a delete-and-create. Visible rather than a dotfile, and it explains itself in a comment: someone who breaks hidden state has no way to learn why their space changed.
- **Rescan button** in the sidebar, shown only when the setting is on, for changes made while the app was not running. It spins while scanning, which required making the scan async — it previously completed inside one main-actor call, so no progress indicator of any kind could have appeared.
- **Loading states** in place of empty ones while the stores read from disk. Launching used to flash "No Documents — create a new document to get started" over a library that was seconds from appearing, and Home greeted returning users with the new-user welcome screen.
- **The mirror is gone** — `MarkdownMirror`, conflict resolution, the sync button and the folder bookmark, about 1,250 lines. Along with everything the three earlier designs left behind and the type names that outlived the concept.

### Testing

663 tests pass; `make lint` clean (SwiftFormat + SwiftLint strict).

The filesystem work is exercised against **real temporary directories**, not mocks, including a read-only one to prove export reports failure rather than claiming success. There is an end-to-end suite that drives the same entry points in the same order the app does, because unit tests per link have not been enough in this codebase — and it earned its place immediately by catching a path comparison that ignored symlinks, where every unit involved had passing tests but every scan created a space named after the root folder.

The re-enable tests were run against the old code first: 7 of 9 fail there.

### Known limitation

*Partial* absence is indistinguishable from deletion. If a sync service removes files locally without them being really gone, a scan reads that as the user deleting them. Everything it does is recoverable — trash, not deletion — but it is a real hazard of living in the home folder and is written down in the design notes.

### Not verified by machine

`DocumentStore.applyExternalChanges` and `SpaceStore`'s folder writes run against main-actor singletons and real Application Support, so the tests drive an equivalent harness rather than the stores themselves. The UI work — rescan button, spin, loading states — is compiled and reasoned about but not observed.


Logue/Services/MarkdownStorageMigrator.swift
```diff
@@ -0,0 +1,679 @@
...
+            logger.error("Could not claim a dropped-in file: \(error.localizedDescription, privacy: .public)")
+            return nil
+        }
+        return content
+    }
+
+    /// Every `.md` file under the root, skipping hidden directories so a `.git` folder is
+    /// never walked.
+    func markdownFiles() -> [URL] {
+        guard let enumerator = FileManager.default.enumerator(
```
shanforge, line 610: **CRITICAL — a root that cannot be walked reads as "every document was deleted".**

This is `enumerator(at:)` without the `errorHandler:` overload, so a traversal failure returns a short list rather than an error. `MarkdownFolderScan.isRootPresent` uses `fileExists`, which **follows symlinks and succeeds**, so the missing-root guard passes and the scan proceeds against an apparently empty library. `ExternalChangePlan` then trashes every document that was not seen.

I reproduced this against real APFS:

```
symlinked root  -> fileExists: true,  files walked: []
real directory  -> fileExists: true,  files walked: ["note.md"]
```

`chmod 000` on the root behaves the same way, and an unreadable *sub*directory silently trashes just that subtree.

The symlink case is the one that worries me most, because pointing `~/Logue` at an Obsidian vault or a Dropbox folder is the most obvious thing to do with a feature advertised as "plain files you can edit anywhere". And it does not self-heal: once trashed, those identifiers hit the `!existing.isTrashed` guard and become `ignoredIdentifiers` permanently.

Suggested fixes, in order of value:
1. **Never trash on absence from the walk alone.** Before trashing document X, `fileExists` its last-known URL — you already keep `cachedFileIndex`. That single check also defeats the concurrent-move race and sync-eviction.
2. Use the `errorHandler:` overload and abort the scan on any enumeration error.
3. `resolvingSymlinksInPath()` on the root, or reject a symlinked root explicitly.

Logue/Services/MarkdownStorageMigrator.swift
```diff
@@ -0,0 +1,679 @@
...
+            }
+        }
+        return (index, duplicates)
+    }
+
+    /// Reads a written file back and confirms it says what the document says.
+    private func verify(_ content: DocumentContent, at url: URL) throws {
+        let contents = try String(contentsOf: url, encoding: .utf8)
+        guard let readBack = MarkdownDocumentFile.content(from: contents) else {
+            throw VerificationError.unreadable
```
shanforge, line 373: **This is the root cause of several separate data-loss findings.**

`verify()` compares `id`, `title` and `body` — three fields out of roughly fifteen on `DocumentContent`. Anything it does not inspect can be silently destroyed by the round trip and still "verify successfully", which is what lets the migration flip the mode with confidence.

Concretely, extending this to compare the full `DocumentContent` (excluding the deliberately-derived `modifiedAt`/`isTrashed`) would have caught, at development time:

- typed properties collapsing to `.text` (see `MarkdownDocumentFile.swift`)
- `storedIsOrganised` coming back `false` instead of `nil`
- `goalMode` never being written at all
- a custom property named `title` shadowing the real one

I would treat this as the highest-leverage change in the PR: one fix, four bugs.

Logue/Services/MarkdownStorageMigrator.swift
```diff
@@ -0,0 +1,679 @@
...
+        content.title = fields.title
+        content.body = fields.body
+        content.tags = fields.tags
+        content.spaceID = spaceFolderMap(in: knownSpaces).spaceID(
+            forComponents: directoryComponents(of: url), in: knownSpaces
+        )
+
+        do {
+            let rendered = MarkdownDocumentFile.render(content)
+            try rendered.write(to: url, atomically: true, encoding: .utf8)
```
shanforge, line 558: **A file caught mid-write is adopted as a new document *and* overwritten.**

`MarkdownDocumentFile.content(from:)` returns `nil` the moment `_logue_id` is not parseable, which is exactly what a truncated file looks like. A non-atomic external writer — shell redirection, `git checkout`, a sync client streaming chunks, an editor without atomic save — is visible in that state, and the debounce (0.2s FSEvents + 0.4s) is easily exceeded by a large file or a network volume.

The whole chain happens in one pass: the truncated file parses as unidentified → `adopt()` writes our frontmatter **over the user's in-flight content** → a duplicate document appears → and because the original document is no longer in `scanned`, the same pass also puts it in `plan.trashed`.

So the scan destroys the file's contents and trashes the document that file *was*. Given this is described as the one write a scan performs, it deserves a guard: refuse to adopt a file modified within the debounce window, refuse to adopt a path that was in the previous `fileIndex` under a different identifier, and never let one pass both adopt a path and trash the document that path held.

Logue/Services/MarkdownStorageMigrator.swift
```diff
@@ -0,0 +1,679 @@
...
+        result.export = export(documents: documents, spaces: spaces, reusing: existing)
+        return result
+    }
+
+    /// Whether the file says something different from the document.
+    ///
+    /// A file that cannot be parsed counts as divergent: it is not what we would write, and
+    /// preserving it costs one item in the Trash.
+    private func divergesFromDisk(_ document: DocumentContent, at url: URL) -> Bool {
+        guard let onDisk = try? String(contentsOf: url, encoding: .utf8) else { return false }
```
shanforge: Silent `try?` here has a specific consequence. Returning `false` means "does not diverge", so an **unreadable** file is not retired — and is then silently overwritten by the export a few lines above.

The method's own doc says a file that cannot be *parsed* counts as divergent because "it is not what we would write, and preserving it costs one item in the Trash". An unreadable file should follow that same rule and return `true`.

(Also flagging the general case: there are ~11 silent `try?` on filesystem I/O in this file, against the CLAUDE.md rule. Most degrade acceptably; `prepareRoot`'s is the other one that matters, since swallowing a directory-read error makes an unreadable folder look **empty**, i.e. safe to adopt.)

Logue/Services/DocumentStorage.swift
```diff
@@ -0,0 +1,485 @@
...
+            )
+        }
+
+        mode = .encrypted
+        invalidateFileIndex()
+
+        // Last, and only once the documents are in hand: if reading the folder went wrong, the
+        // folder is the only place that text exists.
+        if retiringFolder {
+            do {
```
shanforge: **The folder is trashed before the documents are durably re-encrypted.**

The comment above says "only once the documents are in hand" — but *in hand* here means in memory. `mode` is flipped, `retireRoot()` moves `~/Logue` to the Trash, and the caller then persists via an **unawaited** `Task` whose write failures are log-only.

Worst case: a user enables markdown storage and works for a month (documents created in that period have **no** encrypted copy, since `saveDocument` returns before the encrypted branch), then turns it off. The folder goes to the Trash, the detached writes hit a full disk or a permission error, and the failure is a log line. The user is shown success. The only copy of a month's work is now in the Trash, and emptying it is permanent.

Fix: `await` the re-encryption and confirm every document landed on disk *before* calling `retireRoot()`, and surface write failures as a thrown `SwitchError` rather than a log.

Logue/Views/Settings/Tabs/PrivacySettingsTab.swift
```diff
@@ -23,6 +31,118 @@ struct PrivacySettingsTab: View {
...
+
+        do {
+            let restored = try storage.switchToEncrypted(
+                knownSpaces: spaceStore.spaces,
+                retiringFolder: retiringFolder,
+                // What the folder has to account for. Fewer means files are missing rather than
+                // deleted, and switching would prune the encrypted copies of the difference.
+                expectedDocumentCount: documentStore.activeDocuments.count
+            )
+            Task { await documentStore.adoptAfterStorageSwitch(restored) }
```
shanforge: This `Task { ... }` is the unawaited half of the problem above: `switchToEncrypted` has already trashed the folder by the time this runs, and `adoptAfterStorageSwitch` → `replaceAll` → `saveDocument` writes through `Task.detached` whose failures are logged and never surfaced.

The error message a few lines below tells the user "Your documents were re-encrypted, but the folder could not be moved to the Trash" — which is the *opposite* failure from the one that is actually possible here.

Logue/Engine/MarkdownDocumentFile.swift
```diff
@@ -0,0 +1,165 @@
...
+        if let organised = content.storedIsOrganised {
+            fields.append(("organised", .scalar(organised ? "true" : "false")))
+        }
+
+        // Properties, key-sorted for stability.
+        for key in (content.properties ?? [:]).keys.sorted() {
+            guard let value = content.properties?[key] else { continue }
+            switch value {
+            case let .list(items): fields.append((key, .list(items)))
+            default: fields.append((key, .scalar(value.displayString)))
```
shanforge: **Typed properties are destroyed on the first write, and the loss is then applied live.**

Everything that is not a list is written as `value.displayString` and read back as `.text(raw)`. So `.number(5)` → `"5"` → `.text("5")`, `.boolean(true)` → `.text("Yes")`, and `.date` → a **locale-dependent** `DateFormatter.medium` string that cannot be parsed back.

Two consequences beyond the type loss:

1. `ExternalChangePlanner.differs` *does* compare `properties`, so the very next scan sees `.text != .date`, plans an update, and overwrites the in-memory typed value. The DatePicker/number editor is gone for good, and turning the setting off persists the text into encrypted storage.
2. `.unknown(kind:value:)` — the case `PreservedJSON` exists to protect, so a document written by a newer build survives an older one — is flattened to text too, defeating its stated purpose in markdown mode.

The locale-dependent date rendering also breaks the "identical bytes for an identical document" guarantee this file claims.

Fix: encode the kind in frontmatter — ISO-8601 for dates, bare numerics as `.number`, `true`/`false` as `.boolean` — or keep a type map in `DocumentDerived`.

Logue/Engine/MarkdownDocumentFile.swift
```diff
@@ -0,0 +1,165 @@
...
+        else { return nil }
+        return UUID(uuidString: raw)
+    }
+
+    /// Reads content from a file, or `nil` when it carries no identifier.
+    static func content(from markdown: String) -> DocumentContent? {
+        guard let id = identifier(in: markdown) else { return nil }
+
+        let parsed = MarkdownFrontmatter.parse(markdown)
+        var content = WritingDocument().content
```
shanforge, line 192: **This starting point dumps an entire existing library into the Inbox.**

`WritingDocument().content` has `storedIsOrganised == false`, but the invariant deliberately established for this field is that **`nil` means organised** — `isOrganised` is `storedIsOrganised ?? true`, and `WritingDocument.swift` explains why: "Defaulting them to unorganised would dump an entire existing library into the inbox on update."

The write only emits the key when non-nil, so every document written by v1.0.0 (where it is `nil`) round-trips to `false`. Enable markdown storage, relaunch, and the whole pre-existing library is in the Inbox — and `switchToEncrypted` then persists `false` back into encrypted storage.

`MarkdownDocumentFileTests.populated()` sets `isOrganised = true`, which is exactly the case that masks this.

Fix: set `content.storedIsOrganised = nil` before parsing and assign only when the key is present.

Logue/Engine/DocumentFilename.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    /// A safe filename for `document`, avoiding any name in `taken`.
+    static func filename(for document: WritingDocument, avoiding taken: Set<String> = []) -> String {
+        let stem = safeStem(from: document.title)
+        let candidate = "\(stem).\(fileExtension)"
+        guard taken.contains(candidate) else { return candidate }
+
+        // Disambiguate with a prefix of the document identifier: stable for a given
+        // document, so a collision does not produce a different name on every save.
+        let shortID = document.id.uuidString.prefix(8)
```
shanforge, line 43: **`taken` is a `Set<String>` compared case-sensitively, but the filesystem is not.**

`uniqueTitle` dedupes document titles case-sensitively, so `Notes`/`notes` and `TODO`/`todo` coexist in shipped libraries today. Confirmed on this machine's APFS:

```
$ printf 'doc one\n' > Note.md ; printf 'doc two\n' > note.md ; ls -1
Note.md
$ cat Note.md
doc two
```

On enable, the second document of a case-pair resolves to a name `taken` does not contain, and `write(to:atomically:)` renames **over the first document's file**. `verify()` then reads back the winner and passes, so the mode flips and the loser's markdown is gone with no Trash copy. The next scan finds it fileless and trashes the document too, and `switchToEncrypted` refuses to switch back with `.folderIncomplete`.

The same applies on the move path when a destination folder already holds the other casing.

Note `SpaceFolderLayout.component(for:)` already uses `caseInsensitiveCompare` for sibling folders — the rule is understood, this path just missed it. Since the fold is filesystem-dependent, I would also `fileExists`-check the destination before writing.

Logue/Engine/ExternalChangePlan.swift
```diff
@@ -0,0 +1,144 @@
...
+        plan.inserted = adopted
+
+        // A document with no file is a deletion — but only if the app thinks it should
+        // have one. Trashed documents are deliberately absent from the folder, so treating
+        // their absence as a deletion would empty the trash on every scan.
+        // Ambiguous documents are skipped here as well as above. Their files exist — there are two
+        // of them — so trashing them for having none would be the worst possible reading.
+        for document in known
+            where !document.isTrashed && !seen.contains(document.id)
+            && !ambiguous.contains(document.id)
```
shanforge, line 107: This is the rule the whole design rests on — "a document the walk did not return is a deletion" — and it currently has no cross-check.

Every data-loss finding in this review flows through here: an unwalkable or symlinked root, an unreadable subdirectory, a sync client evicting files, a file truncated mid-write, and a document moved between folders while the walk is in flight all produce "absent from the walk" without anything being deleted.

One guard fixes the whole class: before appending to `plan.trashed`, check the document's last-known URL with `FileManager.fileExists`. `cachedFileIndex` already has it. Absence from the walk plus absence from disk is a deletion; absence from the walk alone is not.

Logue/Services/DocumentStorage+Rescan.swift
```diff
@@ -0,0 +1,269 @@
...
+        // Top of each deleted subtree only — `deleteSpace` cascades, and asking it to delete a
+        // child after its parent is gone would find nothing.
+        let roots = vanished.filter { id in
+            guard let parentID = spaceStore.spaces.first(where: { $0.id == id })?.parentID
+            else { return true }
+            return !vanished.contains(parentID)
+        }
+
+        for id in roots {
+            spaceStore.deleteSpace(id: id)
```
shanforge: `retiringFolder` defaults to `true`, and this is the call site its documentation was written for:

> `retiringFolder` is false when the deletion *came from* the folder disappearing. Retiring then is not merely redundant: the path is recomputed from the name, and if the user has already made a new folder there — deleting and recreating one is a normal thing to do — this would move that new folder to the Trash.

Grep confirms **no call site anywhere passes `false`**, so the parameter is currently dead. Today the hazard is defused only because `vanishedSpaceIDs` has just proved the derived path is absent — an invariant two files away, holding back a `trashItem` on a user-named directory.

Worth noting the reason no test catches it: both scan harnesses reimplement `deleteSpace` without folder retirement, so production's destructive branch is never exercised. Passing `retiringFolder: false` here is a one-word fix.

Logue/Services/DocumentStorage+Rescan.swift
```diff
@@ -0,0 +1,269 @@
...
+
+        report(plan, duplicatedFolders: scan.duplicatedSpaceFolders(in: spaceStore.spaces))
+
+        // The plan was diffed against `known`, taken before the folder was read. The user can type
+        // during that read, and applying an update built from the older snapshot would replace what
+        // they just typed with what the file said beforehand. Anything that moved on is dropped and
+        // left for the next scan, which diffs against the newer text.
+        let settled = ExternalChangePlanner.discardingUpdatesThatMovedOn(
+            plan, comparedTo: known, current: documentStore.documents.map(\.content)
+        )
```
shanforge, line 144: The concurrent-edit filter protects `updated` but not `trashed` — `discardingUpdatesThatMovedOn` returns early when `plan.updated.isEmpty`, so insertions and deletions pass through untouched.

The walk runs detached while `save()` runs on the main actor. Moving a document out of a space writes the new file then removes the old, and the enumerator visits the root before descending — so a move that straddles the walk reads as "no file at root, no file in the space" → the document is trashed, and `removeFile` then moves the *new* file to the macOS Trash.

Also in this area: a held-back update is only logged, and `hasPendingScan` is not set, so the external edit is never retried. `kFSEventStreamCreateFlagIgnoreSelf` means our own next write produces no event either, so the user's edit sits unapplied until the app overwrites it. Setting `hasPendingScan = true` whenever `settled.updated.count != plan.updated.count` would close that.

Logue/Models/DocumentStore.swift
```diff
@@ -411,6 +392,15 @@ final class DocumentStore {
     func saveDocument(id: UUID) {
         guard let index = documentIndex(for: id) else { return }
         let doc = documents[index]
+
+        // Checked before the encrypted write, not after: in markdown mode the file *is*
+        // the document, and writing both would recreate the two-copies problem this
+        // storage design exists to avoid.
+        if DocumentStorage.shared.save(doc, spaces: SpaceStore.shared.spaces) {
```
shanforge, line 399: This mode check is right, and the comment explaining it is good — but its sibling `saveToDisk()` further down has no equivalent check and writes encrypted files unconditionally.

`BackupManager.applyImport` uses `saveToDisk()`, so restoring a backup while markdown storage is on creates **no `.md` files** for the restored documents. The next scan reads "document with no file" as a deletion and trashes the entire restore — and scans fire on every app activation. On relaunch it is worse, since `loadDocuments` reads only the folder.

`GeneralSettingsTab`'s "load example data" has the same shape.

Repro: enable markdown storage → Settings → Backup → Import → Replace → switch away and back → every restored document is in the trash.

Fix: have `applyImport` go through `replaceAll(with:)` so each document is routed through `saveDocument`, or add the mode check inside `saveToDisk()`.

Logue/Views/Editor/PropertiesPanelView.swift
```diff
@@ -0,0 +1,242 @@
...
+            // Per keystroke, undebounced, while the rest of the editor debounces. `modifiedAt` is
+            // also what saved views sort on, so opening a panel reordered the library.
+            TextField("Value", text: $draft)
+                .textFieldStyle(.roundedBorder)
+                .font(.callout)
+                .onAppear { draft = value?.displayString ?? "" }
+                .onSubmit { commitDraft() }
+                .onChange(of: isFocused) { wasFocused, isFocused in
+                    if wasFocused, !isFocused {
+                        commitDraft()
```
shanforge, line 196: Committing on any focus loss with no dirty check undoes part of the `PropertyValue.unknown` fix in this same PR.

`commitDraft` turns the value into `.text(draft)`, so simply tabbing through the panel discards the `PreservedJSON` payload that `.unknown` exists to carry — and stamps `modifiedAt`, re-sorting the library, with no user edit. For an `.unknown` holding an object or array, `displayString` is `"—"`, so the committed value becomes literally `.text("—")`.

A dirty check against the original value would fix both the type loss and the spurious `modifiedAt`.

Logue/Views/Sidebar/DocumentListPane+Organise.swift
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    func renameType(_ type: DocumentType, to name: String) {
+        let trimmed = name.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else { return }
+
+        // The filter tracks types by name, so it has to follow the rename or the list empties.
+        if activeTypeName == type.name {
+            activeTypeName = trimmed
+        }
```
shanforge, line 110: Renaming a document type orphans every document of that type.

Documents store their type as a text property written at apply time, and the filter is `$0.typeName == activeTypeName`. After a rename the filter matches zero documents, and since the old name is no longer offered in the menu, those documents become unreachable by type filter permanently.

This is the same rename-breaks-references bug the PR just fixed for document titles via `applyTitle` — `renameType` needs the equivalent repair pass over documents carrying the old name.

Logue/Services/DocumentStorage.swift
```diff
@@ -0,0 +1,485 @@
...
+        guard Self.moveFolderIfLeftBehind(
+            from: Self.legacyMarkdownRootURL, to: Self.markdownRootURL
+        )
+        else { return }
+        logger.info("Moved the documents folder out of Documents into the home folder")
+    }
+
+    private init() {
+        let raw = UserDefaults.standard.string(forKey: AppConstants.UserDefaultsKeys.documentStorageMode)
+        mode = DocumentStorageMode(rawValue: raw ?? "") ?? .encrypted
```
shanforge, line 183: Worth calling out as done well: an unset **and** a corrupt/unknown value both fall back to `.encrypted`. A bad preference cannot silently put a user's notes on disk in plaintext, which is the right direction to fail.

I verified the rest of the opt-in story too — the acknowledgement checkbox gates the button, and the README/SECURITY/CHANGELOG claims were corrected honestly rather than left overstating the encryption promise.

shanforge (changes requested):
## Review summary

Reviewed locally with five parallel agents over an isolated worktree. Every severe finding below I reproduced myself before writing it up — two of them by running code against real APFS rather than reasoning about it.

I want to lead with what is genuinely strong here, because the design work is not in question. The pure/IO split (`ExternalChangePlan`, `SpaceFolder*`, `MarkdownDocumentFile`, `MarkdownFrontmatter`, `DocumentFilename`) is the cleanest in the codebase and made this reviewable. The comments record real bugs and their failure modes rather than restating code — the symlink `/var` vs `/private/var` note and the `SpaceFolderLayout` mid-walk explanation are the kind of thing that saves the next person a day. 214 new tests drive real temporary directories through the real entry points. The old mirror machinery is genuinely gone; I grepped for it.

**Verified correct, and these were my biggest worries going in:**

- **Existing v1.0.0 documents still decode.** No new non-optional persisted field reaches `WritingDocument`; the encrypted read path is untouched.
- **The feature is genuinely opt-in and fails safe** — an unset or corrupt setting falls back to `.encrypted`.
- **The enable path is non-destructive**, exactly as described: write, verify, then flip, with the encrypted originals kept. I traced crash, disk-full and permission-denied and all three leave the setting off and the data intact.
- Meetings, transcripts and audio never leave encrypted storage. `~/Logue` over `~/Documents/Logue` does avoid the iCloud exposure as claimed.

**Requesting changes** because the *disable* path and the *scan* path can lose data, and because the round trip silently drops fields the verification never looks at.

### The single highest-leverage change

`verify()` compares three fields out of roughly fifteen. Extending it to the full `DocumentContent` would have caught four separate findings below at development time — the property flattening, the `isOrganised` inversion, `goalMode`, and reserved-key shadowing. If only one thing changes, I would make it that one.

### Must fix — data loss

1. **A symlinked or unreadable `~/Logue` trashes the entire library**, and it does not self-heal. Reproduced. Pointing the folder at an Obsidian vault or Dropbox is the obvious use of this feature.
2. **A file caught mid-write is adopted and overwritten**, and the document it belonged to is trashed in the same pass.
3. **Two titles differing only in case overwrite each other's files**, with no Trash copy. Reproduced on APFS.
4. **Disable trashes the folder while the encrypted writes are still pending** and unawaited.
5. **Restoring a backup in markdown mode is trashed by the next scan** — `saveToDisk()` never learned about the mode its sibling `saveDocument` checks.
6. **The whole legacy library lands in the Inbox** — the markdown read starts from `false` where the invariant requires `nil`.
7. **Typed properties collapse to text** on the first write, and the loss is then applied live by the next scan.

### Should fix

`goalMode` is never written, so it resets library-wide · "Clear everything" (Troubleshooting) leaves the full library in plaintext in `~/Logue` while wiping the setting that would let the app find it again · the Privacy tab renders "documents are encrypted on disk" directly above the section explaining they are not · deleting `_space.md` can trash a space whose name the app would spell differently · a folder put back from the Trash is never re-adopted, because `SpaceFolderAdoption.resolve` is unreachable · a held-back external edit is never retried · four full folder walks per scan on the main actor, now on every app activation.

### On the #25 fixes carried in this branch

11 of the 14 are fully fixed, several with good regression tests — the backspace corruption, the code-block hijack, table-cell clicks, the unreachable CRUD, and `PropertyValue` no longer losing documents are all properly done. Three are partial:

- `InlineAttributeVisitor.swift:101` still has the grapheme-offset-as-`NSRange` pattern that was fixed in `MarkdownStyler`.
- The title consolidation behind `applyTitle` is good, but `MeetingStore+AI` still assigns `title` directly with no repair and no `uniqueTitle`. To be fair, that file is **not touched by this PR**, so this is an incomplete claim rather than a new defect — worth either fixing or narrowing the claim, since meetings are link targets.
- Two new defects came in with the fixes: renaming a document type orphans its documents, and the Properties panel commits on focus-out with no dirty check.

Ten of the fourteen fixes landed without tests, which is the gap the `MeetingStore` case slipped through.

### Suggestion

Given 17 of the 43 commits belong to #25 and that half is close to mergeable, splitting would let the graph work land while the storage half gets another pass. In priority order for that pass: never trash on absence alone (kills findings 1, 2 and most of the race conditions), extend `verify()` (kills 6, 7 and two others), await the re-encryption before trashing, and compare filenames case-insensitively.

Logue/Services/MarkdownStorageMigrator.swift
westerosweb: Fixed, but **not** by rejecting the symlink — I want to flag the deviation.

Refusing a symlinked root would break the case you identified as most likely: pointing `~/Logue` at an Obsidian vault or a synced folder is the obvious use of a feature advertised as plain files. The actual defect is narrower — `fileExists` follows a symlink and `enumerator(at:)` does not — so the root is now resolved in `init` and a symlinked root *works*. The test asserts that: a symlinked root is walked, its document found, nothing trashed.

The other two suggestions are in as written. The walk uses the `errorHandler:` overload and reports completeness, and a partial walk protects **every** document rather than aborting, because we cannot say which corner went unread. And your first suggestion is the one that matters most — `plan` now takes `stillOnDisk`, built by `fileExists`-ing each document's last-known URL. Absence from the walk plus absence from disk is a deletion; absence from the walk alone is now reported as `unwalkable` and left alone.

`chmod 000` on the root is covered too: `isRootPresent` confirms the folder can actually be *listed*, not merely stat'ed.

Logue/Services/MarkdownStorageMigrator.swift
westerosweb: Done, and it earned its keep immediately — extending it turned three of the four bugs you predicted into verification failures rather than silent losses.

`verify()` now compares tags, icon, goal, pinned, `storedIsOrganised`, width, properties, relationships and `createdAt` (to the second, since the file stores ISO-8601 and memory has sub-second precision), on top of id/title/body. `modifiedAt` is excluded because it is read from the file's own timestamp, and `isTrashed` because a trashed document deliberately has no file.

You were right that this was the highest-leverage change in the PR.

Logue/Services/MarkdownStorageMigrator.swift
westerosweb: Fixed. Adoption now requires the file to have been untouched for two seconds — comfortably longer than the 0.2s FSEvents coalescing plus 0.4s debounce, because the point is to outlast a writer that is not atomic. The window is injectable, so tests adopt immediately rather than sleeping.

The other two halves of the chain are closed as well: adoption is skipped entirely when the walk was incomplete, and the same pass can no longer both adopt a path and trash the document that path held, because trashing now requires the document's last-known file to be absent from disk too.

Tests cover all three: a file written moments ago is not adopted, its contents are unchanged afterwards, and one that has settled is.

Logue/Services/MarkdownStorageMigrator.swift
westerosweb: Both of the ones that matter are fixed.

`divergesFromDisk` returns `true` for an unreadable file, following the rule its own documentation already stated — so it is retired to the Trash rather than silently overwritten. And `prepareRoot` now throws when the folder is there but cannot be listed, instead of treating an unreadable directory as an empty one that is safe to adopt.

On the general case: the remaining `try?` in this file are the degrade-acceptably ones you described — a single unreadable file skipped during an import, a `_space.md` that will not parse — and each is now reachable through a path that treats absence as a question rather than a deletion, which is what made them dangerous before. I did not convert them wholesale, on the grounds that the risk was the *consequence* rather than the syntax.

Logue/Services/DocumentStorage.swift
westerosweb: Fixed as described. `retireRoot()` is out of `switchToEncrypted` entirely and into `retireFolderAfterReEncryption(of:)`, which the caller reaches only after `await`ing the re-encryption — and which checks each document has an encrypted file on disk before touching the folder, throwing `SwitchError.reEncryptionIncomplete` if any is missing.

`DocumentStore.awaitPendingSaves()` is what makes the await possible: `saveDocument` writes in a detached task, so the caller previously had no way to know the writes had landed.

Your worst case was the right one to lead with. Documents created while the setting was on have no encrypted copy at all, so the folder was the only copy of a month's work at the moment it went to the Trash.

Logue/Views/Settings/Tabs/PrivacySettingsTab.swift
westerosweb: Fixed with the parent issue. The `Task` now does the work in order and awaits it — adopt, await the pending saves, and only then retire the folder — and a failure surfaces as the actual error rather than a log line.

The misleading message is gone too. It described the one failure that could not happen while staying silent about the one that could.

Logue/Engine/MarkdownDocumentFile.swift
westerosweb: Fixed by encoding the kind in frontmatter, per your first suggestion: bare numerics for `.number`, `true`/`false` for `.boolean`, ISO-8601 for `.date`, lists as lists.

Two details worth stating because they are trade-offs rather than obvious wins:

- **Text that looks typed is quoted**, so a text property holding `"5"` or `"true"` stays text. The residual cost is inference in the other direction — a text property whose value happens to be an ISO date reads back as a date. I took that over a sidecar type map, because a sidecar means the file alone no longer describes the document, and editing a date in another editor could not work. That is the promise the whole mode rests on.
- **`.unknown` keeps its own tagged form** (`logue-kind:<kind>:<json>`) rather than being flattened, so `PreservedJSON` still does the job it exists for in the mode where the file is the only copy.

The locale-dependent date rendering is gone with it, so the identical-bytes guarantee holds again. And `verify()` now compares properties, so this class of loss fails the migration instead of passing it.

Logue/Engine/MarkdownDocumentFile.swift
westerosweb: Fixed: the read starts from `content.storedIsOrganised = nil` and assigns only when the key is present.

You also identified why the tests did not catch it — `populated()` sets `isOrganised = true`, which is the one value that masks the bug. `verify()` now compares `storedIsOrganised`, so the nil case fails the migration rather than quietly inverting.

Logue/Engine/DocumentFilename.swift
westerosweb: Fixed on both paths. `taken` is matched case-insensitively in `DocumentFilename.filename`, and the move path additionally `fileExists`-checks the destination, since as you say the fold is the filesystem's to decide rather than ours to predict.

The test exports `Notes` and `notes` together and asserts two files exist *and* that both bodies survived — the second part being the one that would have caught the original, since a count check alone passes when one file has been overwritten.

You were right that the rule was already understood elsewhere: `SpaceFolderLayout.component(for:)` had it for sibling folders.

Logue/Engine/ExternalChangePlan.swift
westerosweb: This is the fix the rest of the review pointed at, and it is in as you described it: `plan` takes `stillOnDisk`, built by `fileExists`-ing each document's last-known URL from the file index. Absence from the walk plus absence from disk is a deletion. Absence from the walk alone now lands in `plan.unwalkable`, which is logged and never applied.

An incomplete walk protects every document rather than none, because when the traversal failed we cannot say which subtree went unread.

It does close the class you listed. The symlinked root, the unreadable subdirectory, the evicted file, the truncated file and the document moved mid-walk all produce absence, and none of them now produces a deletion.

Logue/Models/DocumentStore.swift
westerosweb: Fixed inside `saveToDisk()`, so every caller is covered rather than just `applyImport` — the sample-data loader has the same shape, as you noted, and a third caller would have had it too.

In markdown mode it now routes each document through `DocumentStorage.save`, which is what `saveDocument` already did. Your repro is the one I checked it against: enable, import a backup with Replace, switch away and back.

Logue/Views/Editor/PropertiesPanelView.swift
westerosweb: Fixed with a dirty check against the original value, which closes both halves — the `.unknown` payload is no longer discarded by tabbing through, and `modifiedAt` is no longer stamped without an edit.

Worth recording that this came in *with* the `PropertyValue.unknown` fix in the same PR, which is a good argument for the point you made about ten of fourteen fixes landing without tests.

Logue/Views/Sidebar/DocumentListPane+Organise.swift
westerosweb: Fixed: `renameType` now calls `store.retypeDocuments(from:to:)`, which rewrites the type property on every document carrying the old name and saves each one.

You are right that it is the same bug `applyTitle` had just fixed for titles — I built the rename path for types without applying the lesson from the one directly above it in the same PR.

Logue/Services/DocumentStorage.swift
westerosweb: Noted, and left as it is. The fallback to `.encrypted` on an unset or unrecognised value is deliberate for exactly the reason you give: a bad preference must not be able to put notes on disk in plaintext.

One thing did change nearby since you looked — the Privacy tab's encryption line now reads the mode instead of asserting the default, because it was claiming documents are encrypted directly above the section explaining that they are not once this setting is on.

Logue/Services/DocumentStorage+Rescan.swift
westerosweb: Fixed — `retiringFolder: false` is passed at this call site now, so the parameter is no longer dead and the hazard is no longer held back only by an invariant two files away.

The reason you gave for no test catching it is **still true**, and I have not closed it: both scan harnesses reimplement `deleteSpace` rather than driving `SpaceStore`, so production's destructive branch remains unexercised. That is the same class of gap as the `WriteEchoFilter` one — a test that reimplements the thing it is testing cannot fail the way production does. Leaving this thread open as the record of it, since the test is the part that is missing rather than the fix.

Logue/Services/DocumentStorage+Rescan.swift
westerosweb: Both halves addressed, one of them differently from the suggestion.

**The held-back edit is retried.** `hasPendingScan = true` whenever `settled.updated.count != plan.updated.count`, exactly as you proposed. You were right that `IgnoreSelf` left nothing to retrigger it, so the edit sat unapplied until the app overwrote it.

**The move-during-walk case is fixed upstream instead of by extending the filter.** Rather than teaching `discardingUpdatesThatMovedOn` about `trashed`, the plan now cross-checks each document's last-known file with `fileExists` before trashing it — so a move that straddles the walk cannot produce a deletion in the first place, whichever order the enumerator visited things in. That felt like the better place: extending the concurrent-edit filter would have covered this race, while the cross-check covers the whole family (unreadable subtree, evicted file, truncated file) with one rule.

Related, since you flagged the walk cost in the summary: the scan now takes **one** traversal instead of seven, built off the main actor and shared by every step via `FolderSnapshot`. The `fileIndex()` this cross-check needs comes from that same snapshot, so the guard costs nothing extra.

westerosweb:
## Review addressed — `359090d..6cfec88`

All seven must-fix data-loss items are fixed, along with the "should fix" list and the three partial #25 carry-overs. Every claim I verified in the source before changing anything; replies are in each thread with what changed and why. **15 of 16 threads resolved** — the one left open is the test-coverage gap you identified, explained below.

### Must fix

| # | Item | Fix |
| --- | --- | --- |
| 1 | Symlinked/unreadable root trashes the library | Absence cross-checked against disk; walk reports completeness; symlinked root **resolved rather than refused** |
| 2 | File mid-write adopted and overwritten | Adoption requires a 2s settle, is skipped on an incomplete walk, and cannot coexist with trashing the same path's document |
| 3 | Case-only titles overwrite each other's files | `taken` matched case-insensitively; move path also checks the destination on disk |
| 4 | Disable trashes the folder before writes land | Retirement split out, awaited, and refuses unless every document is confirmed encrypted |
| 5 | Backup restore trashed by the next scan | Mode check inside `saveToDisk()`, so every caller is covered |
| 6 | Legacy library lands in the Inbox | Read starts from `nil`; `verify()` compares the field |
| 7 | Typed properties collapse to text | Typed frontmatter — bare numerics, ISO dates, `true`/`false`, `.unknown` kept tagged |

**Your highest-leverage suggestion was right.** `verify()` now compares everything the file carries, and it turned three of the four bugs you predicted into verification failures rather than silent losses.

**One deliberate deviation:** you offered "reject a symlinked root" as an option and I resolved the link instead. Refusing would break the case you called most likely — pointing `~/Logue` at a vault — when the actual defect is that `enumerator(at:)` doesn't follow a symlinked root while `fileExists` does. A symlinked root now works, with a test asserting the document is found rather than trashed.

### Also fixed

`divergesFromDisk` treats unreadable as divergent · `prepareRoot` no longer reads an unreadable folder as empty · `retiringFolder: false` passed at the call site its documentation described · held-back edits queue a rescan · clearing data empties the markdown folder instead of leaving the library in plaintext · the Privacy tab reads the mode instead of contradicting itself · `goalMode` written · reserved-key shadowing closed.

**The four walks per scan are now one.** `FolderSnapshot` reads the folder once, off the main actor, and every step of the scan takes it — including the new cross-check, so that guard costs nothing extra. It also means every step sees the same folder rather than a slightly newer one, which was a race of its own. A test caught hidden I/O I had left in: path components are now frozen at walk time, because resolving them asks the filesystem.

### The #25 carry-overs

All three closed: `InlineAttributeVisitor` had the same grapheme-offset-as-`NSRange` pattern; meeting titles now go through a `MeetingStore.applyTitle` mirroring the document one, so the claim is true rather than narrowed; and both new defects — type renames orphaning documents, the properties panel committing on focus-out — are fixed.

### Still open

The thread on `DocumentStorage+Rescan.swift:210` stays open on purpose. The fix is in, but your explanation of *why no test caught it* still stands: both scan harnesses reimplement `deleteSpace` rather than driving `SpaceStore`, so production's destructive branch is unexercised. That is the same shape as the `WriteEchoFilter` problem — a test that reimplements what it tests cannot fail the way production does. The missing piece is the test, not the fix, so the thread is the record of it.

Two more things I have not done and would rather state than imply: **none of the UI in this branch has been exercised in the running app** — the saved-view and type menus, and the scan behaviour under a real watcher — and the split you suggested still looks right to me. #25 is at `fc8d12f` with its own review addressed, and merging it first would let the graph work land while this half takes another pass.

690 tests, lint clean.

Logue/Services/DocumentStorage+Rescan.swift
westerosweb: Closed now, and by removing the hazard rather than testing around it — `4901304`.

Your objection was sharper than the bug: pinning this call site with a test would have left the danger in place and watched it, and the reason no test caught it (harnesses reimplementing `deleteSpace`) would still have been true of whatever I wrote next.

So the folder now has to agree. `retireFolder(at:claimedBy:)` reads the `_space.md` in the directory it is about to trash and refuses unless it names the space being deleted. A folder that claims nothing — the user's recreated one — or claims a different space is left alone and logged. The flag is still there and still correct, but it is no longer what stands between someone and their folder, so the untested call site is no longer load-bearing.

Six tests at the migrator level, no singleton involved: claims-the-space is retired; recreated-with-the-same-name is not, and its contents survive; a different space's folder is not; an unclaimed folder is not; an already-missing folder is not an error; an empty path retires nothing.

The harness gap you identified is still real as a *general* observation — both scan harnesses still stand in for the stores — and I have not closed that, because making `SpaceStore` drivable in tests means injecting its storage path, which is a pre-existing singleton refactor rather than part of this review. What has changed is that no destructive behaviour now depends on it.

shanforge (approved):
## Approving — all seven blocking findings verified fixed

Re-reviewed at `4901304`. I checked each fix in the code rather than taking the commit messages for it, and every one holds.

| Finding | Fix |
| --- | --- |
| Symlinked or unreadable root trashed the whole library | `documentsStillOnDisk` cross-checks each candidate against the filesystem and routes survivors to `plan.unwalkable` instead of `plan.trashed`. Deliberately not a walk, which is the right call — it is the check *against* the walk that failed. |
| A file caught mid-write was adopted and its document trashed | `hasSettled(_:)` refuses to adopt a file that has changed within the settle window, and the on-disk cross-check independently stops the original being trashed. |
| Case-differing titles overwrote each other's files | `taken` is now folded to lowercase before matching. |
| Disable trashed the folder while writes were still pending | `retireFolderAfterReEncryption` runs after `await adoptAfterStorageSwitch` **and** `await awaitPendingSaves()`. |
| Restoring a backup in markdown mode was trashed by the next scan | `saveToDisk()` now branches on the storage mode, so the sample-data loader is covered too. |
| The legacy library was dumped into the Inbox | `content.storedIsOrganised = nil` before parsing, so absent means organised as the invariant requires. |
| Typed properties collapsed to text | Each value is written in a form that reads back as the same type, with ISO dates, and `verify()` now compares tags, icon, goal, pinned and the rest rather than three fields. |

Fixing `verify()` was the right lever. It was the reason four separate losses could pass as successful, and widening it is what makes the round trip trustworthy rather than merely tested.

Two things I liked in the fixes themselves. `documentsStillOnDisk` carries a comment explaining why it must not reuse the traversal, which is the sort of reasoning that stops someone "optimising" it back into the bug. And `modifiedAt` now comes from the file's own timestamp — I had not raised that, and it prevents an import stamping the whole library with the launch time.

Also confirmed fixed from the should-fix list: `goalMode` is written and verified, and "clear everything" now empties the markdown folder instead of leaving the library readable on disk while the app forgets it exists.

### One narrow gap, not blocking

`adopt(fileAt:)` still builds its content from `DroppedFileImport.Fields`, which carries only title, body and tags, and then rewrites the file. A note dropped in from Obsidian or Jekyll with `author:`, `aliases:` or `publish:` loses those keys to an in-place overwrite. The body and title survive and nothing is trashed, so this is metadata loss on a path the user opts into by hand — worth a follow-up rather than another round here, especially since `MarkdownDocumentFile.content(from:)` already preserves unrecognised keys as properties and could be reused.

CI is green on lint and build. Approving.

PR #32, opened 2026-07-28
Stop the app trapping when two speakers share a name

### LINKED ISSUE #31: App crashes after meeting generated speakers
### What happened?

After having a 15min meeting, the app started gathering the speakers. After getting them, I started renaming one of the speakers (Speaker 1). Let’s say it was “John”. It picked up John as 2 separate speakers, so I renamed Speaker 2 also to John. As soon as I saved, the app crashed, and every time I try to open that meeting, the app keeps crashing. 
I cannot provide the recording because it has company info, but I’m happy to help test out this in different scenarios.

### Steps to reproduce

1. Record a meeting
2. Rename 2 speakers with the same name
3. Try to reopen the meeting after the app crashes

### Logue version

1.0.0

### macOS version

Version 26.5.1 (Build 25F80)

### Mac model

Mac 14,2

### Screenshots or logs

_No response_

westerosweb:
Fixes #31.

## The crash

`MeetingWorkspaceView` built the transcript's speaker colour map keyed by speaker *name*:

```swift
speakerColors: Dictionary(
    uniqueKeysWithValues: meeting.speakers.map { ($0.name, $0.displayColor) }
)
```

`Dictionary(uniqueKeysWithValues:)` traps on a duplicate key, so two speakers named "John" raise
`Fatal error: Duplicate values for key`. Because the expression sits in a view body it trapped both
when the rename was saved and on every subsequent attempt to open the meeting — which is why the
reporter's meeting stayed unopenable.

Keying with `uniquingKeysWith:` fixes it. Being a read-side change, **meetings already broken by this
bug open again with no migration.**

## The rename now means what the user meant

From the report: diarization split one person across two speaker rows, so the reporter renamed the
second row to the same name to put them back together. That is a merge, and it is now treated as one
— transcript segments are relabelled, speaker segments are re-attributed to the surviving speaker,
and the redundant speaker row is dropped. Renaming to an unused name behaves exactly as before.

## Deduplication

Two call sites each carried their own copy of the rename loop and could drift apart:
`MeetingSpeakersPanelView.renameSpeaker(from:to:)` and an inline closure in `MeetingWorkspaceView`.
Both now share `MeetingNote.renamingSpeaker(from:to:)`, which is pure and testable without
`MeetingStore`.

Also verified as *not* affected: `MeetingTools.swift` and `MeetingStore+Diarization.swift` key their
speaker dictionaries by `id`, and `MeetingSpeakersPanelView`'s `ForEach(speakerStats, id: \.name)`
iterates statistics already grouped by name, so its keys are unique by construction.

## Tests

New `LogueTests/SpeakerRenameTests.swift` — 13 tests covering plain rename, id stability, the merge
path (relabelling, segment re-attribution, payload preservation, merging into a segmentless
speaker), the duplicate-name colour map that previously trapped, and the no-op cases (blank name,
whitespace trimming, unchanged name, unknown speaker).

## Verification

- `xcodegen generate` + build: succeeded
- Full non-LLM test suite: 444 tests in 45 suites passed
- SwiftFormat and SwiftLint: clean, no new suppressions

## Follow-up

A companion change recovering transcript-to-speaker alignment quality is in progress on a separate
branch; the design for both is in `docs/specs/2026-07-28-diarization-recovery-design.md`,
included here. That work also documents a resumed-recording transcript-loss bug found along the way.


Logue/Views/Meeting/MeetingWorkspaceView.swift
```diff
@@ -123,14 +123,7 @@ struct MeetingWorkspaceView: View {
...
                         onRenameSpeaker: { oldName, newName in
-                            var updated = meeting
-                            for index in updated.segments.indices where updated.segments[index].speakerLabel == oldName {
-                                updated.segments[index].speakerLabel = newName
-                            }
-                            for index in updated.speakers.indices where updated.speakers[index].name == oldName {
-                                updated.speakers[index].name = newName
-                            }
-                            store.updateMeeting(updated)
+                            store.updateMeeting(meeting.renamingSpeaker(from: oldName, to: newName))
```
shanforge: Nit — this call site lost the early-out that its sibling kept.

`MeetingSpeakersPanelView.renameSpeaker` guards with `guard !trimmed.isEmpty, trimmed != oldName else { return }` and the new comment there says why: *"Skip the persist entirely when the field was committed unchanged."* This one has no guard.

`renamingSpeaker` returning `self` keeps it correct, but `updateMeeting` doesn't compare — it unconditionally sets `modifiedAt = Date()`, invalidates caches and calls `saveMeeting`. So committing a speaker field unchanged from the transcript view rewrites the meeting and bumps it to the top of anything sorted by modified date. Tabbing out of a rename field without typing is a normal thing to do, so this isn't rare.

Mirroring the panel's guard here would make the two paths agree, which is the point of the deduplication.

Logue/Models/MeetingNote+Speakers.swift
```diff
@@ -0,0 +1,62 @@
...
+    /// name: transcript segments are relabelled, speaker segments are remapped onto the surviving
+    /// speaker, and the redundant speaker row is dropped. That is what renaming to an existing name
+    /// means in practice — diarization split one person in two and the user is putting them back
+    /// together.
+    ///
+    /// Returns `self` unchanged when `newName` is blank, unchanged from `oldName`, or when no
+    /// speaker is named `oldName`.
+    func renamingSpeaker(from oldName: String, to newName: String) -> MeetingNote {
+        let trimmed = newName.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty, trimmed != oldName else { return self }
```
shanforge: Worth a line in the doc comment: this guard means a meeting *already* holding two rows named "John" can't be repaired by the obvious action.

Renaming one of them to "John" hits `trimmed != oldName` and returns `self`, so nothing merges. The user has to rename it to something else and then back to "John" for the merge path to fire — two steps, and there's nothing in the UI hinting at that.

That's a smaller deal than it sounds, because the crash fix means those meetings now *open*, which is what the reporter actually needs. But the description says they're recovered "with no migration", and strictly they're recovered in a state the user still can't tidy up in one move. Either a sentence in the doc comment or a follow-up issue would close the gap between what's claimed and what happens.

docs/specs/2026-07-28-diarization-recovery-design.md
```diff
@@ -0,0 +1,378 @@
+# Speaker Diarization Recovery and Duplicate-Name Crash Fix
```
shanforge: Read this properly rather than skimming it, and it's good — the scope decisions in particular are the useful kind, because they record what was *rejected* and why. "Hardcoded two-speaker collapsing is out of scope" with the reasoning about three- and four-person meetings is exactly the note that stops someone re-introducing it in six months.

One observation, not a request: Part 1 documents this PR, but Part 2 is design for work landing on another branch, and it's the larger half. That means whoever reviews this crash fix has to decide whether they're also implicitly signing off on the alignment rewrite. Given the bug currently makes meetings unopenable, there's an argument for letting the fix merge on its own and bringing the design in with the work it describes.

Equally there's an argument for landing the reasoning now while it's written. Your call — I don't think either is wrong.

Separately, noting that the "No identifying information carries over" paragraph is doing real work and I checked it: no `DEVELOPMENT_TEAM`, personal names, branch names, or `Co-authored-by` trailers from the closed branches appear anywhere in this diff.

shanforge (approved):
Reviewed this locally rather than from the description — checked out the branch, built it, ran the tests, and verified the load-bearing claims independently. The fix is right and the reasoning in the description holds up. A few non-blocking notes inline.

**What I verified**

- `Dictionary(uniqueKeysWithValues:)` traps on a duplicate key by definition, so the diagnosis is exact. Being read-side, meetings already persisted with duplicate names genuinely do open again with no migration — that's the part that matters most for the reporter.
- I grepped every `uniqueKeysWithValues` in the tree to check the "not affected" claim rather than take it. It holds: `MeetingTools.swift:229` and `MeetingStore+Diarization.swift:14` both key by `id`. No trap site is left.
- `Speaker.displayColor` is stored RGB per speaker, not derived from array position, so dropping a row during a merge doesn't shift any surviving speaker's colour. Worth stating explicitly because it's the kind of thing that would only surface as a confusing screenshot later.
- 13/13 new tests pass. SwiftFormat 0.62.1 and SwiftLint 0.65.0 — the CI-pinned versions — are both clean.

**Heads-up: CI has never run on this branch.** `gh run list --branch fix/duplicate-speaker-name-crash` returns nothing, and `gh api .../check-runs` on `07f8795` is empty, even though `ci.yml` triggers on `pull_request: branches: [main]`. Actions is working generally — it ran five times on my branch yesterday. So the build/lint/test claims in the description have no CI evidence behind them; I've verified them locally instead, but it's worth an empty push to get a real run before merge, and worth understanding why it didn't fire.

The merge-on-duplicate-name behaviour is the right call. Renaming a row to a name that already exists has no other sensible meaning once you know diarization split one person in two, and treating it as a no-op or allowing two rows to share a name is what got us here.

Logue/Views/Meeting/MeetingWorkspaceView.swift
westerosweb: Fixed in 3fc8df0. Extracted `renameSpeaker(in:from:to:)` onto `MeetingWorkspaceView` with the same guard and the same comment as the panel's, and the callback now goes through it.

Confirmed the mechanism before changing anything — `MeetingStore.updateMeeting` really does stamp `modifiedAt`, `invalidateCaches()` and `saveMeeting` with no comparison, so the no-op commit was a genuine write, not just a redundant call.

Logue/Models/MeetingNote+Speakers.swift
westerosweb: Documented in 3fc8df0 — but the two-step workaround doesn't actually work, and the gap is wider than the comment describes.

I wrote the rename-away-and-back case as a test expecting it to merge, and it failed. The reason is that renames are keyed by *name*, not by speaker id:

```swift
for index in updated.speakers.indices where updated.speakers[index].name == oldName {
    updated.speakers[index].name = trimmed
}
```

With two rows both named "John", renaming "John" → "Temp" moves **both** rows. Renaming "Temp" → "John" then brings both back. The merge branch is never reached in either direction, so there is no rename-only repair for an already-broken meeting — not a two-step one either.

The doc comment now says that rather than the softer version, and two tests pin it so it can't drift:

- `retypingSharedNameDoesNotMerge` — the early-out case you spotted
- `renameAwayAndBackStillDoesNotMerge` — asserts `["Temp", "Temp"]` then `["John", "John"]`

Collapsing those rows needs rename to target a speaker id instead of a name, which changes the callback signatures in both the panel and the transcript view. That felt like more than a crash fix should carry, so I've left it out — happy to open a follow-up issue for the id-based rename if you'd like it tracked.

westerosweb: Follow-up: fixed properly in dcd8d29 rather than left as a doc comment.

I went to make rename id-based and found it can't be driven from the UI. `MeetingSpeakersPanelView` builds `speakerStats` from `segments` grouped by `speakerLabel` and iterates `ForEach(speakerStats, id: \.name)` — so two `Speaker` rows named "John" render as **one** panel row. There is no per-row affordance for an id to address, and exposing one would mean listing `meeting.speakers` directly, which changes what the panel shows (including rows with no segments).

So the fix is to collapse the rows instead of trying to address them:

```swift
extension MeetingNote {
    func collapsingDuplicateSpeakers() -> MeetingNote
}
```

Rows sharing a name merge onto the first of the group, and the absorbed rows' `SpeakerSegment` entries are re-attributed to the survivor. It runs in two places:

- **At the end of `renamingSpeaker`**, which made the merge fall out of the rename rather than being a separate branch — the special-cased survivor lookup and `absorbedIDs` removal are gone. The unchanged-name early-out is relaxed for exactly one case, retyping a name two rows already share, since that's the only way the user can express the repair.
- **On the decode path in `MeetingStore+Persistence`**, so an affected meeting is repaired when it loads instead of waiting for a rename nobody would think to try.

18/18 tests pass. The eleven pre-existing rename tests pass unchanged, so the restructuring is behaviour-preserving everywhere else; the five new ones cover the helper and both rename routes into it.

docs/specs/2026-07-28-diarization-recovery-design.md
westerosweb: Split it — 2b537b5. The crash fix is the part making meetings unopenable, so it should be reviewable without also signing off on the alignment rewrite.

- `docs/specs/2026-07-28-duplicate-speaker-name-crash.md` stays here
- `docs/specs/2026-07-28-diarization-alignment-recovery.md` moved to #33, where the code lands

Each links to the other, and the shared Background and Scope Decisions travel with the alignment doc since that's what they're about. The Testing section split along the same line.

One thing that turned up doing it: the spec was only in the repo because it had been force-added. `.gitignore` had a blanket `docs/` rule, and git doesn't descend into an excluded directory, so every `!docs/...` negation under it was unreachable — the new spec would have been silently dropped. Changed to `docs/*` with `!docs/specs/`, which keeps the same intent and actually works. Same change on both branches, so it merges cleanly.

PR #33, opened 2026-07-28
Attribute transcript segments to speakers by overlap, not midpoint

westerosweb:
Recovers the transcription/diarization work that was in flight before the repo was published and
never landed here. Design doc ships in #32 (`docs/specs/2026-07-28-diarization-recovery-design.md`).

## Why speaker labels are wrong today

Labels are assigned by taking each transcript segment's **midpoint** and finding the nearest speaker
segment within **3.5 seconds**. A midpoint carries no information about how much of a segment each
speaker holds, so a brief interjection landing mid-sentence captures the whole sentence — and a 3.5s
nearest-neighbour window will label a segment from a speaker who was not talking during it at all.

Raw Sortformer output makes it worse: sub-second slivers, single turns split across segments, and
rapid alternations, all fed straight into that matcher. This is the fragmentation the earlier
benchmark work observed (three speakers reported for a two-speaker recording).

## What changed

**`SpeakerAlignment`** (new, pure) — overlap-duration voting; segments over 2s chunked with each
slice voting weighted by duration; a near-tie keeps the previous speaker instead of flipping on a
marginal difference; segments genuinely spanning two speakers split at the boundary with text
distributed proportionally; brief islands smoothed; fallback tightened from 3.5s to 0.5s and applied
only when nothing overlaps.

**`SortformerTimeline`** (new, pure) — filters fragments, merges same-speaker runs, collapses rapid
alternations, closes up overlaps and gaps, before any of it becomes speaker data.

Both live outside the store, so 34 unit tests cover them with no models and millisecond runtime.

## Deliberately not ported

The earlier branches both hardcode a two-speaker world — one remaps everyone outside the top two by
duration onto them, the other collapses any third speaker into the nearest of two. On a real
three- or four-person meeting that destroys correct output. Every step here is speaker-count
agnostic, and two tests exist purely to guard against reintroducing it.

Also skipped: replacing Apple Speech with a streaming FluidAudio engine. Batch Parakeet already
overwrites the final transcript after stop, so that change only affects the live preview, at the cost
of a second model download and ~600 lines of heuristics on the recording hot path.

Two defects in the source branch are documented in the design doc rather than ported: a
word-distribution loop that never terminates when there are fewer words than parts (would hang the
post-recording pipeline), and an ordering clamp that silently deletes any transcript segment under
0.6 seconds. Its `continuityThreshold` guard is unreachable — the previous speaker's overlap can
never exceed the runner-up's — so one guard is implemented, not two.

## Transcript loss on resumed recordings

Found while comparing the branches, fixed here. Both diarizers and batch ASR time their output from
the start of the session's own audio buffer, but `replaceTranscript` **assigned** over the whole
segment array. Recording again into an existing meeting therefore discarded everything transcribed
earlier and misaligned what remained. The session offset is threaded through and only that session's
segments are replaced.

## Reliability and cleanup

Stopping while models were still downloading left Sortformer not streaming, silently degrading to the
batch path. Stop now waits for initialization, capped by `diarizationInitStopWait` so a slow first
run cannot hold up post-recording AI.

Removes the live/periodic diarization path, which has had no callers for some time:
`startPeriodicProcessing`, `stopPeriodicProcessing`, `processIncrementalChunk`,
`processRemainingChunk`, `applyDiarizationResult`, their state, and the two flags gating them.
Comments describing live speaker updates and a `processCompleteRecording()` call site now match what
the pipeline actually does.

No preload work is included — `prewarmGlobalCache()` already warms Sortformer, ASR, and VAD at
launch, which is more than the source branch's Sortformer-only preload.

## Verification

- Build: succeeded
- Full non-LLM suite: **465 tests in 46 suites passed** (was 444 before the two new suites)
- SwiftFormat and SwiftLint: clean, no new suppressions — `awaitDiarizationInit` was placed in the
  diarization extension rather than suppressing a `file_length` violation
- The three guards carrying the most judgement (ambiguity check, island-duration bound, splitting)
  were each mutation-verified: disabling the guard fails the specific test written for it

## What still needs a human

Diarization accuracy cannot be measured from code review. The algorithms behave as specified on
constructed inputs, but **whether this measurably improves labelling on real audio needs a
multi-speaker recording** — worth checking before merge, ideally on the kind of meeting that produced
the original complaint.

### DIFF
--- a/Logue/Engine/SortformerTimeline.swift
+++ b/Logue/Engine/SortformerTimeline.swift
@@ -90,8 +90,15 @@ enum SortformerTimeline {
 
     // MARK: - Continuity
 
-    /// Removes overlaps and oversized gaps so the timeline reads as one continuous conversation.
-    /// A segment entirely swallowed by its predecessor is dropped rather than inverted.
+    /// Removes overlaps, and closes gaps narrow enough to be model jitter, so the timeline reads as
+    /// one continuous conversation. A segment entirely swallowed by its predecessor is dropped
+    /// rather than inverted.
+    ///
+    /// A silence wider than `maxClosableGap` is left exactly as it is: closing it would move the
+    /// later speaker's start back across time nobody was confidently speaking in, and alignment
+    /// compares true transcript times against these adjusted ones. A stray cough or a missed word
+    /// in that window would then be labelled with full confidence instead of honestly falling
+    /// outside every speaker.
     private static func enforceContinuity(
         _ updates: [SortformerSpeakerUpdate]
     ) -> [SortformerSpeakerUpdate] {
@@ -104,10 +111,13 @@ enum SortformerTimeline {
                 continue
             }
 
+            let gap = update.startTime - last.endTime
             var start = update.startTime
-            if start < last.endTime {
+            if gap < 0 {
                 start = last.endTime
-            } else if start - last.endTime > AppConstants.Diarization.timelineMaxGap {
+            } else if gap > AppConstants.Diarization.timelineMaxGap,
+                      gap <= AppConstants.Diarization.maxClosableGap
+            {
                 start = last.endTime + AppConstants.Diarization.timelineMaxGap
             }
             guard update.endTime > start else { continue }
@@ -119,8 +129,12 @@ enum SortformerTimeline {
                 endTime: update.endTime
             )
 
-            // Clamping can bring two same-speaker segments flush together.
-            if last.speakerIndex == adjusted.speakerIndex {
+            // Clamping can bring two same-speaker segments flush together. Only merge when they
+            // genuinely are: a silence left open above must not be swallowed by joining the speaker
+            // to themselves across it.
+            if last.speakerIndex == adjusted.speakerIndex,
+               adjusted.startTime - last.endTime <= AppConstants.Diarization.timelineMaxGap
+            {
                 fixed[fixed.count - 1] = SortformerSpeakerUpdate(
                     speakerIndex: last.speakerIndex,
                     speakerName: last.speakerName,
--- a/Logue/Engine/SortformerTimeline.swift
+++ b/Logue/Engine/SortformerTimeline.swift
@@ -1,0 +1,142 @@
+import Foundation
+
+/// Cleans up raw Sortformer speaker output before it becomes persisted speaker data.
+///
+/// Sortformer emits a fragmented timeline: sub-second slivers, the same speaker split across
+/// consecutive segments, brief alternations where one speaker's turn is interrupted and resumed,
+/// and segments that overlap their neighbours. Feeding that straight into transcript alignment
+/// produces speaker labels that flip mid-sentence.
+///
+/// Every step here is speaker-count agnostic. Nothing collapses the timeline towards a target
+/// number of speakers, because a three- or four-person meeting is as valid as a two-person one.
+enum SortformerTimeline {
+    /// Filters fragments, merges same-speaker runs, collapses rapid alternations, and closes up
+    /// overlaps and gaps. Returns segments ordered by start time.
+    static func normalize(_ updates: [SortformerSpeakerUpdate]) -> [SortformerSpeakerUpdate] {
+        guard !updates.isEmpty else { return [] }
+
+        let ordered = updates
+            .filter { $0.endTime - $0.startTime >= AppConstants.Diarization.minSpeakerSegmentDuration }
+            .sorted { $0.startTime < $1.startTime }
+
+        let merged = mergeAdjacentSameSpeaker(ordered)
+        let collapsed = collapseRapidAlternations(merged)
+        return enforceContinuity(mergeAdjacentSameSpeaker(collapsed))
+    }
+
+    // MARK: - Merging
+
+    /// Joins consecutive segments attributed to the same speaker when the silence between them is
+    /// shorter than `sameSpeakerMergeGap` — one continuous turn Sortformer reported in pieces.
+    private static func mergeAdjacentSameSpeaker(
+        _ updates: [SortformerSpeakerUpdate]
+    ) -> [SortformerSpeakerUpdate] {
+        updates.reduce(into: [SortformerSpeakerUpdate]()) { merged, update in
+            guard let last = merged.last,
+                  last.speakerIndex == update.speakerIndex,
+                  update.startTime - last.endTime <= AppConstants.Diarization.sameSpeakerMergeGap
+            else {
+                merged.append(update)
+                return
+            }
+            merged[merged.count - 1] = SortformerSpeakerUpdate(
+                speakerIndex: last.speakerIndex,
+                speakerName: last.speakerName,
+                startTime: last.startTime,
+                endTime: max(last.endTime, update.endTime)
+            )
+        }
+    }
+
+    // MARK: - Alternation collapsing
+
+    /// Rewrites an A-B-A run that completes inside `alternationWindowSeconds` as a single segment
+    /// belonging to whichever speaker holds more of it. Three turns in under a second is the model
+    /// wavering, not two people talking.
+    private static func collapseRapidAlternations(
+        _ updates: [SortformerSpeakerUpdate]
+    ) -> [SortformerSpeakerUpdate] {
+        guard updates.count >= 3 else { return updates }
+
+        var collapsed = updates
+        var index = 0
+        while index + 2 < collapsed.count {
+            let first = collapsed[index]
+            let middle = collapsed[index + 1]
+            let last = collapsed[index + 2]
+
+            let spansWindow = last.endTime - first.startTime <= AppConstants.Diarization.alternationWindowSeconds
+            guard first.speakerIndex == last.speakerIndex,
+                  first.speakerIndex != middle.speakerIndex,
+                  spansWindow
+            else {
+                index += 1
+                continue
+            }
+
+            let flankDuration = duration(of: first) + duration(of: last)
+            let dominant = flankDuration >= duration(of: middle) ? first : middle
+            collapsed.replaceSubrange(index ... index + 2, with: [SortformerSpeakerUpdate(
+                speakerIndex: dominant.speakerIndex,
+                speakerName: dominant.speakerName,
+                startTime: first.startTime,
+                endTime: last.endTime
+            )])
+            // Step back so a longer alternating run collapses from the inside out.
+            index = max(0, index - 1)
+        }
+        return collapsed
+    }
+
+    // MARK: - Continuity
+
+    /// Removes overlaps and oversized gaps so the timeline reads as one continuous conversation.
+    /// A segment entirely swallowed by its predecessor is dropped rather than inverted.
+    private static func enforceContinuity(
+        _ updates: [SortformerSpeakerUpdate]
+    ) -> [SortformerSpeakerUpdate] {
+        var fixed: [SortformerSpeakerUpdate] = []
+        fixed.reserveCapacity(updates.count)
+
+        for update in updates {
+            guard let last = fixed.last else {
+                fixed.append(update)
+                continue
+            }
+
+            var start = update.startTime
+            if start < last.endTime {
+                start = last.endTime
+            } else if start - last.endTime > AppConstants.Diarization.timelineMaxGap {
+                start = last.endTime + AppConstants.Diarization.timelineMaxGap
+            }
+            guard update.endTime > start else { continue }
+
+            let adjusted = SortformerSpeakerUpdate(
+                speakerIndex: update.speakerIndex,
+                speakerName: update.speakerName,
+                startTime: start,
+                endTime: update.endTime
+            )
+
+            // Clamping can bring two same-speaker segments flush together.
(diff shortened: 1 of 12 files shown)


Logue/Engine/SpeakerAlignment.swift
```diff
@@ -0,0 +1,313 @@
...
+        var nearestDistance = TimeInterval.greatestFiniteMagnitude
+
+        for speakerSegment in speakerSegments {
+            if speakerSegment.startTime > end + AppConstants.Diarization.labelFallbackTolerance {
+                break
+            }
+            guard let name = speakerNamesByID[speakerSegment.speakerId] else { continue }
+
+            let overlap = min(end, speakerSegment.endTime) - max(start, speakerSegment.startTime)
+            if overlap > bestOverlap {
```
shanforge: **🔴 One speaker's two segments count as rivals, so the ambiguity guard hands the segment to the wrong person.**

`bestOverlap` / `runnerUpOverlap` track the top two *speaker segments*, not the top two *speakers*. When one speaker holds several segments across the interval — which is precisely what an interjection produces — that speaker becomes its own runner-up, `isTooCloseToCall` fires, and the segment falls through to `previousLabel`.

I reproduced it against this branch:

```swift
// Alice 0.0–1.0, Bob 1.0–1.2, Alice 1.2–2.0
// transcript segment 0.0–1.9, previous segment labelled "Bob"
```

Alice holds **1.7s**, Bob holds **0.2s**. Result: **Bob**.

Remove the preceding Bob-labelled segment so `previousLabel` is nil, and the same input yields **Alice**. So the overlap maths is right and the guard is what breaks it: walking in start order gives best = Alice(1.0), runner-up = Alice(0.7), and `0.7 >= 1.0 * 0.70` is true.

Splitting does not save it either — `speakerRanges` yields three ranges, but the middle part is 0.2s, under `minSplitPartDuration` (0.4s), so `splitParts` correctly returns nil and the whole segment goes to the vote.

This is the case the PR description opens with: *"a brief interjection landing mid-sentence captures the whole sentence"*. It still does — via a different route.

**Fix:** accumulate overlap per speaker before ranking, which is what `vote()` already does for chunked segments:

```swift
var overlapByName: [String: TimeInterval] = [:]
// … overlapByName[name, default: 0] += overlap
let ranked = overlapByName.sorted { $0.value > $1.value }
```

Then best/runner-up are genuinely two different speakers and `isTooCloseToCall` means what it says.

Worth a test for the interjection shape — the suite covers near-ties between two *distinct* speakers (line 43) but never one speaker on both sides of another, which is why this got through.

Logue/Engine/SortformerTimeline.swift
```diff
@@ -0,0 +1,142 @@
...
+        for update in updates {
+            guard let last = fixed.last else {
+                fixed.append(update)
+                continue
+            }
+
+            var start = update.startTime
+            if start < last.endTime {
+                start = last.endTime
+            } else if start - last.endTime > AppConstants.Diarization.timelineMaxGap {
```
shanforge: **🟠 Silence is compressed without bound, so a speaker segment ends up covering time that speaker was silent for.**

A gap wider than `timelineMaxGap` is closed by moving the *later* segment's start earlier. The shift is as large as the pause, and meetings have long pauses.

Run against this branch:

```
in:  Speaker 1  0.0–10.0 · Speaker 2  30.0–40.0
out: Speaker 1  0.0–10.0 · Speaker 2  10.5–40.0
```

Speaker 2 now claims **19.5 seconds of silence** they had nothing to do with.

That matters because alignment compares *true* transcript times against these *adjusted* speaker times. Anything transcribed in that window — quiet speech Sortformer missed, a cough, an ASR hallucination — now overlaps Speaker 2 and is labelled with full confidence. Without the stretch it would fall outside the 0.5s fallback and stay unlabelled, which is the honest answer for a stretch of audio nobody was confidently speaking in.

I can see the intent: small gaps are model jitter and closing them helps. But the rule does not distinguish a 0.6s jitter gap from a 20s pause, and it is the second case that invents data.

Worth bounding it — close gaps only up to some multiple of `timelineMaxGap`, and leave genuinely long silences as silence. If closing arbitrary gaps is deliberate, the doc comment should say so, because "closes up overlaps and gaps" reads as tidying rather than as extending a speaker over half a minute they were quiet for.

Logue/Services/MeetingStore+Diarization.swift
```diff
@@ -3,68 +3,53 @@ import Foundation
...
+    /// (Parakeet TDT). Called after recording stops to upgrade streaming transcript quality.
+    ///
+    /// `sessionStart` is the meeting timeline offset the session began at. Batch ASR timings are
+    /// relative to the session's own audio buffer, so they are shifted onto the meeting timeline and
+    /// only segments from this session are replaced — otherwise resuming a recording would discard
+    /// every earlier session's transcript.
+    func replaceTranscript(
+        for meetingID: UUID,
+        with segments: [TranscriptSegment],
+        sessionStart: TimeInterval = 0
```
shanforge: **🟡 The parameter that prevents transcript loss defaults to the value that causes it.**

`sessionStart: TimeInterval = 0` means a caller who forgets it gets `earlierSessions = filter { $0.startTime < 0 }` — empty — and the function silently reverts to replacing the entire array. That is exactly the bug this PR fixes.

The default is doing backward-compatibility work that nothing needs: I checked, and the only caller passes it explicitly. Same for the three defaults in `RecordingSessionManager+Diarization` (`processDiarization`, `mergeDiarizationResult`, `applySortformerUpdates`).

Making it required costs one argument at each call site and makes the failure a compile error instead of a silent data loss that only shows up on a resumed recording — the exact scenario least likely to be exercised while developing.

The fix itself is right, incidentally. I traced `timeOffset = lastEnd + 1.0` through the filter and the partition is exact: every earlier segment starts strictly below the offset, every shifted one at or above it.

Logue/Engine/SpeakerAlignment.swift
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    // MARK: - Smoothing
+
+    /// Rewrites a lone segment flanked on both sides by the same other speaker.
+    ///
+    /// Bounded to islands shorter than `maxSmoothedIslandDuration`: a two-second turn is somebody
+    /// speaking, and overriding it would delete a real contribution. Only brief islands are treated
+    /// as diarization flapping. Protected segments are never rewritten.
+    private static func smoothIslands(
```
shanforge, line 312: **🟡 Question: can smoothing undo a split that was just made?**

`splitParts` produces new `TranscriptSegment`s, so they carry fresh UUIDs and are absent from `protectedIDs`. `smoothIslands` therefore treats them as ordinary segments.

A split part must be at least `minSplitPartDuration` (0.4s), and smoothing rewrites islands up to `maxSmoothedIslandDuration` (0.8s) — so a part between 0.4s and 0.8s flanked by the same speaker on both sides looks exactly like flapping and gets relabelled.

If that fires, the two heuristics disagree and the later one wins: splitting decided "two speakers genuinely span this segment", smoothing then decides "no, that was noise". The split boundary is the better-informed judgement of the two, since it came from actual speaker-segment boundaries rather than from what the neighbours happen to be.

I have **not** confirmed this triggers in practice — it needs the durations and word counts to line up and I did not construct a case. Flagging it as a question rather than a finding. If it can happen, adding freshly-split parts to `protectedIDs` before smoothing would settle it.

shanforge (changes requested):
Reviewed this by checking out the branch, building it, running the suites, and probing the algorithms with constructed inputs rather than reading them and reasoning. The structure is right and most of it holds up — but **one defect reproduces in the exact scenario this PR leads with**, so I'm requesting changes rather than approving.

## Verified

- Build clean; **34 tests in the two new suites pass**, matching the description
- SwiftFormat and SwiftLint clean at the CI-pinned versions
- No dangling references to the removed live-diarization API — I grepped for all six removed symbols across `Logue/` and `LogueTests/`, none remain
- The resumed-recording fix is sound. I traced `timeOffset = lastEnd + 1.0` through to `replaceTranscript`'s `filter { $0.startTime < sessionStart }` — every earlier segment is strictly below the offset, so the partition is exact and the shifted segments land contiguously
- Pulling the transcript replacement *before* the Sortformer apply, so alignment runs against final text rather than draft, is a real improvement and easy to miss in review

The "deliberately not ported" section is the best part of this. Declining the two-speaker collapse, and guarding it with tests, is the right call — and documenting the two source-branch defects (the non-terminating word loop, the sub-0.6s clamp) rather than porting them is exactly what should happen to code recovered from a stale branch.

## Why I'm requesting changes

The blocking one is **`SpeakerAlignment.winner()` treating a single speaker's multiple overlapping segments as rival speakers**, which makes the ambiguity guard misfire and hand the segment to the wrong person. It reproduces on an interjection — the case the PR opens with. Detail inline.

Two smaller things inline as well: an unbounded silence compression that fabricates speaker coverage, and `sessionStart` defaulting to `0` in the functions whose whole purpose is to stop a caller losing earlier transcript.

## On what still needs a human

Agreed, and worth repeating: none of this measures accuracy. My probes show the algorithms do what the code says on constructed inputs — they cannot tell you whether labelling improves on real audio. The multi-speaker recording check before merge is the right gate.

Logue/Engine/SpeakerAlignment.swift
westerosweb: Fixed in 90ddf65. `winner()` now totals overlap per speaker before ranking, so best and runner-up are two different people — the shape you suggested, matching what the chunked path was already doing a level up. Ranking also breaks exact ties on name, since dictionary order is not stable and a coin-flip label should not vary between runs.

One correction worth recording, because it changes what the regression test has to look like. Your repro is real, but it sits within floating-point noise of the threshold, and a nearby variant of it does **not** reproduce. My first attempt at the test used Alice 1.0–2.0 / Bob 2.0–2.2 / Alice 2.2–2.9 against a transcript segment 1.0–2.9 — the same 1.0 / 0.2 / 0.7 overlaps as yours — and it passed against unfixed code. Instrumenting `winner()` showed why:

```
WINNER start=1.0 end=2.9 best=1.0 runnerUp=0.6999999999999997 tooClose=false
```

`2.9 - 2.2` lands at 0.6999999999999997, just under `1.0 * 0.70`, where your `1.9 - 1.2` lands at 0.7000000000000002, just over. The guard fires or not on the last bit of a double.

So the test now uses overlaps of 1.0 and 0.8 against 0.2 — comfortably inside the ambiguous band rather than balanced on its edge, and the 0.2s middle still below `minSplitPartDuration` so splitting declines and the vote decides the segment alone, which is the path you identified. Verified both ways: against `HEAD~1` it fails with `[Bob, Bob]`, and passes with the fix.

Logue/Engine/SortformerTimeline.swift
westerosweb: Fixed in 90ddf65. Gaps are now closed only up to `maxClosableGap` (2.0s, 4× `timelineMaxGap`); wider than that the pause is conversation rhythm and the silence is left exactly as it is. Your input is now a test — it produced `10.5` before, `30.0` after.

Bounding it exposed a second route to the same fabrication, which I would have shipped otherwise. The same-speaker merge below the clamp had no gap condition, because until now nothing could reach it with a gap wider than 0.5s. Leaving long silences open meant `Speaker 1 0–10 · Speaker 1 30–40` merged into a single `0–40` segment — worse than the original bug, since it claims the silence outright rather than only its front half. The merge now requires the two to actually be flush. Also a test, and it fails against `HEAD~1` with `count == 1, endTime == 40`.

On the doc comment: rewritten. "Closes up overlaps and gaps" was doing exactly the smoothing-over you describe, so the comment now says which gaps are closed, which are not, and why the distinction matters to alignment rather than just to the timeline.

Logue/Services/MeetingStore+Diarization.swift
westerosweb: Fixed in 90ddf65. All four are required now — `replaceTranscript` plus the three in `RecordingSessionManager+Diarization`. Every call site already passed it explicitly, so it cost nothing at the call sites and the failure mode is now a compile error. The doc comment on `replaceTranscript` says why the parameter has no default, so it does not get one added back as a convenience.

Logue/Engine/SpeakerAlignment.swift
westerosweb: It happens. I built the case you described rather than leaving it as a question: a transcript segment 0–3.0 over Alice 0–1.2 / Bob 1.2–1.8 / Alice 1.8–3.0. Splitting produces three parts with a 0.6s middle — above `minSplitPartDuration`, below `maxSmoothedIslandDuration`, same speaker either side — and against unfixed code the result is `[Alice, Alice, Alice]`. Smoothing wins, exactly as you predicted, and the split boundary is gone.

Fixed the way you suggested: freshly-split parts join `protectedIDs` before smoothing. They are identifiable without changing `split()`'s signature — after splitting, a segment carrying a label that was not protected on the way in is by construction a split part, since originals were either protected or still unlabelled at that point. Reasoning recorded in the comment so the derivation is not something the next reader has to reconstruct.

shanforge (approved):
LGTM, thanks!

PR #36, opened 2026-07-29
Fix/live dictation locale fallback

### LINKED ISSUE #34: Live transcription falls back to English for system languages not in SpeechTranscriber

Live meeting captions use Apple `SpeechTranscriber` only. On macOS 26, that module’s `supportedLocales` often omits languages that system Dictation supports (confirmed: Russian / `ru-RU` is missing from `SpeechTranscriber` but present in `DictationTranscriber`).

When Auto/system language is Russian:

1. **Locale resolution bug:** Auto used `Locale.current`. For an English-only app bundle this stays `en` even when macOS Preferred Languages are `ru-RU`, so live speech starts as English.
2. **No Dictation fallback:** Even with the correct `ru-RU` locale, unsupported Speech locales silently fell back toward English `SpeechTranscriber` instead of using `DictationTranscriber`.

Result: live captions are English/gibberish during recording, while post-recording Parakeet batch ASR already produces a usable Russian transcript. Users cannot get correct **live** captions for Russian (and likely other Dictation-only locales) without a code change.

This makes real-time notes unusable for non–SpeechTranscriber languages despite on-device Dictation support on the same Mac.

### Proposed solution

1. **Resolve Auto from the system language list** (`Locale.preferredLanguages` / mapped `TranscriptionLanguage`), not `Locale.current` (app localization locale).
2. **Engine selection for live speech:**
   - Prefer `SpeechTranscriber` when the locale is in its `supportedLocales` (keep long-form quality for en/de/fr/…).
   - Otherwise use `DictationTranscriber` for that locale (e.g. `ru-RU`) with long-form-friendly options (`volatileResults`, `frequentFinalization`, `farField` as appropriate).
   - Do **not** silently substitute English Speech when Dictation can serve the preferred locale.
3. Keep the existing `SpeechTranscriberEngine` API for `RecordingSessionManager`; Parakeet post-recording ASR unchanged.

Happy to contribute a PR — local Debug build with this approach produces correct Russian live captions on macOS 26.

### Alternatives considered

_No response_

### Privacy check

SubMax:
## Summary

Fixes #34

Live transcription used only Apple `SpeechTranscriber` and resolved Auto via `Locale.current`. On an English-only app bundle that stays `en` even when macOS Preferred Languages are `ru-RU`, and `SpeechTranscriber.supportedLocales` often omits Russian while `DictationTranscriber` supports it — so live captions fell back to English.

This PR:
- Resolves Auto / default locale from `Locale.preferredLanguages` (mapped through `TranscriptionLanguage`)
- Prefers `SpeechTranscriber` when the locale is supported; otherwise uses `DictationTranscriber` with long-form-friendly options
- Avoids silently substituting English Speech when Dictation can serve the preferred locale
- Leaves Parakeet post-recording ASR and `RecordingSessionManager` call sites unchanged

## How was this verified?

- [ ] Ran `make setup` at least once, so the pre-commit hooks are installed
- [x] Builds cleanly (`xcodebuild build -project Logue.xcodeproj -scheme Logue -destination 'platform=macOS'`)
- [ ] Lints cleanly (`make lint`) — runs SwiftFormat **and** SwiftLint in strict mode, the same two checks as CI. Use `make format` to auto-fix formatting.
- [x] No new `swiftlint:disable` without justification
- [ ] Added/updated tests where it makes sense
- [x] Ran `xcodegen generate` if I added/removed `.swift` files or changed `project.yml` *(N/A — no new files / no `project.yml` changes)*

Manual:
- [x] macOS 26, system Preferred Languages `ru-RU`: Console shows DictationTranscriber selection; live captions in Russian during recording
- [ ] Spot-check English still uses `SpeechTranscriber`
- [x] Post-recording Parakeet still produces a usable Russian transcript

## Checklist

- [x] My change keeps all AI inference and data processing **on-device** (no new network calls to third-party services without discussion). *(Uses Apple on-device Speech / Dictation assets only; may download system speech language packs via `AssetInventory`.)*
- [x] I did not commit secrets, API keys, or personal data.
- [x] I read [CONTRIBUTING.md](../CONTRIBUTING.md) and followed the security/concurrency standards.

## Notes for reviewers

- Key files: `SpeechTranscriberEngine.swift`, `MeetingNote.swift` (`TranscriptionLanguage`), small wiring in `RecordingSessionManager.swift`.
- Do **not** trust `SpeechTranscriber.supportedLocale(equivalentTo:)` alone for Russian — it can return `ru_RU` even when that locale is not in `supportedLocales`; selection uses membership in `supportedLocales`.
- Dictation path uses explicit options (`volatileResults`, `frequentFinalization`, `farField`) rather than a short-phrase preset.
- First recording for a Dictation-only locale may trigger an on-device model download.


SubMax:
Closing this in favour of a cleaner PR.

Reviewing my own diff against your notes in #34, I found problems that are easier to fix by
starting over than by patching this branch:

- **Unrelated churn.** The Dictation fallback needed roughly 120 lines, but I rewrote `setup()`
  wholesale and tidied things that had nothing to do with the fix — dropped the `Bug-6` and `C-N5`
  comments, removed the unused `failedToSetup` error case, trimmed `en-CA`/`en-AU` out of the
  English ladder, inlined a local. That noise is on me and it makes the real change harder to see.
- **A regression I introduced.** I added an `AssetInventory.release(reservedLocale:)` call in
  `finish()`. In in-person mode `RecordingSessionManager` runs a second `SpeechTranscriberEngine`
  for the mic on the same locale, so one engine finishing releases a reservation the other still
  needs. The original code never released, and releasing is out of scope here.
- **26 commits for a 3-file change.** The branch was cut from a local `main` carrying the
  markdown-import work that landed separately in #26, so the Commits tab is unreadable even though
  the content diff is only 3 files.
- **No tests.** The locale resolution is pure logic and should be covered.

The replacement PR will keep exactly what you asked for in #34 — stable `SpeechTranscriberEngine`
API, English ladder as a genuine last resort, explicit logging of which engine was selected — plus
unit tests for `TranscriptionLanguage` locale resolution and the BCP-47 matching (the `ru_RU` vs
`ru-RU` trap you flagged), and a single commit rebased on current `main`.

I'll also correct the description: as you pointed out, the Auto bug isn't `Locale.current` — it's
`TranscriptionLanguage.auto.locale` returning `nil` and `setup(locale:)` hardcoding `en-US`, which
makes Auto English for every user, not just English-bundle ones.

Thanks for the detailed review. New PR shortly, still against #34.

Issue #37, opened 2026-07-29
Connect the Chrome extension to Logue, then let Logue work in the browser


shanforge:
## Summary

Logue's Chrome extension is built and ready ([`bitwize-ai/logue-chrome-extension`](https://github.com/bitwize-ai/logue-chrome-extension)) but has nothing to talk to. It expects a local HTTP server inside the macOS app on `127.0.0.1:52452`, and that server does not exist yet — so every panel in the extension currently reports "Logue is not running".

This issue covers connecting the two, and then the larger capability that connection makes possible: letting Logue see and act in the browser the user is already signed into.

Everything stays on the machine. The extension talks to `localhost`, the local MLX model answers, and the answer comes back down the same loopback connection. Nothing here introduces a network call.

---

## Phase 1 — Make the extension work

The bridge itself: discovery, a session handshake, model listing, and chat with streaming. Enough that the extension connects and its chat panel works end to end.

### Logue (macOS app)

- [ ] Loopback HTTP server (`NWListener` bound to `127.0.0.1`, never `0.0.0.0`)
- [ ] Minimal HTTP/1.1 request parsing and response writing, with keep-alive
- [ ] Origin check — only `chrome-extension://` callers, which the browser sets and a page cannot forge
- [ ] `GET /v1/logue/status` — app fingerprint, whether a model is loaded, active model, version
- [ ] `POST /v1/logue/handshake` — session token the extension stores
- [ ] `GET /v1/models` — the model list, marking the active one
- [ ] `POST /v1/logue/chat` — non-streaming answer
- [ ] `POST /v1/chat/completions` — streaming answer as Server-Sent Events
- [ ] Page content passed as `context` wrapped in delimiters before it reaches the model — it is someone else's web page, so it is data, not instructions
- [ ] Every endpoint the extension calls but this phase does not serve answers `501` with a readable message, not a bare failure
- [ ] **Opt-in toggle in Settings → Privacy, off by default.** A listening socket is not something to open on every user's behalf
- [ ] Server starts and stops with the toggle, and stops when the app quits
- [ ] Tests for the parsing and routing rules (pure, no socket needed)

### Chrome extension

- [ ] `host_permissions` covers every port discovery scans — currently only `52452` is permitted while discovery tries three
- [ ] Handle `501` as "this tool isn't available yet" rather than a connection error, so a partial server doesn't read as a broken app
- [ ] Connection status distinguishes *Logue not running* from *running but no model loaded* — different problems, different fixes
[...]

PR #38, opened 2026-07-29
feat: a loopback bridge so the browser extension can use this Mac's models

shanforge:
Phase 1 of #37 — the macOS side.

The Chrome extension has been complete for a while with nothing to talk to: it expects a local HTTP server on `127.0.0.1`, so every panel reported "Logue is not running". This is that server.

## What it serves

| Endpoint | |
|---|---|
| `GET /v1/logue/status` | fingerprint, whether a model is loaded, active model, version |
| `POST /v1/logue/handshake` | session token the extension stores |
| `GET /v1/models` | model list, marking the active one |
| `POST /v1/logue/chat` | answer, streaming or not |
| `POST /v1/chat/completions` | streaming answer as Server-Sent Events |

Everything else the extension might call answers `501` with a message naming the tool. "That button does nothing yet" and "you are talking to the wrong app" are different things, and a `404` says the wrong one.

## The privacy posture is the design

- **Off until the user turns it on**, under Settings → Privacy. A listening socket is not something to open on everyone's behalf.
- **Loopback only.** The listener restricts itself to the loopback interface *and* checks the peer address on every connection. The socket binds to the wildcard address, so "only this Mac can reach it" rests entirely on that filter — more weight than one parameter should carry alone. Verified: the LAN address is refused while `127.0.0.1` works.
- **No authentication, deliberately.** On loopback the meaningful boundary is the browser's own — it sets `Origin` itself and a page cannot forge it, so an ordinary website cannot quietly use the machine's model. What that does *not* stop is another program running as the same user. That is exactly why this is opt-in, and why the Settings copy names the address plainly instead of describing it as a convenience.
- **Page content is wrapped in delimiters** before it reaches the model, the same rule every other prompt in the app follows. It is somebody else's web page, so it is data and not instructions.
- Generation **stops when the client disconnects**. Writing into a closed connection fails quietly, so nothing used to notice — a user pressing stop left the model generating for nobody and holding the inference gate.

## Structure

`HTTPMessage` (parsing and responses) and `BrowserBridgeRoute` (which paths exist, which origins are allowed) are pure, so the rules are testable without opening a socket. `BrowserBridgeServer` does sockets; `+Handlers` is the API surface.

## Verification

**Run against the real app, with a model loaded:**

- Status, handshake, models, CORS preflight, `403` for a non-extension origin, `404`, `405`, `501` naming the tool, `503` when no model is loaded
- The LAN address refused while loopback works
- **Chat streaming end to end** — 47 KB of tokens in correctly-framed SSE, terminated properly
- **Disconnect handling** — cut a client mid-answer and the server logged the stop rather than running to completion

**Unit tests:** 26, covering partial reads, pipelined requests on a keep-alive connection, absurd `Content-Length`, origin rules, method rules, response framing and prompt building.

Build clean, SwiftFormat and SwiftLint clean at the CI-pinned versions.

## Three bugs only running it would have found

- `NWListener` reports a port collision **asynchronously**, so the original code claimed success before the bind failed and reported a port it had not taken.
- `/v1/models` returned `405` — `GET` was allowed only on status, and the extension uses `GET` for both.
- A deliberate `403` arrived as an **empty response**, because closing the connection raced the write. Connections now close only after their bytes are on the wire.

## Not in this phase

The other 16 tool endpoints. The extension no longer exposes them at all (see the companion PR), so nothing routes to a `501` in normal use — the handling stays as a safety net for older builds.

Paired with `bitwize-ai/logue-chrome-extension#…` — that branch is the browser half and the two need each other to be useful.


shanforge:
image

Logue/Services/BrowserBridge/BrowserBridgeRoute.swift
```diff
@@ -0,0 +1,99 @@
...
+    ///
+    /// Not authentication — the user was clear it should just work, and a browser sets this
+    /// header itself so a page cannot forge it. What it does buy is that an ordinary website
+    /// cannot quietly use the machine's model: the browser refuses to hand it the response.
+    /// A request with no `Origin` at all is allowed, because that is `curl` on the user's own
+    /// machine, which they are entitled to do.
+    private static let allowedOriginScheme = "chrome-extension://"
+
+    static func isAllowed(origin: String?) -> Bool {
+        guard let origin, !origin.isEmpty, origin != "null" else { return true }
```
westerosweb: **Blocker — this allows any website on the internet to use the bridge.**

`origin != "null"` *fails* the guard when the origin is the string `"null"`, so the `else` branch runs and returns `true` — allowed. Then `BrowserBridgeServer.swift:340` passes it through as `allowedOrigin` and `HTTPMessage.serialise` emits `Access-Control-Allow-Origin: null`.

Per the Fetch spec an opaque origin serializes to `null` and matches an ACAO of `null`, so the browser hands the response back to the caller. Any page on the internet can do this from a sandboxed iframe:

```html
<iframe sandbox="allow-scripts" srcdoc="<script>
  fetch('http://127.0.0.1:52452/v1/logue/chat',
        {method:'POST', headers:{'Content-Type':'text/plain'},
         body:'{\"message\":\"...\"}'})
    .then(r=>r.text()).then(t=>parent.postMessage(t,'*'))
</script>"></iframe>
```

A `data:` document or a cross-origin redirect chain gives the same `Origin: null`. With `Content-Type: text/plain` it is a *simple* request so there is no preflight to fail — and even with `application/json` the OPTIONS preflight is answered `204` with `ACAO: null` and `Allow-Methods: GET, POST, OPTIONS`, so that path works too. Chromium's Private Network Access preflight blocks some of this in some versions; it is not implemented in Safari or Firefox and its enforcement has repeatedly slipped, so it is not the defense.

The reasoning in the doc comment above is right — a page genuinely cannot forge `Origin`. This one clause hands the boundary away. The `nil` case is fine, because no `Origin` means no ACAO is emitted and the response is unreadable cross-origin; `""` should be refused (it currently emits a malformed empty ACAO).

```suggestion
        guard let origin else { return true }        // absent: curl on the user's own machine
```

...with the `hasPrefix` line below then handling `""`, `"null"`, and everything else. Please add tests for both `"null"` and `""` — `missingOriginAllowed` sits right next to this and asserts the other half of it.

Logue/Services/BrowserBridge/BrowserBridgeRoute.swift
```diff
@@ -0,0 +1,99 @@
...
+        return origin.hasPrefix(allowedOriginScheme)
+    }
+
+    /// Decides how to answer, without doing any of the work.
+    static func decide(method: String, path: String, origin: String?) -> Decision {
+        guard isAllowed(origin: origin) else { return .forbiddenOrigin }
```
westerosweb, line 102: **Blocker — DNS rebinding reads `status` and `models` with no `Origin` involved at all.**

A missing `Origin` is allowed (deliberately, for `curl`), and nothing anywhere validates the HTTP `Host` header — I grepped the diff, the only host logic is the peer-endpoint filter in `BrowserBridgeServer`.

So: attacker serves `evil.com` with a 1s TTL and rebinds it to `127.0.0.1`. The page then fetches `http://evil.com:52452/v1/logue/status`. That is **same-origin**, so no `Origin` header is sent and CORS is not involved — the response is fully readable. The connection arrives on loopback, so the peer filter passes. `GET` is permitted on exactly the two routes this reaches.

Yield: app version, active model display name, and the user's full model inventory including any custom models they added. POST is not exposed this way because browsers send `Origin` on all POSTs even same-origin, so the blast radius is read-only — but it is still an unauthenticated cross-origin read, and it survives the `null`-origin fix above.

Standard defense, about four lines: validate `Host` against `127.0.0.1:<activePort>` / `localhost:<activePort>` and reject anything else with 403 before routing.

Logue/Services/BrowserBridge/BrowserBridgeServer+Tools.swift
```diff
@@ -0,0 +1,164 @@
...
+    /// chat template, and passing through whatever a caller sent is how a template ends up
+    /// rendering something nobody intended.
+    nonisolated static func chatRequest(from body: [String: Any]) -> ChatRequest? {
+        guard let rawMessages = body["messages"] as? [[String: Any]] else { return nil }
+
+        let messages = rawMessages.compactMap(sanitisedMessage)
+        guard !messages.isEmpty else { return nil }
+
+        let tools = (body["tools"] as? [[String: Any]] ?? [])
+            .compactMap { $0 as? ToolSpec }
```
westerosweb: **Blocker — this `compactMap` filters nothing, and the cast emits a compiler warning.**

`ToolSpec` is `public typealias ToolSpec = [String: any Sendable]` (`MLXLMCommon/Tool/Tool.swift:5`). `Sendable` is a **marker protocol**, so casts to it are erased at runtime — `[String: Any] as? ToolSpec` always succeeds. It compiles to `warning: conditional cast from '[String : Any]' to 'ToolSpec' (aka 'Dictionary<String, any Sendable>') always succeeds`, which is a warning shipping into a repo with pre-commit lint hooks.

So whatever arbitrary JSON the caller put in `tools` — nested values and all — goes straight to `tokenizer.applyChatTemplate(messages:tools:)`, i.e. into a Jinja renderer. That is precisely what the doc comment three lines above says does not happen:

> Unknown keys are dropped rather than forwarded: everything here is about to be handed to a chat template, and passing through whatever a caller sent is how a template ends up rendering something nobody intended.

True of `messages` via `sanitisedMessage`, not true of `tools`.

Second-order bug on the same line: `body["tools"] as? [[String: Any]]` fails *wholesale* if the array contains any non-object element, so a single junk entry collapses the cast to `nil` and silently degrades to the no-tools path instead of erroring.

Fix: a real `sanitisedToolSpec(_:) -> ToolSpec?` mirroring `sanitisedMessage` — require `type == "function"` and `function.name` as `String`, keep only `description` (String) and `parameters` (recursively whitelisted JSON object), cap the tool count and the name/description lengths, drop the rest. This function has zero test coverage today, which is how the warning got through.

Logue/Services/BrowserBridge/BrowserBridgeServer.swift
```diff
@@ -0,0 +1,395 @@
...
+    /// The waiting is the point. `NWListener` reports a port collision *asynchronously*, through
+    /// its state handler — the initialiser and `start()` both succeed regardless. Treating those
+    /// as success meant the server reported itself listening on a port it had failed to take, and
+    /// the failure arrived moments later and quietly stopped it again.
+    private func startFirstFreePort() async {
+        for port in AppConstants.BrowserBridge.candidatePorts {
+            do {
+                try await listen(on: port)
+                activePort = port
+                lastError = nil
```
westerosweb, line 85: **Toggling the bridge off during startup can leave a live listening socket.**

There is no `Task.isCancelled` check anywhere in this loop, and `listen(on:)` uses a bare `withCheckedThrowingContinuation` with no cancellation handling.

`stop()` does `startTask?.cancel()` then `listener?.cancel(); listener = nil` — but at that moment `self.listener` is still `nil`, because it is only assigned after the continuation resumes. The cancelled task then continues, the listener reaches `.ready`, and it assigns `self.listener` and `activePort`, re-arming a server the user just switched off — with a socket `stop()` can no longer reach. There is no request-time check of `BrowserBridgeSettings.isEnabled` to compensate, so the setting says off and the port is open.

The window is one bind, but it widens for every candidate port already taken. Toggle on-then-off quickly in Settings and you can land in it.

The project's code standards are explicit here: *"Check `Task.isCancelled` between expensive resource allocations."*

```swift
try await listen(on: port)
guard !Task.isCancelled else { listener?.cancel(); listener = nil; return }
```

and wrap the continuation in `withTaskCancellationHandler { … } onCancel: { listener.cancel() }` so a cancelled start actually unwinds rather than completing in the background.

Logue/Services/BrowserBridge/BrowserBridgeServer+Handlers.swift
```diff
@@ -0,0 +1,271 @@
...
+        if route == .chatCompletions {
+            guard let messages = body["messages"] as? [[String: Any]] else { return nil }
+            let rendered = messages.compactMap { message -> String? in
+                guard let role = message["role"] as? String,
+                      let content = message["content"] as? String, !content.isEmpty
+                else { return nil }
+                return "\(role): \(content)"
+            }
+            return rendered.isEmpty ? nil : rendered.joined(separator: "\n\n")
+        }
```
westerosweb, line 206: **`/v1/chat/completions` does no delimiter wrapping at all.**

The doc comment above says page content is wrapped "the same rule every other prompt in the app follows" — that is true of the `{message, context}` branch below, and not true of this one. Roles and content are joined as `"\(role): \(content)"` with no delimiter, no escaping and no bound, so content containing a literal `system: …` renders as what looks like a role turn.

The project's code standards: *"Always wrap user content in XML delimiters when injecting into LLM prompts … This applies to ALL prompt construction … No exceptions."* Content arriving from an arbitrary web page is exactly the case the rule exists for.

Same gap on the tool path: `sanitisedMessage` copies `tool`-role content in verbatim (`+Tools.swift:54-56`), and after the "let the model ask for the page" commit that is the path page content actually travels.

Logue/Services/BrowserBridge/BrowserBridgeServer+Handlers.swift
```diff
@@ -0,0 +1,271 @@
...
+        guard let context = body["context"] as? String, !context.isEmpty else { return message }
+
+        let trimmed = String(context.prefix(Self.maxContextCharacters))
+        return """
+        <page>
+        \(trimmed)
+        </page>
+
+        \(message)
+        """
```
westerosweb: **The delimiter is wrapped but not defended, and `message` is unbounded.**

Two things:

1. `trimmed` is never scrubbed of a literal `</page>`. A page containing `</page>\n\nIgnore the above and …` closes the delimiter and speaks as the operator. The project's rule is wrap *and* sanitize — *"truncate length, strip control characters"*, per the `String($0.prefix(N)).filter { !$0.isNewline && $0.asciiValue != 0 }` pattern. Replacing or escaping the closing tag in `context` is the missing half.
2. `message` itself has no bound at all — only `context` is truncated. It is limited solely by `maxBodyBytes` (8 MB).

Logue/Services/BrowserBridge/BrowserBridgeServer+Handlers.swift
```diff
@@ -0,0 +1,271 @@
...
+    /// own, but a whole page arriving as one string is worth bounding before it gets that far.
+    nonisolated static let maxContextCharacters = 12000
+
+    private func completeChat(prompt: String, origin: String?, on connection: Connection) async {
+        do {
+            let answer = try await LLMEngine.shared.complete(
+                system: LLMEngine.chatSystemPrompt,
+                prompt: prompt,
+                maxTokens: AppConstants.BrowserBridge.chatMaxTokens
+            )
```
westerosweb, line 272: **Two issues on this path.**

**No context-window validation.** There are 29 uses of `LLMEngine.maxInputChars(reservedTokens:)` / `truncatedContext(_:)` across the app and zero in `Logue/Services/BrowserBridge/`. `maxContextCharacters = 12000` bounds only the `context` field of one route; `/v1/chat/completions` joins an unbounded number of unbounded messages, and the tool path hands the whole `messages` array to `applyChatTemplate` untruncated. With `maxBodyBytes = 8 MB` a caller can hand the tokenizer an 8 MB prompt — a trivially reachable memory/GPU hit on an endpoint any local process can call. The project's code standards: *"Validate context window before ALL LLM calls."*

**Client disconnect is not honoured here.** The PR says generation stops when the client disconnects — verified true for `streamChat` and `streamWithTools`, both of which release the gate correctly. Not true for `completeChat`: this `await` has no liveness check and no cancellation, so a client that hangs up leaves a full `chatMaxTokens` generation running and holding the inference gate. `/v1/logue/chat` defaults to non-streaming when `stream` is absent, so this is a normal path rather than a corner.

`complete()` already supports cancellation via `withTaskCancellationHandler`, so cancelling an outer task propagates — having `Connection.close()` fire a stored `onDisconnect` would be enough.

Logue/Services/BrowserBridge/BrowserBridgeServer+Tools.swift
```diff
@@ -0,0 +1,164 @@
...
+        let wantsStream = (body["stream"] as? Bool) ?? true
+
+        return ChatRequest(messages: messages, tools: tools, wantsStream: wantsStream)
+    }
+
+    /// One message, reduced to the fields a chat template understands.
+    nonisolated private static func sanitisedMessage(_ raw: [String: Any]) -> [String: any Sendable]? {
+        guard let role = raw["role"] as? String,
+              ["system", "user", "assistant", "tool"].contains(role)
```
westerosweb: **Callers can supply their own `system` turn.**

`"system"` is in the role whitelist, and `completeWithTools` never applies `LLMEngine.chatSystemPrompt`. On the tool path that becomes a genuine system turn in the chat template; on the `prompt(from:)` path it becomes literal `system: …` text. Either way the caller, not the app, defines the model's instructions — and combined with the `null`-origin bug, that caller can be an arbitrary website.

Suggest dropping `"system"` from the accepted roles and having the bridge supply the system turn itself.

Logue/Services/BrowserBridge/HTTPMessage.swift
```diff
@@ -0,0 +1,191 @@
...
+        var headers: [String: String] = [:]
+        for line in lines where !line.isEmpty {
+            guard let colon = line.firstIndex(of: ":") else { throw ParseError.malformed }
+            let name = line[line.startIndex ..< colon].trimmingCharacters(in: .whitespaces).lowercased()
+            let value = line[line.index(after: colon)...].trimmingCharacters(in: .whitespaces)
+            guard !name.isEmpty else { throw ParseError.malformed }
+            headers[name] = value
+        }
+
+        let declaredLength = headers["content-length"].flatMap(Int.init) ?? 0
```
westerosweb: **`Transfer-Encoding` is ignored and an unparseable `Content-Length` silently becomes 0.**

Three variants, all giving the same primitive on a connection that deliberately supports pipelining:

1. `Transfer-Encoding: chunked` is never checked. A chunked body is treated as length 0, the chunk data stays in the buffer, and `drainBuffer` parses it as the *next* request.
2. A present-but-unparseable value (`abc`, or a number overflowing `Int`) falls back to `0` via `?? 0` rather than `malformed`.
3. Duplicate `Content-Length` headers last-wins at line 78 instead of being rejected.

Also at 73-79: obs-fold continuation lines beginning with SP/HTAB are parsed as independent headers, so one containing a colon is smuggled in as its own header.

With one trusted loopback client this is not an internet-facing smuggling bug — but the code is built for pipelining, and the `null`-origin issue makes "one trusted client" untrue today. Suggest: reject any request carrying `Transfer-Encoding` with 501, distinguish "absent" (length 0) from "present but unparseable" (`malformed`), reject duplicate `Content-Length`, and reject lines starting with whitespace.

Logue/Services/BrowserBridge/BrowserBridgeServer.swift
```diff
@@ -0,0 +1,395 @@
...
+                method: request.method, path: request.path, origin: origin
+            )
+            let allowedOrigin = BrowserBridgeRoute.isAllowed(origin: origin) ? origin : nil
+
+            Task { [weak self] in
+                guard let self else { return }
+                await BrowserBridgeServer.shared.handle(
+                    decision: decision, request: request, origin: allowedOrigin, on: self
+                )
+            }
```
westerosweb: **Pipelined requests are dispatched concurrently, so responses can come back out of order.**

`drainBuffer` loops parsing requests and calls `dispatch` for each; `dispatch` spawns this unstructured `Task` and returns immediately. Two pipelined requests therefore run concurrently and their responses are written in completion order rather than request order. HTTP/1.1 requires strict ordering on a persistent connection — a client would attribute response #2 to request #1.

`fetch` does not pipeline, so this is latent for the intended client. But `readsPipelinedRequests` tests the *parser's* ability to handle pipelining, which advertises support the server does not actually provide. It also interacts badly with SSE: a second pipelined request's response would be written into the middle of an in-flight event stream.

Fix: serialise per connection — keep a `Task` chain on `Connection` and await the previous one, or stop draining after one dispatch and resume once its response is on the wire.

Logue/Services/BrowserBridge/BrowserBridgeServer+Handlers.swift
```diff
@@ -0,0 +1,271 @@
...
+    /// there is no authentication here by design. It is worth being blunt about that rather than
+    /// letting the ceremony imply a guarantee it does not provide.
+    private func handshakePayload() -> [String: Any] {
+        [
+            "app": Self.appIdentifier,
+            "session": UUID().uuidString,
+            "port": Int(activePort ?? 0),
+            "version": Self.appVersion,
+        ]
+    }
```
westerosweb, line 124: The doc comment is admirably blunt that this secures nothing, and I'd rather have that than ceremony pretending otherwise. But the *wire shape* still teaches the extension author that auth exists: `X-Logue-Session` appears only in `Access-Control-Allow-Headers` and one test — nothing reads it, stores it, or compares it.

Generation is fine (`UUID` is CSPRNG-backed on Darwin). Suggest either wiring it up properly (issue it, require it on later requests, compare in constant time — which would also fix the "any installed extension is trusted" problem) or renaming the field so nobody builds on it.

Logue/Views/Settings/BrowserExtensionSection.swift
```diff
@@ -0,0 +1,65 @@
...
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 10) {
+            Text("Browser Extension")
+                .font(.headline)
+
+            Text("Let the Logue browser extension use this Mac's models — chat and writing tools "
+                + "in your browser, answered by the model already running here. "
+                + "Nothing is sent anywhere: the extension talks to Logue over your Mac's own "
+                + "loopback address, and the reply comes back the same way.")
```
westerosweb: **"Nothing is sent anywhere" is false when an external provider is configured.**

`LLMEngine.isReady` is `modelContainer != nil || apiClient != nil`, and both `complete()` and `completeStream()` prefer `apiClient` when it is set. So on the `/v1/logue/chat` and `/v1/chat/completions` paths, a browser page's content is POSTed to whatever external endpoint the user configured.

The second block at line 37 is about *inbound* reachability and is accurate. This one is an outbound claim, it is in the Privacy tab, and it is the product's central promise — so it should be exactly right.

Either refuse the bridge when the active model is an API provider (503, same as no model loaded), or reword to something like "answered by whichever model Logue is set to use — if you have configured an external provider, the text goes there."

Minor, same file: `@State private var isEnabled = BrowserBridgeSettings.isEnabled` at line 11 snapshots UserDefaults at view init and never re-reads. Fine today since only this view writes it, but it will go stale if anything else ever toggles the setting.

westerosweb (commented):
Review of the bridge. Most of this is careful work — the purity split between `HTTPMessage`/`BrowserBridgeRoute` and the socket layer is what makes the rules testable at all, the async port-collision handling is correct and genuinely non-obvious (waiting for a *settled* state, with a single-resume guard), the write/close fix is the right fix, `isLoopback` handles IPv4-mapped IPv6, and I traced the stream cancellation through `onTermination` in `buildMLXStream`/`completeWithTools` — it really does release the inference gate.

The most important design decision is that the tool loop runs in the extension rather than in the app. Nothing reachable through this socket touches `MeetingStore`, the filesystem, or the Accessibility API. In an unsandboxed process that is what keeps prompt injection from mattering, and it is worth writing down somewhere as a boundary the extension side must not erode.

Three blockers, inline. The pattern behind them is worth naming: **each is a place where a comment asserts a protection the code does not implement.** The origin comment says a page cannot forge `Origin` (true) but the check lets `null` through; the tool comment says unknown keys are dropped (they are not); the settings copy says nothing is sent anywhere (not true with an external provider configured).

Not anchored inline:

- **No idle timeout or head-size cap.** `maxConcurrentConnections = 16`, and connection entries are only removed on peer close, read error, or `stop()`. Sixteen sockets that open and send nothing wedge the bridge indefinitely; they also accumulate naturally from keep-alive sockets left by a crashed client. Separately the only bound on the request *head* is `maxBodyBytes`, so headers can reach 8 MB each. Suggest a per-connection idle timer (interval in `AppConstants.Delays`, per the constants rule) and a separate ~32 KB head cap.
- **Port squatting.** The three candidate ports are scanned in order by both sides and the only identity signal is the constant `"logue-mlx"`. A local process that binds 52452 at login receives every page the user runs through the extension while Logue falls back to 52453 and looks healthy. That is materially worse than the acknowledged "another local process can use the model" — it is inbound capture of browsing content rather than outbound use of compute.
- **Test gaps sit exactly on the risk.** `isLoopback` carries the whole privacy claim and has zero tests, despite being a pure static function over `NWEndpoint` — `.ipv4("127.0.0.1")`, `.ipv6("::1")`, `.ipv6("::ffff:127.0.0.1")`, `.ipv4("192.168.1.5")` and `.hostPort(host: .name("localhost"))` are all trivially constructible. `chatRequest`/`sanitisedMessage`/tool-spec parsing have zero tests — one assertion would have caught the tools issue. `missingOriginAllowed` covers `nil` and `""` and stops one case short of the bug. `pageContextIsBounded` asserts only `< huge.count`, so it passes if truncation drops a single character; assert against `maxContextCharacters`. A small socket-level suite (bind an ephemeral port, drive it with `URLSession`) would cover the connection lifecycle, disconnect, and SSE framing in ~100 lines.
- **Worth revisiting with the user:** even with the `null` bug fixed, `hasPrefix("chrome-extension://")` admits *every* installed extension, including one that later ships a malicious update and only needs to add `http://127.0.0.1/*` to its host permissions. The "it should just work" rationale is worth re-testing now the surface is concrete — a pairing code shown in Settings turns "any extension" into "this extension", and the unused `X-Logue-Session` is already the natural place to bind it.
- **Merge conflict** is `project.pbxproj`: both branches regenerated it with fresh XcodeGen UUIDs. Take main's version wholesale and re-run `xcodegen generate` — a manual merge of that file is not meaningful.

Logue/Services/BrowserBridge/BrowserBridgeRoute.swift
shanforge: Confirmed and fixed in 4a167d7, taking your suggestion exactly:

```swift
guard let origin else { return true }        // absent: curl on the user's own machine
return origin.hasPrefix(allowedOriginScheme) // "", "null" and everything else
```

You are right about the mechanism and right that Private Network Access is not the defence — it is unimplemented in Safari and Firefox and has slipped repeatedly in Chromium, so nothing here should have been resting on it.

Tests for both halves, next to `missingOriginAllowed` as you asked: `opaqueOriginRefused` covers `"null"` at both `isAllowed` and `decide` (so the `.forbiddenOrigin` path is pinned, not just the predicate), and `emptyOriginRefused` covers `""`.

Logue/Services/BrowserBridge/BrowserBridgeRoute.swift
shanforge: Agreed, and this was the one I was most glad you separated out — it survives the `null`-origin fix precisely because no `Origin` is involved, which makes it easy to think the first fix covered it.

`Host` is now validated before anything else is considered (4a167d7). Exact match against `127.0.0.1:<port>`, `localhost:<port>` or `[::1]:<port>`, case-insensitive, port included — a rebound request reaches us on our port while naming its own host, so the port is the half that cannot be faked. Absent `Host` is refused too: HTTP/1.1 requires it, and "omit the header" is the obvious next thing to try against a name check.

One implementation note. The port comes from the listener that accepted the connection, not from `activePort`:

```swift
listener.newConnectionHandler = { connection in ... self?.accept(connection, boundPort: port) }
```

`activePort` is main-actor state read from a connection callback that is not, and it can legitimately differ from the port a given connection arrived on while the server is rebinding. Carrying it on the `Connection` keeps the check synchronous and unambiguous.

Tests: `onlyLoopbackHostAnswered`, `reboundHostRefused` (asserting `.forbiddenHost` on both GET routes this actually reached), `wrongPortRefused`, `missingHostRefused`.

Logue/Services/BrowserBridge/BrowserBridgeServer+Tools.swift
shanforge: Confirmed — the compiler says it out loud, which makes it worse rather than better:

```
BrowserBridgeServer+Tools.swift:40:30: warning: conditional cast from '[String : Any]' to
'ToolSpec' (aka 'Dictionary<String, any Sendable>') always succeeds
```

Fixed in 35d25b1 with the `sanitisedToolSpec` you describe: requires `type == "function"` and a `String` name, keeps only `description` and `parameters`, and rebuilds rather than passes through. The name is stripped to identifier characters so it cannot carry template syntax or newlines into the render. `parameters` is arbitrary nested JSON by definition, so it is whitelisted by *type* — object, array, string, number, bool — to a bounded depth, with caps on tool count, name and description length.

Your second-order bug is covered too: the array is read as `[Any]` and each element cast individually, so one junk entry drops itself instead of collapsing the cast to nil and silently taking the no-tools path. `junkToolDoesNotDropTheRest` pins it.

You were right that the missing coverage is how the warning got through — the function had none. It has seven cases now.

Checked against the shipped extension before changing the shape: `CHAT_TOOLS` in `src/sidepanel/tools/definitions.ts` is already exactly `{type: "function", function: {name, description, parameters}}`, and the names (`read_current_page`, `list_page_elements`, …) survive the identifier filter unchanged.

Logue/Services/BrowserBridge/BrowserBridgeServer+Tools.swift
shanforge: Agreed, and dropped in 35d25b1 — `system` is no longer an accepted role, and the bridge supplies the turn itself:

```swift
let systemTurn: [String: any Sendable] = ["role": "system", "content": LLMEngine.chatSystemPrompt]
let messages = [systemTurn] + boundedToContextWindow(sanitised)
```

The `prompt(from:)` path drops it as well, so neither shape lets a caller define the model's instructions.

Verified against the shipped extension first: `system` appears in `WireMessage`'s type union in `src/shared/types.ts` but nothing anywhere constructs one — `toWireMessages` emits only `user` and `assistant`. So this costs the real client nothing.

`callerSystemTurnRefused` and `appSuppliesTheSystemTurn` cover both halves.

Logue/Services/BrowserBridge/BrowserBridgeServer+Handlers.swift
shanforge: Correct — the doc comment described the `{message, context}` branch and quietly spoke for both. Fixed in 35d25b1.

Turns are delimited and labelled by attribute now rather than concatenated:

```
<turn role="user">
…
</turn>
```

so content containing a literal `system:` is text inside a turn rather than something that looks like a role boundary. `system` is not an accepted role on this path either.

The tool path you point at is the important half, since after "let the model ask for the page" that is where page text actually travels — `tool`-role content goes through the same escaping now, and the whole conversation is bounded to the context window.

Logue/Services/BrowserBridge/BrowserBridgeServer+Handlers.swift
shanforge: Both taken (35d25b1).

Wrap *and* sanitize: `escapingDelimiters` truncates, neutralises any closing tag the text could use to escape the block it is in, and strips control characters — keeping newlines and tabs, without which page text is unreadable. The closing bracket is replaced rather than the tag deleted, so the text still reads as what the page said instead of silently losing a fragment.

`pageContentCannotEscapeItsDelimiter` asserts exactly one `</page>` survives in the rendered prompt: the one this file wrote. `callerCannotForgeATurnBoundary` does the same for `</turn>`.

And `message` is bounded now — you are right that it had nothing but the 8 MB body cap. `maxMessageCharacters` per message, `maxPromptCharacters` across the whole rendered prompt so many individually-legal messages cannot add up to the same problem.

Logue/Services/BrowserBridge/BrowserBridgeServer+Handlers.swift
shanforge: Both real, both fixed (35d25b1 and b57774a).

**Context window.** You are right that there were zero uses in this directory. `boundedToContextWindow` now trims oldest-first to `LLMEngine.maxInputChars(reservedTokens:)` — the same accounting the other 29 call sites use — reserving room for the output cap and the system turn. The newest turn is truncated rather than dropped if it alone exceeds the budget, since dropping it answers a question nobody asked.

**Disconnect.** Fixed, though not with a liveness check — there is no loop here to check in. `Connection.close()` now cancels that connection's response chain, which propagates into `complete()`'s own `withTaskCancellationHandler`, so the gate is released rather than held for a full `chatMaxTokens` generation. Your suggestion of an `onDisconnect` hook amounts to the same thing; chaining got it for free because I needed the chain anyway for the pipelining fix on the other thread.

Worth stating plainly since you flagged it as a normal path rather than a corner: `/v1/logue/chat` is still non-streaming by default, and that path now unwinds.

Logue/Services/BrowserBridge/BrowserBridgeServer.swift
shanforge: Confirmed, including the detail that makes it stick — `stop()` has nothing to cancel at that moment because `self.listener` is only assigned after the continuation resumes. Fixed in b57774a with both halves you suggest:

```swift
try await listen(on: port)
guard !Task.isCancelled else { listener?.cancel(); listener = nil; return }
```

and the continuation wrapped in `withTaskCancellationHandler { … } onCancel: { listener.cancel() }`, so a cancelled start unwinds instead of completing in the background and arming a server the user just switched off.

Fair citation of the project's own standard; that rule exists for exactly this shape.

Logue/Services/BrowserBridge/HTTPMessage.swift
shanforge: All four taken (b57774a), and I agree with how you framed it: not internet-facing smuggling, but the code is built for pipelining and "one trusted client" was not true while the `null`-origin hole was open. Defence in depth is cheaper than the argument about whether it is reachable today.

- `Transfer-Encoding` present → `501` and the connection closes. This was the dangerous one: length 0 left the chunk data in the buffer to be parsed as the next request.
- `Content-Length` absent → 0; present but unparseable → `malformed`. No more `?? 0` resolving a framing disagreement in the sender's favour.
- Duplicate `Content-Length` → `malformed`.
- A header line starting with SP or HTAB → `malformed`, so an obs-fold continuation containing a colon cannot be smuggled in as its own header.

The header work moved into `parseHeaders` and `bodyLength` — with the extra guards, `parseRequest` went one over SwiftLint's cyclomatic cap, and splitting it reads better than suppressing.

Four tests, one per case.

Logue/Services/BrowserBridge/BrowserBridgeServer.swift
shanforge: Agreed, and fixed in b57774a by serialising per connection — the first of your two suggestions:

```swift
pending.withLock { chain in
    let previous = chain
    chain = Task { await previous?.value; … }
}
```

Your point about the test is the one that decided it. `readsPipelinedRequests` tests the parser, and leaving the dispatch concurrent meant the suite advertised support the server did not actually provide. Ordering is now a property of the connection rather than of how fast each handler happens to finish.

The SSE interaction you mention is the one that would have been ugliest to debug — a second response written into the middle of an in-flight event stream, on a channel where the client is parsing frames.

It also turned out to be the cleanest fix for the disconnect issue on the other thread: closing a connection cancels its chain, which unwinds a non-streaming generation that nobody is waiting for.

Logue/Services/BrowserBridge/BrowserBridgeServer+Handlers.swift
shanforge: You are right about the hazard, and it is further along than the comment suggests: the shipped extension already builds on it. `src/background/api-client.ts` reads `handshake.session`, persists it, sends `X-Logue-Session` on every request, and treats a `401` as "session expired, re-handshake". None of which the server does anything with.

That rules out the rename, which was my first move and which I backed out. `handshake.session` going missing would leave `cachedSession` undefined, and `ensureSession` re-handshakes whenever it is falsy — so every request would carry a handshake. A wire rename is a coordinated release with the extension, not a free tidy-up.

Wiring it up properly is the other option you offer, and I do not think it buys what it looks like it buys. On loopback any local process can call `/v1/logue/handshake` and be issued a token of its own, so unless the token is bound to something the client cannot choose, "authenticated" would mean "asked first" — which is a stronger-sounding claim than the current one and no truer. First-handshake-wins would also break the extension on every service-worker restart. That trade seems worth your call rather than mine.

What I did do (7c8c013) is make the field honest and useful. It is stable for the life of the process instead of a fresh UUID per call, so it answers the one question a client actually has — did Logue restart? — and the doc says in as many words that it is not a credential, that nothing compares it, and that `X-Logue-Session` is ignored.

I also corrected the extension's own comments in the sibling repo, since they claimed "a session nonce for subsequent authenticated requests" and a comment asserting auth exists is how the next person builds on one that does not.

Happy to do the full rename plus a matching extension release if you would rather have the clean shape than the compatible one.

Logue/Views/Settings/BrowserExtensionSection.swift
shanforge: You are right, and this was the one worth getting exactly right rather than approximately — it is the Privacy tab, and it is the product's central promise.

Reworded (7c8c013) to "answered by whichever model Logue is set to use", and when the active model *is* an external provider the section now says so directly, where the user is deciding:

> Your active model is an external provider, so text sent from the browser goes to it rather than staying on this Mac. Switch to an on-device model to keep it local.

I chose that over refusing the bridge with a 503. Someone who deliberately configured a provider and turns on a browser bridge is not obviously making a mistake, and silently disabling a feature teaches less than saying what will happen. If you would rather it refuse outright, that is a small change and I will make it.

The second block at line 37 stays as it was, since as you say it is about inbound reachability and is accurate.

Minor point taken too: `isEnabled` is `@AppStorage` now rather than a `@State` snapshot taken at view init, so it cannot go stale if anything else ever toggles the setting.

Issue #15, opened 2026-07-26
Let the code block language be chosen from the UI


westerosweb:
## Context

Code blocks store a language (`case codeBlock(id:language:code:)`) and `Logue/Engine/CodeSyntaxHighlighter.swift` can highlight it — but there's no way to *set* the language except by typing the fence in raw markdown. A block created from the slash menu gets `language: ""` and no highlighting.

## What to do

Add a language picker to the code block, so highlighting works for blocks created in the editor.

## Acceptance criteria

- [ ] A focused code block shows a language control (picker or searchable menu)
- [ ] Choosing a language updates the block and re-renders with highlighting
- [ ] The language persists into the markdown fence (` ```swift `) and survives a reload
- [ ] Blocks with no language still render as plain code, no crash and no "undefined" label
- [ ] Language choice is preserved when the block is duplicated
- [ ] The list covers at least the common languages the highlighter supports — read the highlighter, don't guess

## Pointers

- `Logue/Views/Editor/BlockEditor/Views/BlockRowView.swift` — `codeBlockView(id:language:code:)` and the existing `codeBlockHeader`
- `Logue/Engine/CodeSyntaxHighlighter.swift` — the supported languages are the source of truth for the list
- `Logue/Views/Editor/BlockEditor/Model/BlockSerializer.swift` — fence serialization already handles the language; probably no change needed

## Nice-to-have (skip if it grows the PR too much)

Remember the most recently chosen language and default new code blocks to it.

**Difficulty:** small-medium.

---

## Getting set up

1. `git clone --recurse-submodules https://github.com/bitwize-ai/Logue.git`
2. `xcodebuild -downloadComponent MetalToolchain` — one-time, required by MLX
3. `brew install xcodegen swiftformat swiftlint`
4. `make setup` — **please don't skip this.** It installs the git hooks that run SwiftFormat and SwiftLint before each commit. CI runs the same two checks, and a formatting miss fails the build.
5. `xcodegen generate`, then open `Logue.xcodeproj`

Run `xcodegen generate` again any time you add a new `.swift` file.

Requires macOS 26 (Tahoe) on Apple Silicon and Xcode 26+.

- Setup guide: [docs/dev-setup.md](docs/dev-setup.md)
- Coding standards (please read before your first PR): [CLAUDE.md](CLAUDE.md)
- Contributing guide: [CONTRIBUTING.md](CONTRIBUTING.md)

[...]

Issue #14, opened 2026-07-26
Support editable callout blocks (> [!NOTE], GitHub alerts)


westerosweb:
## Context

Callouts are common in notes and docs:

```md
> [!NOTE] Local-first
> This note stays readable outside the app.

> [!WARNING]
> Careful here.
```

Today these parse as a plain block quote, so the callout type is lost visually and the styling is flat.

## What to do

Add a callout block type that renders styled and stays editable, while round-tripping the markdown exactly.

There's a **worked example to copy**: the mermaid and math blocks were added end-to-end recently. Follow the same path.

## Acceptance criteria

- [ ] `> [!NOTE]`, `> [!TIP]`, `> [!IMPORTANT]`, `> [!WARNING]`, `> [!CAUTION]` render as distinct styled blocks (icon + colour)
- [ ] The body stays editable in the editor
- [ ] **Round-trip is exact**: `parse` → `serialize` returns the original markdown, including the optional title after the type
- [ ] An unrecognised type (`> [!BANANA]`) degrades to a normal block quote rather than being dropped or crashing
- [ ] A plain `> quote` with no `[!TYPE]` still parses as a block quote, unchanged
- [ ] Reachable from the `/` slash menu
- [ ] Tests cover parse, serialize, round-trip, and the unrecognised-type case

## Where the changes go

Adding a block type touches a consistent set of places:

1. `Logue/Views/Editor/BlockEditor/Model/Block.swift` — new `case`, plus `id`, `isEmpty`, `searchableTexts`, `Equatable`, a factory
2. `Logue/Views/Editor/BlockEditor/Model/BlockSerializer.swift` — parse and serialize
3. `Logue/Views/Editor/BlockEditor/Views/BlockRowView.swift` — the row view
4. `Logue/Views/Editor/RichTextEditorHelpers.swift` — `BlockType` case + `makeBlock`, which is what puts it in the slash menu
5. `Logue/Views/Editor/BlockEditor/Model/DocumentSearchState.swift` — so find-in-document searches the body

The compiler will point you at most of these: `Block` is matched exhaustively in a few places.

## Reference commit

Look at how `case mermaid` and `case math` were added — same shape, and `LogueTests/DiagramMathBlockTests.swift` is a good template for the tests.

**Difficulty:** medium. Very learnable, because there's a pattern to follow.

---

## Getting set up

1. `git clone --recurse-submodules https://github.com/bitwize-ai/Logue.git`
[...]

Issue #13, opened 2026-07-26
Add pane layout shortcuts (Cmd+1 / Cmd+2 / Cmd+3)


westerosweb:
## Context

The main window shows a sidebar, a content list, and a detail/inspector area. There's no quick way to collapse down to just the editor for focused work — you have to drag dividers.

`Logue/Views/Editor/EditorLayoutMode.swift` already defines the model, with 6 tests in `LogueTests/EditorLayoutModeTests.swift`:

```swift
enum EditorLayoutMode { case editorOnly, editorAndList, allPanels }
// init?(shortcutNumber:), showsList, showsInspector, label
```

Nothing consumes it yet.

## What to do

Wire `Cmd+1` / `Cmd+2` / `Cmd+3` to the three layouts, and make the main window honour them.

## Acceptance criteria

- [ ] `Cmd+1` shows the editor only; `Cmd+2` editor + list; `Cmd+3` all panes
- [ ] Menu items appear under View with the shortcuts and `EditorLayoutMode.label` text
- [ ] The chosen layout persists across relaunch
- [ ] Layout decisions read `showsList` / `showsInspector` rather than re-deriving them
- [ ] Focus Mode (which already hides chrome) is not broken by this
- [ ] Manually dragging a divider afterwards still works
- [ ] `EditorLayoutModeTests` still pass

## Pointers

- `Logue/Views/MainWindowView.swift` — the `NavigationSplitView` this has to drive
- `Logue/App/LogueApp.swift` — `CommandGroup(after: .toolbar)` already holds the zoom items; the layout items belong nearby
- `Logue/Views/Editor/BlockEditor/Focus/FocusModeState.swift` — existing precedent for hiding chrome

## Why this is `help wanted` rather than `good first issue`

Driving `NavigationSplitView` column visibility from external state is fiddlier than it looks, particularly making it agree with user-dragged dividers. Very doable, but expect to iterate.

**Difficulty:** medium.

---

## Getting set up

1. `git clone --recurse-submodules https://github.com/bitwize-ai/Logue.git`
2. `xcodebuild -downloadComponent MetalToolchain` — one-time, required by MLX
3. `brew install xcodegen swiftformat swiftlint`
4. `make setup` — **please don't skip this.** It installs the git hooks that run SwiftFormat and SwiftLint before each commit. CI runs the same two checks, and a formatting miss fails the build.
5. `xcodegen generate`, then open `Logue.xcodeproj`

[...]

Issue #12, opened 2026-07-26
Add a per-document icon picker


westerosweb:
## Context

`WritingDocument` has an `icon: String?` field and there's a validator, `DocumentIcon.sanitised(_:)`, with 7 tests in `LogueTests/DocumentIconTests.swift`. Nothing in the UI sets or displays it yet.

This issue is **UI only** — the model and validation are done.

## What to do

Let someone pick an emoji icon for a document, and show it wherever documents are listed.

## Acceptance criteria

- [ ] An icon picker is reachable from the document editor (the ⋯ menu or the title area)
- [ ] Picking an emoji sets `WritingDocument.icon` and persists it
- [ ] The icon renders in the document list rows and on document cards
- [ ] Documents with no icon look exactly as they do today — no layout shift, no placeholder box
- [ ] Clearing the icon works
- [ ] All input goes through `DocumentIcon.sanitised(_:)` — do not assign raw user input to `icon`
- [ ] `DocumentIconTests` still pass

## Pointers

- `Logue/Models/DocumentIcon.swift` — the validator
- `Logue/Views/Sidebar/FolderIconPicker.swift` — an existing icon picker; likely reusable as-is or with light generalisation
- `Logue/Views/Editor/DocumentWorkspaceView+Toolbar.swift` — where the menu item goes
- `Logue/Views/Content/DocumentListContentView.swift`, `Logue/Views/Home/DocumentCardView.swift` — display sites

## Why validation matters here

The icon is rendered in lists **and** interpolated into LLM prompts alongside the document title. `DocumentIcon.sanitised` therefore rejects anything that isn't a single grapheme without control characters. It deliberately **rejects** over-long input rather than truncating, because truncating could split a multi-scalar emoji (like 👩‍💻) into a broken fragment. Please keep that behaviour.

**Difficulty:** small-medium.

---

## Getting set up

1. `git clone --recurse-submodules https://github.com/bitwize-ai/Logue.git`
2. `xcodebuild -downloadComponent MetalToolchain` — one-time, required by MLX
3. `brew install xcodegen swiftformat swiftlint`
4. `make setup` — **please don't skip this.** It installs the git hooks that run SwiftFormat and SwiftLint before each commit. CI runs the same two checks, and a formatting miss fails the build.
5. `xcodegen generate`, then open `Logue.xcodeproj`

Run `xcodegen generate` again any time you add a new `.swift` file.

[...]

PR #39, opened 2026-07-29
Wire up the editor affordances whose models were already written (#9, #10, #11, #12, #13, #14, #15)

### LINKED ISSUE #9: Zoom In shortcut does not fire on keyboards where + requires Shift
## Context

The editor has zoom controls in the **View** menu, bound in `Logue/App/LogueApp.swift`:

```swift
Button("Zoom In") { applyZoom { $0.zoomIn() } }
    .keyboardShortcut("+", modifiers: .command)
```

On many non-US keyboard layouts, typing `+` requires Shift, so `Cmd`+`+` never matches this binding and Zoom In silently does nothing. Zoom Out (`-`) and Actual Size (`0`) are unaffected.

## What to do

Add `=` as an additional binding for Zoom In — this is the standard approach, and what browsers and most editors do, because `=` is the unshifted key in the same physical position.

You'll need a second `Button` with the `=` shortcut (SwiftUI does not accept two shortcuts on one button). Consider whether the duplicate should be hidden from the menu so the View menu doesn't show "Zoom In" twice.

## Acceptance criteria

- [ ] Zoom In works via both `Cmd`+`+` and `Cmd`+`=`
- [ ] The View menu does not show a confusing duplicate entry
- [ ] Zoom still clamps to the bounds in `AppConstants.Editor` (min 0.5, max 3.0) — `EditorZoom` already handles this, don't reimplement the clamping
- [ ] `LogueTests/EditorZoomTests.swift` still passes

## Pointers

- `Logue/App/LogueApp.swift` — the `CommandGroup(after: .toolbar)` block
- `Logue/Views/Editor/EditorZoom.swift` — the clamped zoom model (already unit-tested)

**Difficulty:** small — likely a handful of lines.

---

## Getting set up

1. `git clone --recurse-submodules https://github.com/bitwize-ai/Logue.git`
2. `xcodebuild -downloadComponent MetalToolchain` — one-time, required by MLX
3. `brew install xcodegen swiftformat swiftlint`
4. `make setup` — **please don't skip this.** It installs the git hooks that run SwiftFormat and SwiftLint before each commit. CI runs the same two checks, and a formatting miss fails the build.
5. `xcodegen generate`, then open `Logue.xcodeproj`

Run `xcodegen generate` again any time you add a new `.swift` file.

Requires macOS 26 (Tahoe) on Apple Silicon and Xcode 26+.

- Setup guide: [docs/dev-setup.md](docs/dev-setup.md)
- Coding standards (please read before your first PR): [CLAUDE.md](CLAUDE.md)
- Contributing guide: [CONTRIBUTING.md](CONTRIBUTING.md)

Tests use **Swift Testing** (`@Suite` / `@Test` / `#expect`), not XCTest. Run a single suite with:

```bash
xcodebuild test -project Logue.xcodeproj -scheme Logue -destination 'platform=macOS' \
  -only-testing:LogueTests/<SuiteStructName>
```

Note `-only-testing:` takes the **struct name**, not the `@Suite("display name")`.

Happy to answer questions on the issue — please comment if anything here is unclear or turns out to be wrong.

westerosweb:
Closes #9, #10, #11, #12, #13, #14, #15.

Seven issues. Five of them described a model or validator that was already in the tree and unit-tested, with nothing in the UI reaching it. One was a shortcut that never fired. One was a new block type with a worked example to follow.

## #9 — Zoom In on keyboards where `+` needs Shift

I measured this against the running app rather than reasoning about it. A key equivalent of `+` matches `⌘+` and nothing else; a key equivalent of `=` matches `⌘=` and nothing else. So both chords need separate bindings.

The alias is handled in `MainWindowView` rather than as a second menu item, because SwiftUI allows one shortcut per button and a second button makes the View menu list "Zoom In" twice.

**Verified on the running app:** `⌘=` and `⌘+` both zoom in, `⌘-` and `⌘0` unaffected, the View menu lists **one** Zoom In, and 25 presses of `⌘=` stop at `AppConstants.Editor.maxZoom` (3.0) — clamping still comes from `EditorZoom`.

`applyZoom` in `LogueApp` became `EditorZoom.mutatePersisted`, so the menu and the window share one clamped implementation instead of the menu owning a binding the window cannot reach.

## #10 — App-wide default document width

The control is in Settings → General. New documents adopt it in `draft(_:titled:spaceID:)`, the path both `createDocument` and the batch import share, so a 500-note import and a single ⌘N agree.

It is stamped only when the default is not `normal`, since `widthMode` already reads an absent value as `normal` — writing it regardless would add a redundant `width:` line to the frontmatter of every new file in markdown storage mode. Existing documents are untouched and the per-document ⋯ toggle still wins.

## #11 — Quick open (`⌘P` / `⌘O`)

**Answering the design question in the issue:** a separate surface, not a mode inside the existing palette. The command palette groups results by category and runs a debounced full-text search across bodies and transcripts; quick open is a single flat list ranked by title alone and filtered synchronously on every keystroke. Sharing one view meant threading a mode flag through the grouping, the debounce and the keyboard handling alike.

It is presented as a **sheet**, which the issue named as an acceptable shape. That was not a style preference in the end: an `.overlay` on this `NavigationSplitView` does not draw. Clicking the menu item posted the notification and flipped the state, and nothing appeared. **The same is true of the existing ⌘K command palette, which is built the same way and does not open either** — pre-existing, out of scope here, and left alone, but flagged in a code comment and worth its own issue. A sheet is also the modality quick open wants: it takes focus, and Esc dismisses it.

`⌘P` is a File menu item posting a notification rather than only an in-window key handler, so it fires while the editor's text view holds focus — a main-menu key equivalent is handled before the key event reaches any view. `⌘O` stays as the in-window binding.

All ranking is `QuickOpenMatcher`'s — there is no second matching implementation. `QuickOpenItem.candidates` takes plain arrays so that excluding trashed items, naming untitled ones and ordering by recency are tested rather than inlined in the view. Recency is not cosmetic: the matcher is stable within a rank, so the order handed to it is what breaks ties between equally good matches.

**Verified against the running app:** the menu item opens the sheet; results arrive recency-ordered with a Document/Meeting label per row and the first row selected; typing `hauz` filters 248 documents down to two, case-insensitively; Down moves the selection; Return opens the document in the editor; Esc dismisses without opening anything.

## #12 — Per-document icon picker

An emoji picker, since `DocumentIcon.sanitised` accepts exactly one grapheme — an SF Symbol name could never be a valid value, so `SpaceIconPicker` was not reusable. Every path out of the picker, grid and typed field alike, goes through `sanitised`; the reject-rather-than-truncate behaviour is unchanged.

`DocumentIconLabel` puts the icon **in place of** the document symbol in list rows and cards rather than beside it, so a document with no icon renders exactly as it does today — same glyph, same slot, no placeholder and nothing that shifts when one is set.

## #13 — Pane layout shortcuts

`⌘1` / `⌘2` / `⌘3` drive the split view's column visibility and the editor's tools sidebar, both reading `showsList` / `showsInspector` rather than re-deriving what each mode means. The menu builds its items from `1 ... allCases.count` and resolves each through `init(shortcutNumber:)`, so it cannot disagree with the model.

**A bug this turned up.** My first version made the column-visibility binding two-way, reconstructing the mode when the sidebar reappeared. Every `⌘3` pressed from editor-only landed on editor-and-list. SwiftUI echoes the new visibility back through the binding right after the menu sets a mode, and the echo arrives while the binding still reads the previous mode — `@AppStorage` caches, so a setter cannot see the write it is echoing, and the stale mode showed no inspector.

That rule now lives in `EditorLayoutMode.modeAfterVisibilityReport(listIsVisible:)`, where the reasoning for its asymmetry belongs and where tests can reach it: a visible-list report never names a mode (checked for every mode, including the one whose stale `showsInspector` caused the bug), a collapse is recorded as editor-only, and recording a collapse is idempotent because the split view echoes our own `.detailOnly` back at us.

**Verified on the running app:** all six transitions correct via the menu, and `⌘1` / `⌘2` / `⌘3` each land on the right mode. Focus Mode is untouched.

## #14 — Callout blocks

`> [!NOTE]` through `> [!CAUTION]` render with an icon, accent and heading, and the body stays editable through the same `editableTextView` a block quote uses.

Callouts are split out of the source before it reaches cmark, for the same reason `$$` fences already are: cmark reads a callout as a block quote and folds its line breaks into spaces, so a multi-line body could not be written back the way it arrived. `> [!BANANA]` is left in the markdown run and parsed as the block quote it already was; a plain `> quote` is unchanged.

**On round-tripping — narrowed after review.** A recognised callout on its own is byte-exact, including the optional title, a blank body line (`>`, not `"> "`), a header-only callout, a nested `> > inner`, and indented body lines. Two shapes are not, and review was right to push on the broader claim I made first:

- An unquoted line straight after a callout gains a blank line before it (`> [!NOTE]\n> Body\nRegular paragraph` → `> [!NOTE]\n> Body\n\nRegular paragraph`). CommonMark lazy continuation: that line belongs to the quote, and ending the callout at the last `>` separates it rather than absorbing it. The separation is what we want — cmark previously folded the paragraph into the quote and lost it as a block — but in markdown storage mode the file is the document, so this rewrites it on open. It is **idempotent**: the separated form round-trips exactly, so a file gains the blank line once, not once per save.
- An unrecognised type folds its line breaks (`> [!BANANA]\n> Body` → `> [!BANANA] Body`), because it degrades to a block quote and block quotes have always done that — `> One\n> Two` → `> One Two` on `main` too.

Both are pinned in `CalloutRoundTripEdgeTests`, alongside the plain-quote baseline that shows which half is inherited.

Reachable from the `/` menu, and from "Turn into", which also lets a callout go back to a plain quote. `CalloutKind` is closed on purpose: a type we cannot name is a type we cannot write back.

## #15 — Code block language picker

The menu is built from `CodeSyntaxHighlighter.supportedLanguages`, which lives next to the `rules(for:)` switch that consumes the same identifiers, so the offered list cannot drift from what highlighting actually exists. Aliases (`py`, `js`, `golang`) are deliberately absent — they highlight identically to their canonical spelling.

The existing free-text field stays alongside the menu, so a language that arrived in imported markdown, or one the highlighter does not know, can still be written into the fence by hand. The nice-to-have is in: the last language picked from the menu becomes the default for the next code block, and a remembered value that is no longer supported is discarded rather than written into a fence. Duplication already preserved the language; there is now a test saying so.

## Tests

813 tests across 71 suites pass (the LLM inference suites were skipped — they need a local model). SwiftFormat and SwiftLint `--strict` are both clean.

- `CalloutBlockTests` — new suite: every kind parses, titles are captured, multi-line bodies keep their breaks, exact round-trip per kind, blank body lines, header-only callouts, `[!BANANA]` degrading to a block quote without swallowing what follows, plus the block-level behaviour (`textContent`, `isEmpty`, `searchableTexts`, equality, the factory, the slash-menu entry).
- `QuickOpenCandidateTests` — new suite: both kinds included, trashed excluded on either side, recency ordering across kinds (including the interleaving a naive documents-then-meetings concatenation would get wrong), default titles for untitled items, IDs preserved so selection can open them, and matching run over the built list without reviving a trashed item.
- `EditorLayoutModeTests` — the issue pointed at `LogueTests/EditorLayoutModeTests.swift`; those six tests actually live in `DocumentIconTests.swift`. I added to them there rather than creating a second suite: every mode is reachable from a shortcut number (the View menu builds items from `allCases.count`, so a fourth case added without a number would silently go unbound), an unrecognised stored value is not a layout, and the four cases covering the write-back rule above.

## What is not verified

#9, #11 and #13 are verified end to end against the running app as described. The UI paths for #10, #12, #14 and #15 rest on unit tests and the build — their logic is covered, but I did not click through them.

One thing picked up separately: the ⌘K command palette does not open, for the same overlay reason quick open did not. It predates this branch and is now filed as #44, with the isolation steps and the reason it was deliberately not fixed here.

### DIFF
--- a/Logue/Views/CommandPalette/QuickOpenPaletteView.swift
+++ b/Logue/Views/CommandPalette/QuickOpenPaletteView.swift
@@ -1,0 +1,274 @@
+import SwiftUI
+
+/// Jump straight to a document or meeting by name — `Cmd+P` / `Cmd+O`.
+///
+/// A separate surface from `CommandPaletteView` rather than a mode inside it, because the two
+/// answer different questions and are shaped differently for it. The command palette groups
+/// results by category and runs a debounced full-text search over bodies and transcripts;
+/// quick-open is a single flat list, ranked by title only, filtered synchronously on every
+/// keystroke. Sharing one view would have meant a mode flag threaded through the grouping,
+/// the debounce and the keyboard handling alike.
+///
+/// All ranking is `QuickOpenMatcher`'s — there is no second matching implementation here.
+struct QuickOpenPaletteView: View {
+    @Binding var isPresented: Bool
+
+    @Environment(DocumentStore.self) private var documentStore
+    @Environment(MeetingStore.self) private var meetingStore
+
+    @State private var query = ""
+    @State private var selectedIndex = 0
+    @FocusState private var isSearchFocused: Bool
+
+    /// Documents and meetings, most recently modified first.
+    ///
+    /// The order matters: `QuickOpenMatcher` is stable within a rank, so the caller's ordering
+    /// is what breaks ties — recency, here. Trashed items are excluded by `activeDocuments`
+    /// and `activeMeetings`.
+    private var candidates: [QuickOpenItem] {
+        let documents = documentStore.activeDocuments.map {
+            (modifiedAt: $0.modifiedAt, item: QuickOpenItem(
+                id: $0.id,
+                title: $0.title.isEmpty ? AppConstants.defaultDocumentTitle : $0.title,
+                kind: .document
+            ))
+        }
+        let meetings = meetingStore.activeMeetings.map {
+            (modifiedAt: $0.modifiedAt, item: QuickOpenItem(
+                id: $0.id,
+                title: $0.title.isEmpty ? AppConstants.defaultMeetingTitle : $0.title,
+                kind: .meeting
+            ))
+        }
+        return (documents + meetings)
+            .sorted { $0.modifiedAt > $1.modifiedAt }
+            .map(\.item)
+    }
+
+    private var results: [QuickOpenItem] {
+        QuickOpenMatcher.match(query: query, in: candidates)
+    }
+
+    var body: some View {
+        VStack(spacing: 0) {
+            searchField
+
+            Divider()
+
+            if results.isEmpty {
+                emptyState
+            } else {
+                resultList
+            }
+
+            Divider()
+            HStack(spacing: 16) {
+                footerHint(icon: "arrow.up.arrow.down", text: "Navigate")
+                footerHint(icon: "return", text: "Open")
+                footerHint(icon: "escape", text: "Close")
+            }
+            .padding(.horizontal, 16)
+            .padding(.vertical, 8)
+        }
+        .frame(width: 520)
+        .background(
+            .regularMaterial,
+            in: RoundedRectangle(cornerRadius: AppThemeConstants.radiusXLarge, style: .continuous)
+        )
+        .overlay(
+            RoundedRectangle(cornerRadius: AppThemeConstants.radiusXLarge, style: .continuous)
+                .strokeBorder(.separator, lineWidth: 0.5)
+        )
+        .shadow(
+            color: .black.opacity(AppThemeConstants.panelShadowOpacity),
+            radius: AppThemeConstants.panelShadowRadius,
+            x: 0,
+            y: AppThemeConstants.panelShadowY
+        )
+        .onAppear {
+            isSearchFocused = true
+            selectedIndex = 0
+        }
+        .onChange(of: query) {
+            selectedIndex = 0
+        }
+        .onKeyPress(.upArrow) {
+            if selectedIndex > 0 {
+                selectedIndex -= 1
+            }
+            return .handled
+        }
+        .onKeyPress(.downArrow) {
+            if selectedIndex < results.count - 1 {
+                selectedIndex += 1
+            }
+            return .handled
+        }
+        .onKeyPress(.escape) {
+            isPresented = false
+            return .handled
+        }
+        .onKeyPress(.return) {
+            openSelected()
+            return .handled
+        }
+    }
+
+    // MARK: - Subviews
+
+    private var searchField: some View {
+        HStack(spacing: 10) {
+            Image(systemName: "arrow.right.doc.on.clipboard")
+                .font(.system(size: 15, weight: .medium))
+                .foregroundStyle(.secondary)
+
+            TextField("Go to document or meeting…", text: $query)
+                .textFieldStyle(.plain)
+                .font(.system(size: 16))
+                .focused($isSearchFocused)
+                .accessibilityLabel("Quick open search")
+                .accessibilityHint("Type part of a document or meeting title")
+                .onSubmit { openSelected() }
+
+            if !query.isEmpty {
+                Button {
+                    query = ""
+                } label: {
+                    Image(systemName: "xmark.circle.fill")
+                        .foregroundStyle(.tertiary)
+                }
+                .buttonStyle(.plain)
+                .accessibilityLabel("Clear search")
+                .accessibilityHint("Clears the quick open search text")
+            }
+        }
+        .padding(.horizontal, 16)
+        .padding(.vertical, 12)
+    }
+
+    private var emptyState: some View {
+        VStack(spacing: 8) {
+            Image(systemName: "magnifyingglass")
+                .font(.title2)
+                .foregroundStyle(.tertiary)
+            Text(candidates.isEmpty ? "Nothing to open yet" : "No matches for \"\(query)\"")
+                .font(.subheadline)
+                .foregroundStyle(.secondary)
+        }
+        .frame(maxWidth: .infinity, minHeight: 120)
+    }
+
+    private var resultList: some View {
+        ScrollViewReader { proxy in
+            ScrollView {
+                LazyVStack(alignment: .leading, spacing: 0) {
+                    ForEach(Array(results.enumerated()), id: \.element.id) { index, item in
+                        QuickOpenRow(item: item, isSelected: index == selectedIndex) {
+                            open(item)
+                        }
+                        .id(item.id)
+                    }
+                }
+                .padding(.vertical, 4)
+            }
+            .frame(maxHeight: 340)
+            .onChange(of: selectedIndex) { _, newIndex in
+                // Bounds-checked because the result set shrinks as the query grows, and the
+                // index can be stale for a frame after a keystroke.
+                guard newIndex >= 0, newIndex < results.count else { return }
+                proxy.scrollTo(results[newIndex].id)
+            }
+        }
+    }
+
+    private func footerHint(icon: String, text: String) -> some View {
+        HStack(spacing: 4) {
+            Image(systemName: icon)
+                .font(.caption2)
+            Text(text)
+                .font(.caption2)
+        }
+        .foregroundStyle(.tertiary)
+    }
+
+    // MARK: - Opening
+
+    private func openSelected() {
+        guard selectedIndex >= 0, selectedIndex < results.count else { return }
+        open(results[selectedIndex])
+    }
+
+    /// Selecting an item opens it: `MainWindowView` observes both stores' selection and
+    /// switches to the editor or the meeting workspace, so setting the ID is the whole action.
+    private func open(_ item: QuickOpenItem) {
+        isPresented = false
+        switch item.kind {
+        case .document:
+            meetingStore.selectedMeetingID = nil
+            documentStore.selectedDocumentID = item.id
+        case .meeting:
+            documentStore.selectedDocumentID = nil
+            meetingStore.selectedMeetingID = item.id
+        }
+    }
+}
+
+// MARK: - Row
+
+private struct QuickOpenRow: View {
+    let item: QuickOpenItem
+    let isSelected: Bool
+    let action: () -> Void
+    @State private var isHovered = false
+
+    private var kindLabel: String {
+        switch item.kind {
+        case .document: "Document"
+        case .meeting: "Meeting"
+        }
+    }
+
+    private var kindIcon: String {
+        switch item.kind {
+        case .document: "doc.text"
+        case .meeting: "waveform"
+        }
+    }
+
+    var body: some View {
+        Button(action: action) {
+            HStack(spacing: 10) {
+                Image(systemName: kindIcon)
+                    .font(.system(size: 14, weight: .medium))
+                    .foregroundStyle(isSelected ? .primary : .secondary)
+                    .frame(width: 24, height: 24)
+
+                Text(item.title)
+                    .font(.system(size: 14))
+                    .foregroundStyle(.primary)
+                    .lineLimit(1)
+
+                Spacer()
+
+                Text(kindLabel)
+                    .font(.caption2)
+                    .foregroundStyle(.tertiary)
+            }
+            .padding(.horizontal, 16)
+            .padding(.vertical, 6)
+            .background(
+                RoundedRectangle(cornerRadius: 6, style: .continuous)
+                    .fill(isSelected
+                        ? AppThemeConstants.accent.opacity(AppThemeConstants.opacityMedium)
+                        : (isHovered ? Color.primary.opacity(AppThemeConstants.opacityLight) : Color.clear))
+                    .padding(.horizontal, 8)
+            )
+            .contentShape(Rectangle())
+        }
+        .buttonStyle(.plain)
+        .onHover { isHovered = $0 }
+        .accessibilityLabel("\(item.title), \(kindLabel)")
+        .accessibilityHint("Opens this \(kindLabel.lowercased())")
+        .accessibilityAddTraits(isSelected ? .isSelected : [])
+    }
+}
--- a/Logue/Views/CommandPalette/QuickOpenPaletteView.swift
+++ b/Logue/Views/CommandPalette/QuickOpenPaletteView.swift
@@ -22,27 +22,10 @@ struct QuickOpenPaletteView: View {
 
     /// Documents and meetings, most recently modified first.
     ///
-    /// The order matters: `QuickOpenMatcher` is stable within a rank, so the caller's ordering
-    /// is what breaks ties — recency, here. Trashed items are excluded by `activeDocuments`
-    /// and `activeMeetings`.
+    /// The raw arrays go in rather than `activeDocuments` / `activeMeetings`, so that excluding
+    /// trashed items is `QuickOpenItem.candidates`' job and therefore a tested one.
     private var candidates: [QuickOpenItem] {
-        let documents = documentStore.activeDocuments.map {
-            (modifiedAt: $0.modifiedAt, item: QuickOpenItem(
-                id: $0.id,
-                title: $0.title.isEmpty ? AppConstants.defaultDocumentTitle : $0.title,
-                kind: .document
-            ))
-        }
-        let meetings = meetingStore.activeMeetings.map {
-            (modifiedAt: $0.modifiedAt, item: QuickOpenItem(
-                id: $0.id,
-                title: $0.title.isEmpty ? AppConstants.defaultMeetingTitle : $0.title,
-                kind: .meeting
-            ))
-        }
-        return (documents + meetings)
-            .sorted { $0.modifiedAt > $1.modifiedAt }
-            .map(\.item)
+        QuickOpenItem.candidates(documents: documentStore.documents, meetings: meetingStore.meetings)
     }
 
     private var results: [QuickOpenItem] {
--- a/Logue/Views/Editor/BlockEditor/Model/BlockSerializer.swift
+++ b/Logue/Views/Editor/BlockEditor/Model/BlockSerializer.swift
@@ -22,7 +22,17 @@ enum BlockSerializer {
             case let .math(latex):
                 blocks.append(.math(id: UUID(), latex: latex))
             case let .markdown(text):
-                blocks += parseMarkdownSegment(text)
+                // Callouts get pulled out for the same reason as math fences: cmark reads one
+                // as a block quote and folds its line breaks into spaces, so a multi-line
+                // callout body could not be written back the way it arrived.
+                for calloutSegment in calloutSegments(in: text) {
+                    switch calloutSegment {
+                    case let .callout(kind, title, body):
+                        blocks.append(.callout(id: UUID(), kind: kind, title: title, body: body))
+                    case let .markdown(plain):
+                        blocks += parseMarkdownSegment(plain)
+                    }
+                }
             }
         }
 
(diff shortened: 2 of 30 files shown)


Logue/Views/Editor/BlockEditor/Model/BlockSerializer.swift
```diff
@@ -99,6 +109,91 @@ enum BlockSerializer {
...
+        while index < lines.count {
+            guard let header = calloutHeader(in: lines[index]) else {
+                pending.append(lines[index])
+                index += 1
+                continue
+            }
+
+            // Body lines are the quote lines that follow, stopping at the first line that is
+            // not quoted — a blank line ends the quote in CommonMark, and a non-quote line
+            // starts a new block.
```
shanforge, line 160: I built the callout parser into a scratch suite and ran your shapes plus a few you didn't list. Everything you claimed holds — `> [!NOTE]\n> Body`, titled callouts, blank body lines (`>` not `"> "`), header-only, nested `> > inner`, and indented body lines all round-trip byte-exact. The one-space rule in `quotedContent` does what its comment says.

Two shapes come back changed, and I don't think either is a blocker — but **"Round-trip is byte-exact"** in the description is broader than what's true, and this is the file where someone will rely on it.

**1. A paragraph on the line straight after a callout gains a blank line.**

```
in : "> [!NOTE]\n> Body\nRegular paragraph"
out: "> [!NOTE]\n> Body\n\nRegular paragraph"
```

That's CommonMark lazy continuation: the unquoted line is *part of* the quote, and splitting the callout out first ends the block at the last `>`. Worth knowing that the previous behaviour was **worse** — cmark folded the lot into `> [!NOTE] Body Regular paragraph` — so this is an improvement, not a regression. But in markdown storage mode the file is the document, so opening and saving rewrites the user's file, and that's the kind of thing that deserves a sentence in the doc comment rather than being discovered.

**2. `[!BANANA]` folds its line break**, because it degrades to a block quote and that's what quotes do:

```
in : "> [!BANANA]\n> Body"
out: "> [!BANANA] Body"
```

I checked whether that's yours: it isn't. A plain `> Quoted\nRegular paragraph` becomes `> Quoted Regular paragraph` on `main` too. Pre-existing cmark behaviour, and consistent with your stated design — just not byte-exact, which the description implies for everything.

Suggest narrowing the claim to the recognised kinds, and adding a line here noting that an unquoted line after a callout is separated rather than absorbed.

Logue/Views/Editor/EditorLayoutMode.swift
```diff
@@ -30,4 +30,31 @@ enum EditorLayoutMode: String, CaseIterable, Codable, Sendable {
...
+    /// Only a collapse is inferred, and that asymmetry is the whole point. SwiftUI echoes the
+    /// new visibility back through its `columnVisibility` binding immediately after a menu item
+    /// sets a mode, and the echo arrives while the binding still reads the previous mode —
+    /// `@AppStorage` caches, so a setter cannot see the write it is echoing. A rule that tried
+    /// to name the mode for a *re-appearing* list therefore read the old mode's
+    /// `showsInspector`, which is `false` in editor-only, and turned every ⌘3 pressed from
+    /// editor-only into editor-and-list. Nothing needs that direction inferred: every path that
+    /// shows the list again already names the mode it wants.
+    static func modeAfterVisibilityReport(listIsVisible: Bool) -> EditorLayoutMode? {
+        listIsVisible ? nil : .editorOnly
```
shanforge, line 52: This is the best-explained thing in the PR. The asymmetry is genuinely subtle — `@AppStorage` caching means the setter can't see the write it's echoing — and putting the rule in a testable function with the bug written into the doc comment is exactly where it belongs. The test that covers *every* mode for the visible-list case, rather than just the one that broke, is what stops it regressing.

One behaviour worth confirming is intended rather than incidental: because there's no editor-plus-inspector mode, dragging the sidebar closed while in `allPanels` reports a collapse and lands on `.editorOnly`, which also hides the inspector. Consistent with a 1/2/3 progression, and I don't think there's a better answer without a fourth case — but a user who drags the divider to get more room may be surprised to lose the inspector too.

Not asking for a change; flagging it so the answer is on the record.

Logue/Views/MainWindowView.swift
```diff
@@ -218,11 +253,40 @@ struct MainWindowView: View {
...
+        // on this `NavigationSplitView` does not draw: clicking the menu item posted the
+        // notification and flipped the state, and nothing appeared. The same is true of the
+        // ⌘K palette, which is why it does not open either — a separate problem, left alone
+        // here. The issue for this work named a sheet as an acceptable shape, and a sheet is
+        // also the modality quick-open wants: it takes focus, and Esc dismisses it.
+        .sheet(isPresented: $showQuickOpen) {
+            QuickOpenPaletteView(isPresented: $showQuickOpen)
+        }
         .onKeyPress(keys: [KeyEquivalent("k")], phases: .down) { keyPress in
             guard keyPress.modifiers.contains(.command) else { return .ignored }
```
shanforge, line 268: Worth stating plainly for the record, because it's easy to read this as a style choice: `modifiers == .command` is exact equality here while the `⌘K` handler above uses `.contains(.command)`, and the difference is load-bearing. `⌘⇧=` **is** `⌘+` on a US layout, and the View menu already binds that. Using `.contains` here would have both this handler and the menu fire on the same chord.

I didn't verify the key-equivalent matching on a real keyboard — you did, and the reasoning holds up. Leaving it as-is.

Logue/Views/CommandPalette/QuickOpenPaletteView.swift
```diff
@@ -0,0 +1,257 @@
...
+        .shadow(
+            color: .black.opacity(AppThemeConstants.panelShadowOpacity),
+            radius: AppThemeConstants.panelShadowRadius,
+            x: 0,
+            y: AppThemeConstants.panelShadowY
+        )
+        .onAppear {
+            isSearchFocused = true
+            selectedIndex = 0
+        }
```
shanforge, line 74: Selection handling is right, and I checked the failure mode this class of code usually has: both subscripts (`proxy.scrollTo` and `openSelected`) guard `selectedIndex < results.count`, and the index resets on query change and on appear. So results shrinking under a stale selection can't crash or open the wrong item. That's the CLAUDE.md bounds rule being followed rather than recited.

One observation, not a request. `results` is a computed property, so `QuickOpenMatcher.match` runs on every body evaluation — several times per keystroke — and `rank` calls `item.title.lowercased()` per item per pass. At 248 documents that's invisible, which matches what you measured. At an order of magnitude more it becomes a per-keystroke allocation of every title. If it ever feels sluggish, lowercasing once when candidates are built is the cheap fix; the matcher's shape doesn't need to change.

shanforge (approved):
Reviewed this against the running code rather than the description: checked out the branch, built it, ran the full suite, ran the CI-pinned linters, and wrote a scratch suite to drive the callout parser through shapes the PR doesn't list. Approving — the substance holds up, and the two things I found are a documentation overreach and an observation, neither of which should hold this.

**What I verified independently**

- **907 tests, 87 suites.** 23 issues, every one in `GrammarClarityLLMTests`, `MeetingLLMTests`, `ReviewPanelLLMTests`, `VerifyPanelLLMTests`, `VocabularyLLMTests` — the real-model suites, which fail the same way on my own branches. **Zero non-LLM failures.** `CalloutBlocks`, `QuickOpenCandidates` and `EditorLayoutMode` all pass.
- **SwiftFormat 0.62.1 and SwiftLint 0.65.0 `--strict`**, the CI-pinned versions: both clean, no new suppressions.
- **The callout round-trip**, driven through the real `BlockSerializer` rather than read. Details inline — your listed shapes are exact; two unlisted ones are not.
- **`Block` is not `Codable`.** I checked this specifically, because adding a case to a persisted enum is how you break every existing document on upgrade. It's a transient view model serialised through markdown, so there's no migration risk. Worth having confirmed rather than assumed.
- **`EditorZoom.scale`'s setter clamps**, so `mutatePersisted` reading a stored value can't produce an unclamped zoom. The claim that clamping stays in one place is true.

**What I liked**

`modeAfterVisibilityReport` is the standout. A SwiftUI echo arriving against a stale `@AppStorage` read is the kind of bug that gets fixed with a flag and a shrug; extracting the rule, writing the reasoning into the doc comment, and testing every mode rather than only the one that broke is what stops it coming back.

The reasoning for a separate quick-open surface is right, and the honesty about the overlay is more valuable than the feature — discovering that `⌘K` has never opened, saying so, and *not* fixing it here is the correct call on all three counts.

Splitting callouts before cmark for the same reason `$$` already is means the change follows an established shape rather than inventing one.

**On what you didn't verify**

You were explicit that #10, #12, #14 and #15 rest on unit tests and the build. That's the right thing to write down. Their logic is covered and the risky part of #14 — the parser — I've now exercised directly, which is the piece I'd have worried about most.

The `⌘K` palette not opening deserves its own issue; happy to file it if you haven't.

Logue/Views/Editor/BlockEditor/Model/BlockSerializer.swift
westerosweb: Both confirmed against the real serializer before I touched anything, and you're right on both counts — including that the second one isn't mine.

```
in : "> [!NOTE]\n> Body\nRegular paragraph"
out: "> [!NOTE]\n> Body\n\nRegular paragraph"     blocks: callout + paragraph

in : "> [!BANANA]\n> Body"
out: "> [!BANANA] Body"                          blocks: blockQuote

baseline, no callouts involved:
in : "> One\n> Two"          out: "> One Two"
in : "> Quoted\nRegular..."  out: "> Quoted Regular paragraph"
```

The baseline is the part I'd have missed: a plain two-line block quote has never round-tripped exactly, so "byte-exact" was overclaiming for block quotes generally, not just for unrecognised markers.

Fixed in db4e891:

- **The claim is narrowed** — in `calloutSegments`' doc comment, in `serializeCallout`, in the `CalloutBlockTests` suite header, and in the PR description. All four now say a *recognised callout on its own* is exact, and point at the exceptions.
- **The lazy-continuation rewrite is written down** where you asked for it, with the reason the separation is preferable (cmark used to fold the paragraph away entirely) and the reason it matters anyway (in markdown storage mode the file is the document, so opening rewrites it).
- **`CalloutRoundTripEdgeTests`** pins both shapes, asserts the plain-quote baseline next to the `[!BANANA]` fold so the inherited half is visible, and adds one you didn't ask for but which decides whether the rewrite is tolerable: **it's idempotent.** The separated form round-trips exactly on a second pass, so the file gains a blank line once rather than one per save. That was the thing I actually wanted to know before leaving the behaviour in.

Your three exact shapes — blank-line-separated, nested `> > inner`, indented body lines — are in that suite too, as the boundary of what's guaranteed.

Logue/Views/Editor/EditorLayoutMode.swift
westerosweb: Intended, and you're right that it should be on the record rather than inferred.

The reasoning: there's no editor-plus-inspector rung, so "no list" can only resolve to the narrowest mode. Special-casing it here would mean the write-back inventing a mode that doesn't exist — which is the same class of mistake as the stale read this function was extracted to prevent. If losing the inspector on a divider drag turns out to bother anyone, the honest fix is a fourth case, not a branch in this rule.

db4e891 adds that as a paragraph on the doc comment, and asserts it rather than leaving it incidental:

```swift
@Test("Collapsing from all-panels also gives up the inspector")
func collapseFromAllPanelsDropsTheInspector() {
    #expect(EditorLayoutMode.allPanels.showsInspector)
    let afterCollapse = EditorLayoutMode.modeAfterVisibilityReport(listIsVisible: false)
    #expect(afterCollapse == .editorOnly)
    #expect(afterCollapse?.showsInspector == false)
}
```

Worth noting the recovery is cheap — ⌘3 brings both panes back in one keystroke — which is part of why a fourth case didn't seem worth adding pre-emptively.

Logue/Views/MainWindowView.swift
westerosweb: Correct, and worth having in the code rather than only in the review thread — someone tidying for consistency would have "fixed" it into a bug. db4e891 adds:

```
// The exact `==` is load-bearing, not a stylistic difference from the `.contains` used
// for ⌘K below. On a US layout ⌘⇧= *is* ⌘+, which the View menu already binds; with
// `.contains(.command)` that one chord would fire the menu item and this handler both,
// and zoom would jump two steps.
```

On the key-equivalent matching, for the record since you didn't verify it yourself: I measured it with `key code 24` against the running app, after finding that AppleScript's `keystroke "="` doesn't reliably deliver that chord at all — which had me draw a wrong conclusion earlier in this branch's history before I switched to raw key codes. With the binding set to `"+"`, ⌘⇧= fired and ⌘= did nothing; with it set to `"="`, exactly the reverse. Hence two bindings.

Logue/Views/CommandPalette/QuickOpenPaletteView.swift
westerosweb: Taking this as an observation and leaving the code alone, deliberately rather than by omission — so here's the decision on record.

You're right about the shape: `results` is computed, so `match` runs several times per body evaluation, and `rank` lowercases every title on each pass. At 248 documents it's invisible, which matches what I saw driving it.

Two reasons not to change it now. The cheap fix you name — lowercasing once at candidate build — means `QuickOpenItem` carrying a second title field that exists only for the matcher, which then has to prefer it; that's a wider change than it sounds, and `QuickOpenMatcher` is the one piece of this feature that arrived already tested. And the memoising alternative trades an O(n) pass for `@State` that has to stay in sync with the query, which is a worse class of bug than a per-keystroke allocation.

The threshold I'd act on: it's the *number of items*, not the query, that drives this, so a library in the low thousands with a slow-feeling palette is the signal. At that point lowercasing at build time is right, and `QuickOpenItem.candidates` is now the single place it would go — which is a side benefit of having pulled it out of the view for testability.

Also: thank you for checking the bounds guards against the actual failure mode rather than reading for the pattern. `selectedIndex` surviving a shrinking result set is exactly where this kind of code breaks.

westerosweb:
Thanks — the independent verification is the part I valued most, especially checking that `Block` isn't `Codable` and that `EditorZoom.scale`'s setter clamps. Both were assumptions I'd have been embarrassed to have wrong.

Everything actionable is in db4e891. Replies are inline on each thread; the summary:

**Your two round-trip findings were both right, and one of them was mine to fix.** I reproduced both against the real serializer before changing anything. The lazy-continuation rewrite is now documented where you asked, with the reason the separation beats folding and the reason it matters anyway. `CalloutRoundTripEdgeTests` pins both shapes plus the plain-quote baseline that shows the `[!BANANA]` fold is inherited — and one thing you didn't ask for that decided whether I was comfortable leaving the behaviour: **the rewrite is idempotent**, so a file gains the blank line once rather than once per save. The claim is narrowed in four places, including the PR description.

**The two you flagged as questions rather than changes are now answered in the code** — the all-panels collapse giving up the inspector, with a test asserting it rather than leaving it incidental, and why `==` is load-bearing on the ⌘= handler where `.contains` is fine for ⌘K. That second one especially: as it stood, someone tidying for consistency would have turned it into a bug.

**The perf observation I'm leaving alone**, with the reasoning and the threshold I'd act on written into the thread rather than just declining.

**⌘K is filed as #44**, with the isolation steps — including that a menu-driven notification *does* reach `MainWindowView`, which is what narrowed it from "dead shortcut" to "the overlay doesn't draw" — and the reason it deliberately isn't fixed here.

One correction to my own earlier verification, since it's the kind of thing that should be said out loud: partway through this I ran a check on the wrong branch (the working tree had been switched to `fix/editor-scroll-performance`) and briefly concluded your first finding didn't reproduce. It does, exactly as you described. I re-ran it on the right branch before touching anything.

823 tests across 72 suites, both linters clean.

Issue #42, opened 2026-07-30
Collect block frames only during drags, not on every scrolled frame


westerosweb:
## Summary

Every block in the document editor reports its frame through a SwiftUI preference on every layout pass. Because the frames are measured in the scroll coordinate space, *all* of them change on every frame of a scroll, so the cost of this bookkeeping grows with document length even though the frames are only ever read during a cross-block drag. Collecting them only while a drag is in progress would make scrolling cost independent of document size.

This is a follow-up to the scroll-performance fix that moved these frames out of observed view state. That change removed the expensive part — a full editor body rebuild per scrolled frame — but deliberately left the collection itself untouched to keep the diff small. This issue covers the remainder.

## Background

Each row publishes its frame via a `GeometryReader` in a `.background`:

- [Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift:6](Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift#L6) — `BlockFramePreferenceKey`
- [Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift:93-99](Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift#L93-L99) — per-row `GeometryReader` writing the preference
- [Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift:115-117](Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift#L115-L117) — `coordinateSpace` + `onPreferenceChange`
- [Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift:765](Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift#L765) — `blockAtPoint`, the only consumer

The only caller of `blockAtPoint` is the `leftMouseDragged` monitor that implements multi-block selection. Nothing renders from these frames.

## Evidence

Layout passes per second observed during an identical synthetic scroll gesture, after the frame-store fix, on two generated test documents:

| Document | Layout passes/sec while scrolling |
| --- | --- |
| 45 blocks (916 words) | ~40–55 |
| 450 blocks (9,751 words) | ~11–20 |

The shorter document reaches the bottom of its content partway through the gesture and still records roughly 3× more passes than the long one. That asymmetry is the tell: the per-pass cost scales with block count.

To reproduce, instrument the `onPreferenceChange` closure with a counter and log it on a one-second timer, then scroll a short document and a long one and compare rates.

## Suggested approach

[...]

PR #43, opened 2026-07-30
Stop rebuilding every block on every scrolled frame

westerosweb:
## Problem

Scrolling a document stuttered, and it stuttered whether the document was small or large. Idle was fine — the cost only appeared while scrolling.

Every block published its frame through a SwiftUI preference, measured in the `editorScroll` coordinate space. Scrolling changes *every* block's frame, so on each layout pass:

preference recomputes → `onPreferenceChange` writes `@State blockFrames` → invalidates `BlockEditorView.body` → rebuilds **all** rows and runs `updateNSView` on **all** block text views → produces the next layout pass → repeat.

The scroll fed itself. The tell, when instrumented, was that body evaluations and preference callbacks came out to exactly the same count: every single body evaluation was caused by that state write.

## Why the frames don't belong in view state

`blockFrames` had exactly one consumer — `blockAtPoint`, reached from the `leftMouseDragged` monitor that implements cross-block selection. Nothing renders from it. Writing it therefore should not invalidate anything, so it moves into a plain reference type (`BlockFrameStore`) held by `@State` — the reference is stable, so SwiftUI sees no change.

## Measurements

Instrumented body evaluations, row builds and `updateNSView` calls, then drove identical synthetic scroll gestures over two generated documents.

**45 blocks / 916 words** — a small document, the case that should never have been slow:

| Metric | Before | After |
| --- | --- | --- |
| body evaluations/sec | 48 | **0** |
| row builds/sec | 2,208 | **0** |
| `updateNSView` calls/sec | 2,208 | **~18** |

**450 blocks / 9,751 words:**

| Metric | Before | After |
| --- | --- | --- |
| layout passes/sec | **4–6** (≈5fps) | 17–20 |
| row builds/sec | ~2,700 | **0** |
| `updateNSView` calls/sec | ~2,700 | **~18** |

Idle was 0 in both builds, which is why this only showed up in motion. On the long document the one-second sampling timer itself slipped to 1.24s before the change — the main thread was saturated, which is what the stutter was.

## Behaviour

The hit test returns the same block as before for the non-overlapping frames the editor actually produces. The one deliberate difference: overlapping frames now resolve to the topmost block instead of whichever one dictionary iteration order happened to surface first, so the result is deterministic.

Cross-block drag selection was exercised with the same synthetic drag against both the original and the patched build and behaves identically.

## Testing

- `LogueTests/BlockFrameStoreTests.swift` — 6 tests covering containment, x-independence, misses above and below content, the empty store, deterministic overlap resolution, and that replacing frames discards the previous layout. All passing.
- `swiftlint` and `swiftformat` clean on the changed files.
- Verified interactively on both document sizes, plus a drag-selection comparison against `main`.

## Follow-up, not included here

The per-block `GeometryReader` still collects frames on every layout pass, so that bookkeeping remains O(number of blocks) per frame — it is now the only thing in the scroll path that scales with document length. Comparing post-fix layout passes per second between the two documents above (~40–55/s at 45 blocks vs ~11–20/s at 450) shows the remaining cost. That is tracked separately in #42 and is deliberately out of scope here to keep this diff small.

Also noticed while in this code, and **not** addressed: `blockAtPoint` compares a point in the window's `contentView` coordinates against frames captured in the `editorScroll` coordinate space. Those spaces do not share an origin, so cross-block drag selection appears not to trigger in practice. Pre-existing, unchanged by this PR, and noted in #42.

### DIFF
--- a/Logue/Views/Editor/BlockEditor/Model/BlockFrameStore.swift
+++ b/Logue/Views/Editor/BlockEditor/Model/BlockFrameStore.swift
@@ -29,13 +29,20 @@ final class BlockFrameStore: @unchecked Sendable {
     /// Hit-tests which block sits at the given point.
     ///
     /// Vertical-only: blocks span the full content width, so the x coordinate carries no
-    /// information. Ties are impossible in practice because block frames don't overlap
-    /// vertically, but a deterministic answer still beats an arbitrary one, so the topmost
-    /// match wins.
+    /// information.
+    ///
+    /// The bounds are inclusive at both ends, so any two frames that touch or overlap both
+    /// contain the point on their shared edge. The editor's current layout does not produce
+    /// that case — its `VStack` spacing leaves a measured 4pt gap between consecutive row
+    /// frames, and a point inside a gap matches nothing — but the store cannot assume a
+    /// caller's frames are disjoint, and dropping that spacing to zero would create a tie at
+    /// every boundary. Resolving to the topmost match keeps the answer stable either way;
+    /// picking whichever key dictionary iteration happened to yield first would not.
     func blockID(at point: CGPoint) -> BlockID? {
         lock.lock()
         defer { lock.unlock() }
-        return frames
+        // `lazy` so the filter doesn't build an intermediate dictionary on every hit test.
+        return frames.lazy
             .filter { point.y >= $0.value.minY && point.y <= $0.value.maxY }
             .min { $0.value.minY < $1.value.minY }?
             .key
--- a/Logue/Views/Editor/BlockEditor/Model/BlockFrameStore.swift
+++ b/Logue/Views/Editor/BlockEditor/Model/BlockFrameStore.swift
@@ -1,0 +1,43 @@
+import Foundation
+
+// MARK: - BlockFrameStore
+
+/// Holds each block's on-screen frame so a cross-block drag can hit-test which block the
+/// pointer is over.
+///
+/// Deliberately a plain reference type rather than `@State` storage or `@Observable`.
+/// Every block reports its frame in the scroll coordinate space, so *all* of those frames
+/// change on every frame of a scroll. Keeping them in observed view state meant each scrolled
+/// frame invalidated the whole editor body, which rebuilt every block row and ran
+/// `updateNSView` on every block's text view — measured at ~48 body evaluations and ~2,200
+/// text-view updates per second on a 45-block document, which is what made scrolling stutter.
+/// Nothing renders from these frames, so writing them must not invalidate anything.
+///
+/// Thread-safety: every access goes through `lock`. The frames are written from SwiftUI's
+/// preference callback and read from an `NSEvent` monitor; both run on the main thread today,
+/// but the lock means that isn't load-bearing.
+final class BlockFrameStore: @unchecked Sendable {
+    private let lock = NSLock()
+    private var frames: [BlockID: CGRect] = [:]
+
+    func replaceAll(with newFrames: [BlockID: CGRect]) {
+        lock.lock()
+        defer { lock.unlock() }
+        frames = newFrames
+    }
+
+    /// Hit-tests which block sits at the given point.
+    ///
+    /// Vertical-only: blocks span the full content width, so the x coordinate carries no
+    /// information. Ties are impossible in practice because block frames don't overlap
+    /// vertically, but a deterministic answer still beats an arbitrary one, so the topmost
+    /// match wins.
+    func blockID(at point: CGPoint) -> BlockID? {
+        lock.lock()
+        defer { lock.unlock() }
+        return frames
+            .filter { point.y >= $0.value.minY && point.y <= $0.value.maxY }
+            .min { $0.value.minY < $1.value.minY }?
+            .key
+    }
+}
--- a/Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift
+++ b/Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift
@@ -59,7 +59,9 @@ struct BlockEditorView: View {
 
     // Extension-visible: +Reorder
     @State var multiBlockSelection = MultiBlockSelectionState()
-    @State private var blockFrames: [BlockID: CGRect] = [:]
+    /// Block frames for drag hit-testing. Held in a store rather than in view state because
+    /// they change on every frame of a scroll — see `BlockFrameStore`.
+    @State private var blockFrameStore = BlockFrameStore()
     @State private var dragMonitor: Any?
     @State private var mouseUpMonitor: Any?
     @State private var isDraggingAcrossBlocks = false
@@ -113,8 +115,8 @@ struct BlockEditorView: View {
                         .padding(.vertical, AppThemeConstants.editorVerticalInset)
                     }
                     .coordinateSpace(name: "editorScroll")
-                    .onPreferenceChange(BlockFramePreferenceKey.self) { newFrames in
-                        blockFrames = newFrames
+                    .onPreferenceChange(BlockFramePreferenceKey.self) { [blockFrameStore] newFrames in
+                        blockFrameStore.replaceAll(with: newFrames)
                     }
                     .onScrollPhaseChange { _, newPhase in
                         // Dismiss floating selection toolbar when scrolling starts
@@ -763,16 +765,7 @@ extension BlockEditorView {
 
     /// Hit-tests which block is at the given point (in content view coordinates).
     private func blockAtPoint(_ point: CGPoint) -> BlockID? {
-        // Convert content view point to scroll coordinate space
-        // blockFrames are in the "editorScroll" coordinate space
-        // We need to find which block's frame contains the point
-        for (blockID, frame) in blockFrames {
-            // Use a vertical-only hit test (ignore X since blocks span full width)
-            if point.y >= frame.minY, point.y <= frame.maxY {
-                return blockID
-            }
-        }
-        return nil
+        blockFrameStore.blockID(at: point)
     }
 
     /// Selects all blocks in the contiguous range between two block IDs.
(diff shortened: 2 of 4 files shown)


Logue/Views/Editor/BlockEditor/Model/BlockFrameStore.swift
```diff
@@ -0,0 +1,43 @@
...
+        lock.lock()
+        defer { lock.unlock() }
+        frames = newFrames
+    }
+
+    /// Hit-tests which block sits at the given point.
+    ///
+    /// Vertical-only: blocks span the full content width, so the x coordinate carries no
+    /// information. Ties are impossible in practice because block frames don't overlap
+    /// vertically, but a deterministic answer still beats an arbitrary one, so the topmost
```
shanforge: Small correction to the reasoning, not the code. **Ties aren't rare here — they happen at every block boundary.**

The bounds are inclusive at both ends (`>= minY`, `<= maxY`), and the editor lays rows out adjacent, so a block ending at y=100 and the next starting at y=100 both contain y=100. I built a probe suite against the real store to check:

- a point exactly on a boundary matches **two** blocks and resolves to the upper one
- stable across 200 repeated lookups on the same store
- stable across 50 freshly built stores, so dictionary seeding doesn't leak through
- genuinely overlapping frames resolve topmost, as you describe

So the behaviour is right and the determinism claim is right — it's just that the comment says ties can't happen when they're guaranteed once per boundary. That matters because it makes the `min` look like defensive dead code rather than the thing doing real work on every drag. Worth rewording to something like "adjacent frames share their boundary, so a point on the edge matches both; the topmost wins."

This also makes the change a slightly bigger behavioural improvement than the description claims: before, a drag landing exactly on a boundary picked whichever block dictionary iteration surfaced first.

Logue/Views/Editor/BlockEditor/Model/BlockFrameStore.swift
```diff
@@ -0,0 +1,43 @@
...
+    /// match wins.
+    func blockID(at point: CGPoint) -> BlockID? {
+        lock.lock()
+        defer { lock.unlock() }
+        return frames
+            .filter { point.y >= $0.value.minY && point.y <= $0.value.maxY }
+            .min { $0.value.minY < $1.value.minY }?
```
shanforge, line 47: Aside, since this is a performance PR and it'd be odd not to mention it: `filter` on a `Dictionary` returns a new `Dictionary`, so every hit test allocates one. That's once per `leftMouseDragged`, not per scrolled frame — nowhere near the path you fixed, and invisible next to what this removes.

A single pass would avoid it:

```swift
frames.lazy
    .filter { point.y >= $0.value.minY && point.y <= $0.value.maxY }
    .min { $0.value.minY < $1.value.minY }?
    .key
```

Entirely optional — flagging it rather than asking for it.

Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift
```diff
@@ -113,8 +115,8 @@ struct BlockEditorView: View {
                         .padding(.vertical, AppThemeConstants.editorVerticalInset)
                     }
                     .coordinateSpace(name: "editorScroll")
-                    .onPreferenceChange(BlockFramePreferenceKey.self) { newFrames in
-                        blockFrames = newFrames
+                    .onPreferenceChange(BlockFramePreferenceKey.self) { [blockFrameStore] newFrames in
```
shanforge, line 118: I checked the mechanism rather than taking the numbers on faith, and it holds: `BlockFrameStore` is a plain `final class` with no `@Observable`, so writing through the reference leaves the `@State` value — the reference itself — unchanged, and SwiftUI has nothing to invalidate. That's the whole fix, and it's the right shape.

I also verified the premise by grep: `blockFrames` had exactly one reader, `blockAtPoint` at line 767, reached only from the drag monitor. Nothing renders from it, so this genuinely is state that never needed to be observed.

The explicit `[blockFrameStore]` capture is worth keeping even though it reads as redundant — it captures the reference rather than the enclosing view, which is what stops the closure retaining more than it needs.

The thread-safety note carries its weight too: `@unchecked Sendable` justified by an actual lock, and honest that main-thread-only isn't being relied on. That's the CLAUDE.md rule being followed rather than recited.

Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift
```diff
@@ -763,16 +765,7 @@ extension BlockEditorView {
 
     /// Hit-tests which block is at the given point (in content view coordinates).
     private func blockAtPoint(_ point: CGPoint) -> BlockID? {
```
shanforge, line 767: Confirming the coordinate-space problem you flagged, because it's the thing a future reader will trip over here.

Frames are captured at line 97 in `.named("editorScroll")` — SwiftUI's scroll space, origin at the content's top-left, y increasing downward. This receives `contentView.convert(windowPoint, from: nil)`, which is AppKit contentView coordinates: different origin, and `NSView` isn't flipped by default, so y runs the other way. The two are not comparable, and cross-block drag selection can't work reliably as a result.

You were right to leave it. This PR is a behaviour-preserving performance change, and "identically broken" is the correct outcome for it — fixing the hit test here would have made the diff about two things and made the performance claim harder to trust.

Worth being explicit about one consequence though: `BlockFrameStoreTests` can only ever test the store in isolation. Every assertion in it uses coordinates in one consistent space, so a green suite says nothing about whether the feature works end to end. That's fine as long as #42 carries the integration fix — just noting that the tests shouldn't be read as coverage of the drag feature.

shanforge (approved):
Reviewed against the running code: checked out the branch, ran the new suite plus the editor suites, ran the CI-pinned linters, and wrote a scratch probe suite to test the hit test at boundaries the PR's own tests avoid. Approving — the diagnosis is right, the fix is the smallest one that addresses it, and the honesty about what it doesn't fix is what makes the rest credible.

**What I verified independently**

- **The mechanism, not just the numbers.** `BlockFrameStore` is a plain `final class` with no `@Observable`, so writing through the reference leaves the `@State` value unchanged and SwiftUI has nothing to invalidate. That's the fix, and it's sound by inspection rather than only by measurement.
- **The premise.** Grepped every reference: `blockFrames` had exactly one reader, `blockAtPoint`, reached only from the drag monitor. Nothing renders from it, so this really was state that never needed observing.
- **Tests.** The 6 new tests pass, and 44 tests across the editor suites (`BlockPaste`, `BlockMove`, `BlockTypeSlashMenu`, `DiagramMathBlock`, `BlockFrameStore`) pass on this branch.
- **SwiftFormat 0.62.1 and SwiftLint 0.65.0 `--strict`** — both clean.
- **Boundary behaviour**, with probes the PR doesn't include. Details inline; short version is that the behaviour is correct and deterministic, but the doc comment's stated reason is wrong.

**What I did not verify**

The per-second figures. Reproducing them needs the instrumentation you built, and you documented the method well enough that I'd rather say plainly that I checked the causal story and not the counts than imply otherwise. The story — preference write → state invalidation → full rebuild → next layout pass → repeat — is consistent with the code, and the fact that body evaluations and preference callbacks came out equal is the kind of detail that's hard to get by accident.

**The judgement I most agree with**

Leaving the coordinate-space bug alone. It would have been tempting to fix the hit test in the same change, and it would have made this diff about two things and the performance claim harder to trust. Preserving behaviour — including behaviour that's broken — is the right shape for a performance PR, and you documented it rather than letting a reviewer discover it.

Same for scoping the remaining per-block `GeometryReader` cost to #42. The measurement you give for it (~40–55 layout passes/sec at 45 blocks vs ~11–20 at 450) is exactly what someone picking that up will need.

Nothing here blocks. The two inline notes are a comment correction and an optional allocation tweak.

Logue/Views/Editor/BlockEditor/Model/BlockFrameStore.swift
westerosweb: You're right that the comment was wrong, and wrong for the reason you give — inclusive bounds at both ends mean touching frames both contain the shared edge, so asserting ties are impossible made the `min` read as dead code. Reworded.

I measured the real geometry before rewording, though, because "they happen at every block boundary" turns out not to hold either. The rows are laid out by `VStack(spacing: 4)`, and the frame-reporting `GeometryReader` sits on the already-padded row, so consecutive frames are separated rather than adjacent. Dumping the live preference values on a real 15-block document:

```
[0] minY=36.0  maxY=102.0
[1] minY=106.0 maxY=164.0
[2] minY=168.0 maxY=214.0
[3] minY=218.0 maxY=377.0
gap[0->1] = 4.0
gap[1->2] = 4.0
gap[2->3] = 4.0
```

Every gap is exactly 4.0pt. So today there are no shared edges — instead there's a 4pt band between each pair of blocks that belongs to no block at all.

I think your probes found ties because they constructed adjacent frames by hand, which is a fair statement about the store's semantics but not about what the editor feeds it.

That also means the downstream inference doesn't hold: this isn't a bigger behavioural improvement than described, because no boundary ties arise under the current spacing. The `min` isn't doing work on every drag — it's doing nothing today.

Which left the comment needing to say two things rather than either one alone, so it now states that the bounds admit ties, that the current 4pt spacing avoids them, and that the tie-break is what stops a zero-spacing layout from silently becoming dictionary-order dependent. And I added tests for both halves so the geometry is asserted instead of described: a point in the gap matches neither block, and a shared edge resolves upward across 50 lookups.

Logue/Views/Editor/BlockEditor/Model/BlockFrameStore.swift
westerosweb: Took it. `frames.lazy.filter { … }.min { … }?.key`, with a short comment saying why the `lazy` is there so it doesn't get tidied away later.

Agreed on the magnitude — once per `leftMouseDragged`, against a dictionary the size of the visible document, so it isn't measurable next to what the PR removes. It's free and strictly better, which is enough reason.

Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift
westerosweb: On the `[blockFrameStore]` capture — that was deliberate for exactly the reason you name, so I've left it. Worth noting it's also what makes the closure safe to keep as-is if `onPreferenceChange`'s action becomes `@Sendable` in a later SDK: it captures the reference, not the view, so there's no `@State` write inside it to become an isolation problem.

No changes from this one.

Logue/Views/Editor/BlockEditor/Views/BlockEditorView.swift
westerosweb: The point about the tests being misreadable as coverage of the drag feature is a real risk, so I've put it in the code rather than leaving it to this thread. `BlockFrameStoreTests` now carries a suite-level doc comment saying it exercises the store in isolation, that every case supplies coordinates in one consistent space, and that a green suite therefore says nothing about whether cross-block drag selection works — with the coordinate-space mismatch named as the reason.

Your reading of the mismatch matches mine: `editorScroll` is origin-top-left with y increasing downward, `contentView` is a different origin and unflipped, so the comparison is between two spaces that agree on neither. Still leaving it for #42.

One thing the measurement in the other thread adds here: the 4pt inter-row gaps mean that even with the coordinate spaces reconciled, roughly 4pt in every ~50pt of vertical travel hit-tests to nothing. Whoever takes #42 will probably want the hit test to resolve a gap to its nearest neighbour rather than returning nil, otherwise a drag will flicker as it crosses each boundary. I've noted that on the issue.

PR #45, opened 2026-07-31
feat: tell people what a release changed

shanforge:
Everything in 1.0.0 and 1.0.1 shipped with nowhere to announce it, so features people
had already downloaded went unfound — and there was no mechanism to do better next
release. This adds one, for the two audiences it has.

## What it does

**A first install** gets a short tour of what Logue does, *after* the onboarding wizard
rather than inside it. The wizard is about permissions and a model; folding a sales pitch
into it makes setup longer at exactly the moment the user wants it shortest.

**Someone upgrading** gets only what changed since the release they last saw — several
versions at once if they skipped some, each card badged with the version that brought it.

The two lists are separate on purpose and neither is derived from the other. A newcomer is
answering "what is this app for"; someone upgrading is answering "what changed". Handing a
newcomer the upgrade deck is a dozen cards of features they have no context for yet, which
is how a tour gets skipped. So the fresh-install list is hand-picked, short, and does not
change every release.

Both are reachable any time from **Help → What's New** and **Settings → General**.

## The case that arrives first

Everyone on 1.0.0 or 1.0.1 today has no stamp, because the stamp did not exist. They are
not new installs, and showing them a tour of features they have used for months would read
as the app having forgotten them. So *onboarded but unstamped* is treated as "has been told
nothing and is owed everything".

That is why **1.1.0's block is the whole feature set** rather than only what 1.1.0 added:
1.1.0 is the release that introduces What's New, so for every existing user this is the
first time any of it is surfaced. It reads in two halves — what Logue already does, then
what is new, because the second only means anything against the first.

**It is a one-off.** Every release after it lists only its own additions, and the card cap
is 6 for a normal release against 13 for 1.1.0, so a repeat fails the build.

## Image sequences

A card can hold several screenshots instead of one. They play in order, looping, with a
step indicator underneath:

```swift
screenshots: ["whatsnew-asklogue-1-where", "whatsnew-asklogue-2-answer"]
```

A still cannot answer *"where do I find this?"*, which is most of what the un-illustrated
cards need to say. Ask Logue uses it for the menu bar item, which is the part nobody finds
on their own. One image still renders as a still, so nothing changes for cards that already
had art.

Reduce Motion advances the steps without the crossfade rather than freezing on the first —
the order is the information. Art is decoded into the card's own state and released with
it, and the well is a fixed height so a change of image does not make the caption jump.

## Sparkle's empty notes pane

Sparkle has offered updates with a blank "what's new" pane since 1.0.0. The appcast
generator has accepted `--notes` the whole time; the workflow just never passed any. Both
that pane and the GitHub Release body now come from the `CHANGELOG.md` section for the
tagged version, so release notes have one home and it is the one already maintained.

Attribution links render as links rather than literal `[@name](https://…)`. Only `http(s)`
targets are linkified, so a changelog cannot put a `javascript:` URL into the update
dialog's web view.

## Adding a release

Appending a block to `WhatsNewCatalog.releases` is the whole job. The catalog is bounded by
the running build, so a block for a version that has not been tagged stays invisible until
it is — blocks can be written as the work lands rather than at tag time.

Full runbook: **[`docs/WHATS_NEW.md`](docs/WHATS_NEW.md)** — which list serves which
audience, what to append at every cut, when to leave the fresh-install tour alone, how to
shoot a sequence, and the checklist for cutting a version.

## Fixed after review

Two review passes over this PR. The one that mattered:

**`markSeen()` stamped the running build version regardless of what was displayed.** The
Help menu and Settings open exactly one release. On a 1.2.0 build a user who opened What's
New by name saw 1.2.0's cards and had 1.1.0 marked seen without ever being shown it — and
the stamp only ever rises, so it was gone for good. The mode now reports what it actually
displayed, the view stamps its own mode, and the defaulted argument is removed so the shape
cannot come back by accident.

Four more from the same pass:

- Help → What's New was unguarded while a sheet was open, replacing a multi-release
  catch-up deck with a single-release one mid-read.
- The tour handoff wrote its sheet after a 600 ms sleep with no re-check, so a sheet the
  user asked for during that window was silently swapped for the tour.
- Reduce Motion was captured once outside `body`, so toggling it mid-sequence did nothing,
  and the step indicators animated regardless.
- The static image cache was never evicted — tens of megabytes of decoded screenshots held
  for the process lifetime, for a sheet opened about once per release.

Escape now works on the last card; it hung off a button that is hidden there.

Earlier in the branch: the deck opened on four symbol-only cards with every screenshot
behind them, which reads as the art being broken; the catch-up block was labelled 1.0.1,
which shipped without any of this; and the tour handoff was keyed to a flag that flips
before the wizard's sheet is dismissed.

The second pass was about cleanliness. Two comments were factually wrong, one AppKit fact
was stated four times, `allScreenshotsResolve` existed only for a test that already computed
it, and `launchReadsStoredStamp` passed with `markSeen` gutted. Net minus 26 lines.

## Verification

- 35 tests across `AppVersionTests` and `WhatsNewGateTests`: gating decisions, version
  ordering, what the Help menu opens on a build older than the catalog, a guard that every
  screenshot the catalog names is in the bundle, sequence well-formedness, and the ones
  that pin the stamping bug — **and they now run in CI**, which they did not when this was
  first opened.
- 24 tests for `scripts/extract_release_notes.py`, which had none, covering the converter
  and both escaping bugs found in review.
- Both sheets rendered through `ImageRenderer` against the real app bundle, and the deck
  driven end to end in a running build, to confirm the art resolves and the sequence plays.
- Release-notes extraction and the appcast `--notes` round-trip exercised against a *copy*
  of `appcast.xml`: the new item carries the description, existing entries untouched.
- SwiftFormat and SwiftLint `--strict` clean on `Logue/` at the versions CI pins. Not on
  `LogueTests/`: CI does not cover it (SwiftFormat runs with a working directory of
  `Logue`, SwiftLint against `Logue/`), and 7 pre-existing test files fail the formatter
  today. The two files this PR touches are clean; the gap is real and left for its own
  change rather than buried in a 7-file reformat here.

## Changed after review

Seventeen review comments, all of them real. The ones that changed behaviour:

- **Settings could permanently retire releases it never showed.** Opening What's New by
  name displayed one release and stamped it, and the stamp only rises — so a user two
  releases behind lost the one in between, without it ever being drawn. Asking by name now
  offers the same backlog a launch would.
- **The release-notes extractor mangled links and code spans.** A query string came out
  double-escaped (`&amp;amp;`), and single-backtick pairing meant the repo's own changelog
  line rendered with a stray delimiter that swallowed the following span. Both are
  regressions with named tests now.
- **A release could ship with blank notes and stay green.** The workflow now stops before
  publishing anything.
- **`AppVersion` rejected semver build metadata**, which made the whole version unreadable
  and showed nothing on every launch of such a build.
- **CI never ran the tests**, so the four catalog rules guarded nothing. It does now — as
  does the extractor's suite.
- Plus: the gate no longer carries a branch nothing could reach, the deck's art is read off
  the main actor, a stray delay joined `AppConstants.Delays`, three tests that could not
  fail were fixed or removed, and four documented numbers now match what shipped.

Two things I did **not** do, both argued on the threads rather than skipped quietly:
extending CI to lint `LogueTests/` (rewrites 7 unrelated files), and making the tour's
stamp testable (needs an injectable version on `Mode` — production shape changed for a
test's benefit).

## Before tagging 1.1.0

`MARKETING_VERSION` is still `1.0.1`, so the 1.1.0 block is clamped out of the build by
design and nothing displays yet. Bumping it is a release action, not part of this PR — step
4 of the checklist in `docs/WHATS_NEW.md`.

## Not covered by tests

The live onboarding → tour sheet handoff, which needs a real window server. Worth
exercising on a real fresh install:

```
defaults delete com.bitwize.logue hasCompletedOnboarding
defaults delete com.bitwize.logue lastSeenWhatsNewVersion
```


Logue/Views/WhatsNew/WhatsNewView.swift
```diff
@@ -0,0 +1,350 @@
...
+
+        /// What "What's New" opens when the user asks for it by name, from the Help menu
+        /// or Settings: the notes for the release this build is. Falls back to the tour
+        /// when this build predates every catalogued release.
+        static var latestNotes: Mode {
+            guard let latest = WhatsNewCatalog.latestRelease(notNewerThan: AppVersion.current) else {
+                return .discover
+            }
+            return .whatsNew([latest])
+        }
```
westerosweb: **This can permanently retire releases that were never shown.**

The `seenThrough` fix below is right as far as it goes — it stamps what was displayed rather than the running build. But `latestNotes` deliberately displays exactly one release, and then stamps it. Because the stamp only ever rises, everything older than that release becomes unreachable.

The Help path is protected by `guard hasCompletedOnboarding, whatsNew == nil` in `LogueApp.swift`. The Settings path is not, and it lives in a different window, so that guard can't cover it.

Concrete, with a future 1.1.0 / 1.2.0 / 1.3.0 catalog — user stamped at 1.1.0 updates to 1.3.0:

1. The launch deck `[1.3.0, 1.2.0]` is up in the main window.
2. They hit ⌘, — a window-modal sheet doesn't block the menu bar.
3. Settings → General → What's New → `latestNotes` shows `[1.3.0]` only.
4. Close it → `onDisappear` stamps 1.3.0. The 1.2.0 cards they never reached are gone for good.

A quieter variant needs no open deck at all: relaunch with the main window closed from the previous session, so `AppRootView` never appears and the launch `.task` never runs — then Settings → What's New is the *first* thing that touches the stamp.

Suggested fix: route the user-initiated entry points through the same backlog the launch gate computes. If `WhatsNewGate.presentationForLaunch()` yields `.whatsNew(releases)`, present those; otherwise fall back to the single latest release. That makes "open it by name" a superset of "shown at launch", and the stamp can never outrun what was rendered.

Logue/Views/Settings/Tabs/GeneralSettingsTab.swift
```diff
@@ -214,6 +221,7 @@ struct GeneralSettingsTab: View {
         .sheet(isPresented: $showWelcomeTour) {
             OnboardingV2View()
         }
+        .sheet(isPresented: $showWhatsNew) { WhatsNewView(mode: .latestNotes) }
```
westerosweb: Two things on this line.

**1. Two boolean `.sheet` modifiers now sit on one view.** The line above is `.sheet(isPresented: $showWelcomeTour)`. `LogueApp.swift:177-179` gives the rationale for the item-driven sheet as *"two boolean `.sheet` modifiers on one view do not reliably both present"* — which is this exact pattern.

If that observation is accurate for this SwiftUI version, the second-attached modifier is the one that silently never fires, and this button either does nothing or shows the welcome tour. Either the comment in `LogueApp` is wrong and should be corrected, or this is a latent dead button. I couldn't resolve it by reading — it needs a click-through of both Settings buttons. If it reproduces, collapse both into one `.sheet(item:)` driven by an enum, matching the `WhatsNewSheetItem` pattern already used in `LogueApp`.

**2. This is the unguarded entry point** for the stamping issue noted on `WhatsNewView.Mode.latestNotes`. Unlike the Help menu item, nothing here checks whether a deck is already open.

.github/workflows/release.yml
```diff
@@ -446,6 +446,39 @@ jobs:
         if: github.event_name != 'workflow_dispatch'
         run: echo "TAG_NAME=${GITHUB_REF#refs/tags/}" >> "$GITHUB_ENV"
 
+      # ── Release notes from CHANGELOG (feeds both the Release body and Sparkle) ──
+      # Written to /tmp because the appcast step below checks out another branch.
+      - name: Extract release notes from CHANGELOG
+        run: |
+          VERSION="${{ steps.version.outputs.version }}"
+          python3 scripts/extract_release_notes.py --version "$VERSION" --format html > /tmp/release_notes.html
+          python3 scripts/extract_release_notes.py --version "$VERSION" --format markdown > /tmp/release_notes.md
```
westerosweb, line 456: **This ships empty release notes silently — the regression the step exists to fix.**

`extract_release_notes.py` exits 0 with only a stderr warning when there's no matching section and no `[Unreleased]` fallback. Then `if [[ -s /tmp/release_notes.md ]]` below just omits the section, and `update_appcast.py` writes no `<description>` when `--notes` is empty. Net result: a release goes out with a blank Sparkle pane and a bare GitHub Release body, and every step is green. A warning on stderr inside a green log is invisible.

The heading regex is stricter than it looks: it requires `^##\s+\[?v?1\.2\.0\]?(\s|$)`. `## v1.2.0 — Big Release` happens to match because the em-dash is `\s`-adjacent, but `## Release 1.2.0` or `## [1.2.0](https://…)` would not.

The script's instinct not to fail a signed, notarised build is correct *for the script*. The decision to abort belongs here, and it can abort before `Create GitHub Release` so nothing half-published exists:

```bash
[[ -s /tmp/release_notes.html ]] || { echo "::error::No release notes extracted for $VERSION"; exit 1; }
```

scripts/extract_release_notes.py
```diff
@@ -0,0 +1,168 @@
...
+    escaped = re.sub(r"\*\*(.+?)\*\*", r"<b>\1</b>", escaped)
+    escaped = re.sub(r"`([^`]+)`", r"<code>\1</code>", escaped)
+    # Attribution links are all over the changelog; without this they reach Sparkle's
+    # pane as literal "[@name](https://…)". Only http(s) targets become links, so a
+    # javascript: URL in a changelog cannot execute in the update dialog's web view.
+    escaped = re.sub(
+        r"\[([^\]]+)\]\((https?://[^)\s]+)\)",
+        lambda m: '<a href="{}">{}</a>'.format(html.escape(m.group(2), quote=True), m.group(1)),
+        escaped,
+    )
```
westerosweb, line 91: **URLs with query strings are double-escaped, producing broken links.**

`html.escape(text, quote=False)` on line 58 has already escaped the whole string, turning `&` into `&amp;`. Re-escaping the captured URL here turns it into `&amp;amp;`. Verified:

```
inline_html('See [docs](https://example.com/a?b=1&c=2)')
→ 'See <a href="https://example.com/a?b=1&amp;amp;c=2">docs</a>'
```

A browser resolves that as `…?b=1&amp;c=2` — the wrong URL. It's latent today because every changelog link is a bare `https://github.com/name`, but the first tracking parameter, GitHub compare link, or `?tab=` ships a dead link into the Sparkle dialog *and* the Release body.

The URL is already escaped by the initial pass, so don't re-escape it — `m.group(2).replace('"', "&quot;")` is enough to keep it safe inside the attribute.

scripts/extract_release_notes.py
```diff
@@ -0,0 +1,168 @@
...
+        line for line in markdown.splitlines()
+        if not re.match(r"^\[[^\]]+\]:\s*\S+", line)
+    )
+
+
+def inline_html(text: str) -> str:
+    """Escape, then re-introduce the few inline marks the changelog actually uses."""
+    escaped = html.escape(text, quote=False)
+    escaped = re.sub(r"\*\*(.+?)\*\*", r"<b>\1</b>", escaped)
+    escaped = re.sub(r"`([^`]+)`", r"<code>\1</code>", escaped)
```
westerosweb: This regex has no notion of fences or doubled delimiters, so an odd number of backticks on a line shifts every pair after it.

It already mangles the repo's own `CHANGELOG.md:56`, which produces:

```
durable in markdown through <code> </code>`<code>mermaid </code> and <code>$$</code> fences.
```

— a stray backtick, and the `$$` span pulled out of position. Given `` Fence: ```mermaid``` and `code`. `` it also corrupts the *following*, well-formed `` `code` `` span.

Fix: match runs of backticks (`` (`+)(.+?)\1 ``), or more simply strip fenced blocks before inline processing.

scripts/extract_release_notes.py
```diff
@@ -0,0 +1,168 @@
...
+        # slip, and its content is what is shipping. Say so loudly, then use it.
+        print(
+            f"warning: no '## [{version}]' section in {path}; falling back to [Unreleased]",
+            file=sys.stderr,
+        )
+        section = find_section(text, re.compile(r"^##\s+\[?Unreleased\]?(\s|$)", re.IGNORECASE))
+
+    if section is None:
+        print(f"warning: no [Unreleased] section in {path} either; emitting nothing", file=sys.stderr)
+        return
```
westerosweb, line 180: Returning quietly here is the right call for the script — a missing changelog section shouldn't fail a signed, notarised build halfway through. But it means the caller has no way to distinguish "no notes" from "notes extracted fine" except by testing the file, which the workflow does only for the markdown copy and not the HTML one. See the note on the workflow step; the assertion belongs there.

Separately: this whole file is 168 lines of regex section-matching and a hand-rolled Markdown→HTML converter feeding both the Sparkle dialog and the Release body, and it has no tests and no CI invocation. Given `to_html` handles bullet continuations, heading levels and inline escaping, a handful of `unittest` cases would be cheap insurance — it's the only untested new logic in the PR, and it already has one live escaping bug.

docs/WHATS_NEW.md
```diff
@@ -0,0 +1,200 @@
...
+
+Features are defined once in the private `Feature` enum at the top of the file and composed
+into both lists, so the copy cannot drift between them. Add new ones there.
+
+**List only what this release added.** A user upgrading from the previous version has
+already seen everything before it, and `WhatsNewGate` will hand a user who skipped three
+releases all three decks at once, newest first, each card badged with the version that
+brought it. Repeating older features means seeing them twice.
+
+Four rules the tests enforce, so a slip fails CI rather than shipping:
```
westerosweb: **CI never runs tests, so none of these four rules fail CI.**

`.github/workflows/ci.yml` is SwiftFormat, SwiftLint, the sandbox verification, and `xcodebuild build`. There is no `xcodebuild test` in `ci.yml`, and none in `release.yml` either. So `everyDeckOpensOnArt`, `catalogScreenshotsResolve`, `releaseNotesAreBounded` and `sequencesAreWellFormed` are guards only if a human remembers to run them — which is precisely the failure mode they were written to prevent.

The better fix is the automation rather than the wording: `WhatsNewGateTests` and `AppVersionTests` need no MLX model, unlike `LLMIntegration/`, so

```
xcodebuild test -only-testing:LogueTests/WhatsNewGateTests -only-testing:LogueTests/AppVersionTests
```

is cheap to add. The whole design of these tests assumes something runs them.

docs/WHATS_NEW.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Why 1.1.0 carries the whole back catalogue
+
+1.1.0 is the release that introduces What's New, so every user arriving from 1.0.0 or 1.0.1
+has been told nothing in-app — everything shipped so far is new *to them*. Its block is
+therefore the full feature set rather than only what 1.1.0 added.
+
+**It is a one-off. Do not copy it as the pattern.** Every release after it lists only its
+own additions, which is a handful of cards. The 12-card cap exists to make an accidental
+repeat fail the build.
```
westerosweb: The cap is **13**, not 12 — line 61 above says "At most 6 cards in a block, and 13 for 1.1.0", and `WhatsNewGateTests.swift:285` uses `cap = release.version == backCatalogue ? 13 : 6`. The 1.1.0 block has exactly 13 features, so it sits precisely on its cap; a reader following the 12 here would think the block is already over budget.

Also: "fail the build" has the same problem as line 53 — the test would fail a local run, not CI.

docs/WHATS_NEW.md
```diff
@@ -0,0 +1,200 @@
...
+matches a file, which is why the test walks the catalog.
+
+### Sequences: showing where a feature lives, then what it does
+
+A card can hold several images instead of one. They play in the order written, looping,
+with a step indicator under the image:
+
+```swift
+screenshots: [
+    "whatsnew-asklogue-1-button",   // where the button is
```
westerosweb: These filenames don't match what shipped. The example names `whatsnew-asklogue-1-button` / `-2-menu` / `-3-answer`; the actual resources are `whatsnew-asklogue-1-where` and `whatsnew-asklogue-2-answer` (two frames, not three). Anyone following this as a template gets names that fail `catalogScreenshotsResolve`.

docs/WHATS_NEW.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Adding a screenshot
+
+1. Put the PNG in `Logue/Resources/`, named `whatsnew-<slug>.png`.
+2. Reference it by base name, no extension: `screenshot: "whatsnew-<slug>"`.
+3. Run `xcodegen generate`. Resources are picked up from the folder, so there is nothing
+   to add by hand.
+
+Resize to **900px wide**. The image well is 250pt tall, so 900px covers Retina with room
+to spare and keeps each file near 200 KB rather than 500 KB. Every screenshot ships inside
```
westerosweb: "near 200 KB" doesn't match the shipped art: `whatsnew-asklogue-2-answer.png` is 695 KB and `whatsnew-asklogue-1-where.png` is 344 KB. Both are correctly 900px wide — PNG content just compresses worse than this guidance assumes. Worth restating as a width rule rather than a size rule, since width is the part that's actually held to.

LogueTests/WhatsNewGateTests.swift
```diff
@@ -0,0 +1,366 @@
...
+            #expect(stored == "1.1.0")
+        }
+    }
+
+    @Test("The stamp is what was shown, not what the build is")
+    func stampRecordsOnlyWhatWasDisplayed() {
+        // The Help menu and Settings open one release, not the whole backlog. Stamping
+        // the running version there marks everything older as seen without showing it,
+        // and because the stamp only rises those releases are gone for good.
+        let catalog = self.catalog
```
westerosweb: **Running `swiftformat` from the repo root rewrites this into code that doesn't compile.**

`.swiftformat` sets `--self remove` with `redundantSelf` enabled, so this becomes `let catalog = catalog` — "variable used within its own initial value". I confirmed it by formatting a copy and diffing.

CI doesn't catch it: `ci.yml` runs `swiftformat --lint .` with `working-directory: Logue` and `swiftlint --strict Logue/`, so `LogueTests/` is unformatted and unlinted. Anyone who runs the formatter from the root breaks the build.

The binding is unnecessary anyway — `withScratchDefaults` takes a non-escaping closure, so the closure below can reference the `catalog` property directly. Deleting this line is the whole fix.

LogueTests/WhatsNewGateTests.swift
```diff
@@ -0,0 +1,366 @@
...
+        )
+        #expect(presentation == .none)
+    }
+
+    // MARK: - Stamp
+
+    /// A scratch defaults suite, so tests never touch the real preferences.
+    private func withScratchDefaults(_ body: (UserDefaults) throws -> Void) rethrows {
+        let name = "WhatsNewGateTests-\(UUID().uuidString)"
+        guard let defaults = UserDefaults(suiteName: name) else { return }
```
westerosweb: If the suite can't be created this returns without running the body, and every test using `withScratchDefaults` passes vacuously — green, having asserted nothing.

`let defaults = try #require(UserDefaults(suiteName: name))` fails loudly instead. Same spirit as the project's "never silent `try?`" rule.

(The isolation itself is the right design — UUID-named suite per call, `defer` cleanup, nothing touching `.standard`. It's only the failure branch that's too quiet.)

LogueTests/WhatsNewGateTests.swift
```diff
@@ -0,0 +1,366 @@
...
+    func stampCoversEveryReleaseShown() {
+        let shown = WhatsNewView.Mode.whatsNew([release(1, 2, 0), release(1, 1, 0)])
+        #expect(shown.seenThrough == version(1, 2, 0))
+    }
+
+    @Test("The tour stamps the running build")
+    func tourStampsCurrentVersion() {
+        // A fresh install has been told what Logue is; it is not owed the catch-up deck.
+        #expect(WhatsNewView.Mode.discover.seenThrough == AppVersion.current)
+    }
```
westerosweb: This is tautological — it asserts `Mode.discover.seenThrough == AppVersion.current` against an implementation that is literally `case .discover: AppVersion.current`. It computes the expected value with the same expression it's testing, so it cannot fail regardless of what `AppVersion.current` returns, including `nil`.

Worth keeping as a pin on the decision, but it verifies no behaviour. Making it meaningful needs an injectable version, the way the gate's `presentation(...)` already takes `current:` — which is what makes the ten gating tests around it genuinely load-bearing.

Logue/Services/WhatsNewGate.swift
```diff
@@ -0,0 +1,85 @@
...
+        releases: [WhatsNewRelease]
+    ) -> Presentation {
+        // Guessing would replay the notes on every launch.
+        guard let current else {
+            logger.error("No parseable app version; skipping What's New")
+            return .none
+        }
+
+        // The caller sequences the tour after the wizard, not this.
+        guard hasCompletedOnboarding else { return .discoverTour }
```
westerosweb: `.discoverTour` is dead in production. The only caller, `LogueApp.swift`, does `guard hasCompletedOnboarding else { return }` *before* calling `presentationForLaunch()`, so this branch is never reached; the tour actually arrives from the onboarding sheet's `onDismiss` via `presentDiscoverTour()`.

Two consequences worth a moment:

- The function reads as the whole decision, but half of it is test-only — and two tests (`freshInstallGetsTour`, `freshInstallIgnoresStamp`) pin a branch nothing in the app reads, while the real sequencing (`onDismiss` → 600 ms handoff → `whatsNew == nil` re-check) has no coverage at all. The coverage is pointed away from the part with the actual timing hazard.
- A future caller who wires this case up would double-present alongside `presentDiscoverTour()`.

Either drop the case or comment it as a contract for a caller that doesn't exist yet.

Separately, on line 35 above: the `guard let current` returns `.none` *before* this onboarding check, so an unparseable `Info.plist` denies a new user the tour too — even though the tour doesn't depend on the version for anything but stamping. Probably not intended, though it barely matters in practice.

Logue/App/AppVersion.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    /// Parses "1", "1.2", "1.2.3" and "1.2.3-beta.1"; absent components are zero, and a
+    /// pre-release suffix orders as the release it leads up to, so a beta tester is not
+    /// shown the same notes again on the final build.
+    ///
+    /// Returns nil for anything else, so a garbled Info.plist reads as "unknown" rather
+    /// than as 0.0.0 — which would compare as older than every real release and replay
+    /// the notes on every launch.
+    init?(_ string: String) {
+        let core = string.split(separator: "-", maxSplits: 1, omittingEmptySubsequences: false).first ?? ""
```
westerosweb: Splitting only on `-` means semver build metadata is rejected: `"1.2.3+456"` leaves component `"3+456"`, which the digits guard below correctly refuses, so the whole version parses to `nil` — and the gate then logs an error and shows nothing, on every launch of that build.

Marketing versions rarely carry `+`, so this is minor, but the pre-release path was handled deliberately and this is its twin. One more split on `+` closes it.

Logue/Views/WhatsNew/WhatsNewView.swift
```diff
@@ -0,0 +1,350 @@
...
+    private struct ShowcaseSequence: View {
+        let urls: [URL]
+        let label: String
+
+        @State private var images: [NSImage] = []
+        @State private var step = 0
+        @Environment(\.accessibilityReduceMotion) private var reduceMotion
+
+        private static let logger = Logger(subsystem: AppConstants.bundleID, category: "WhatsNew")
+        private static let crossfade: TimeInterval = 0.35
```
westerosweb: `CLAUDE.md` requires all delay constants in `AppConstants.Delays` — which this PR correctly did for `sheetHandoff` and `whatsNewSequenceStep`. This one got left behind.

westerosweb (commented):
Reviewed `27066ee..d26896d` on two passes: correctness/behaviour, and standards + test quality. **No Critical findings — this is merge-with-fixes.**

What holds up under tracing, not just reading:

- `AppVersion` parsing is strict in the ways that matter. `"1.10.0" > "1.9.0"`, and `""` / `"abc"` / `"1.x.0"` / `"1.2.3.4"` / `"1..2"` / `"+1.2.3"` / `"-1.2.3"` all return `nil`. Returning `nil` rather than `0.0.0` for garbage is the right call — `0.0.0` would compare older than every release and replay the notes forever.
- The `javascript:`/`data:` rejection genuinely works. I ran it: `[bad](javascript:alert(1))` survives as escaped literal text, and raw `<script>` is escaped by the initial pass.
- The gate's decision function is pure with state injected, which is what makes the 10 gating tests worth anything. The "onboarded but unstamped" bootstrap is correct.
- UserDefaults isolation in the tests is done properly — a UUID-named suite per call with `defer` cleanup, nothing touching `.standard`. Swift Testing runs tests in parallel by default and this is the usual place that bites; it doesn't here.
- Test count is exactly 34 as claimed, and `catalogScreenshotsResolve` is non-vacuous — `TEST_HOST` is set, so `Bundle.main` really is the app bundle.
- SwiftLint 0.65.0 `--strict` is clean across all 11 changed Swift files.
- `MARKETING_VERSION` staying at 1.0.1 is fine: the archive step overrides it from the tag, so the 1.1.0 block is not stranded behind the clamp at release time.

The substantive findings are inline. In priority order:

1. **`Mode.latestNotes` can retire releases that were never rendered** — the one path here that loses data irreversibly.
2. **The release workflow ships empty notes silently** — the exact regression it was written to fix.
3. **Two stacked boolean `.sheet` modifiers** in `GeneralSettingsTab`, which is the pattern `LogueApp.swift:177` comments as unreliable. Needs a click-through to confirm or dismiss.
4. **`docs/WHATS_NEW.md` says the catalog rules fail CI. CI runs no tests at all.**
5. **`swiftformat` from the repo root rewrites a test into code that doesn't compile**, and CI's `Logue/`-scoped lint can't see it.

1, 2 and 4 are small and self-contained; 5 is a one-line deletion. 3 needs two minutes with the app.

Logue/Views/WhatsNew/WhatsNewView.swift
```diff
@@ -0,0 +1,350 @@
...
+            // because the loop below reads it: toggling it mid-sequence must take effect.
+            .task(id: TaskKey(urls: urls, reduceMotion: reduceMotion)) {
+                images = urls.compactMap { url in
+                    guard let image = NSImage(contentsOf: url) else {
+                        // A packaging problem, not a user-facing one — the card still reads.
+                        Self.logger.error("Unreadable What's New art: \(url.lastPathComponent, privacy: .public)")
+                        return nil
+                    }
+                    return image
+                }
```
westerosweb, line 338: `.task` inherits `@MainActor` from the View, so this is synchronous file I/O on the main actor — `NSImage(contentsOf:)` with no suspension point, up to ~700 KB per file, and two files on the Ask Logue card. A brief hitch on every card change.

`await Task.detached { urls.compactMap { NSImage(contentsOf: $0) } }.value` moves the decode off and keeps the assignment on the main actor.

The task hygiene around it is right, for what it's worth — `try? await Task.sleep` followed by `guard !Task.isCancelled` avoids the classic cancelled-sleep busy-loop, and keying on `reduceMotion` gives correct restart-on-change semantics.

Logue/Views/WhatsNew/WhatsNewView.swift
shanforge: Confirmed and fixed in 6b1a15f. The Settings path is unguarded, and both routes you describe reproduce: a window-modal sheet does not block the menu bar, and on a relaunch with the main window closed from the previous session `AppRootView` never appears, so Settings is the first thing to touch the stamp at all.

Taken your suggested shape. `Mode.askedForByName(defaults:)` returns the launch backlog when there is one, and falls back to the single newest release only when there is not — so asking by name is a superset of what launch offers, and the stamp cannot outrun what was rendered.

Two details worth naming:

- It takes `defaults:` rather than reading `.standard`, so the new test drives it against a scratch suite instead of the machine's real preferences.
- The fallback still exists for the case where there is genuinely no backlog — a build predating the catalog, or an unreadable version — where the newest notes beat an empty sheet. That branch was already covered; the new test covers the other one and asserts equality against `presentationForLaunch` rather than restating the expected list, so the two cannot drift apart.

Renamed from `latestNotes`, because the name was the bug in miniature: it described showing the latest, which is exactly what it did.

Logue/Views/Settings/Tabs/GeneralSettingsTab.swift
shanforge: Both parts taken, in 6b1a15f.

**1.** I could not resolve the presentation question by reading either, and I did not want to leave a button whose behaviour depends on which answer is right — so the pair is gone rather than tested. Both are now one `.sheet(item:)` driven by a `Sheet` enum, matching `WhatsNewSheetItem` in `LogueApp`. One modifier cannot have the problem, so whether the observation in `LogueApp` is accurate for this SwiftUI version stops mattering here. If it turns out to be wrong, that comment is the thing to correct, and this tab is no worse for the change.

**2.** Fixed as described on the other thread — this entry point now opens the same backlog a launch would. The mode is resolved when the button is pressed rather than when the tab is built, because the stamp may have moved since the launch deck was dismissed.

.github/workflows/release.yml
shanforge: Agreed, and taken verbatim in bd4e2a7 — the abort belongs to the workflow, not the script, and placing it before `Create GitHub Release` means nothing half-published exists when it fires.

Both files are checked, not just the HTML one: the markdown copy feeds the Release body and the HTML copy feeds Sparkle, and either being empty is the same silent failure.

```bash
if [[ ! -s /tmp/release_notes.html || ! -s /tmp/release_notes.md ]]; then
  echo "::error::No release notes extracted for ${VERSION}. Add a '## [${VERSION}]' section to CHANGELOG.md (or move [Unreleased] into one) and re-run."
  exit 1
fi
```

The now-unreachable `if [[ -s /tmp/release_notes.md ]]` below it is gone rather than left as a condition that can no longer be false.

On the heading regex being stricter than it looks: you are right that `## Release 1.2.0` and `## [1.2.0](https://…)` both miss. I have left it strict rather than widening it to shapes this changelog does not use — a looser pattern has its own failure mode, matching the wrong section silently. What changes is the consequence: a miss now stops the release with a message naming the version and the file, instead of shipping blank notes. The tests cover the forms that are supported (`## [1.1.0] - date`, `## 2.0.0`, `## v2.0.0 — Big`, `## [Unreleased]`) so the supported set is at least written down.

scripts/extract_release_notes.py
shanforge: Reproduced exactly as you wrote it, and fixed in 7276f3b with your suggested fix — the initial pass already escaped the URL, so the only escaping left to do is the one that closes the attribute:

```python
lambda m: '<a href="{}">{}</a>'.format(m.group(2).replace('"', "&quot;"), m.group(1))
```

Before / after on your case:

```
'See [docs](https://example.com/a?b=1&c=2)'
  before -> See <a href="https://example.com/a?b=1&amp;amp;c=2">docs</a>
  after  -> See <a href="https://example.com/a?b=1&amp;c=2">docs</a>
```

Both are pinned now: `test_query_string_is_escaped_once`, and `test_quotes_in_url_cannot_close_the_attribute` for the case the `replace` is there to handle.

scripts/extract_release_notes.py
shanforge: Reproduced, including the part that is easy to miss — that it also corrupts the *following*, well-formed span. Fixed in 7276f3b.

Worth reporting that your first suggestion, matching runs with `` (`+)(.+?)\1 ``, is not quite enough on this changelog. The real line is a **1**-backtick span whose content contains a 3-run:

```
  ` ```mermaid ` and `$$` fences. (#7)
```

A bare backreference lets the opening 1-run close on the *first* backtick of the ``` run, so it still mispairs:

```
before      -> durable in markdown through ``<code>mermaid</code>`<code> and </code>$$` fences.
(`+)(.+?)\1 -> durable in markdown through <code> </code><code>`mermaid </code> and <code>$$</code> fences.
```

What CommonMark actually requires is that the closing run be the same length *and not part of a longer run*, which needs the lookarounds:

```python
CODE_SPAN = re.compile(r"(?<!`)(`+)(?!`)(.+?)(?<!`)\1(?!`)")
```

```
after -> durable in markdown through <code>```mermaid</code> and <code>$$</code> fences.
```

Also strips the one padding space each side, which is the mechanism that lets a span hold a backtick run at all — otherwise it renders as `` ```mermaid `` with stray spaces, which is not what the changelog shows on GitHub.

I went with the regex rather than stripping fenced blocks first because there are no multi-line fences in a changelog section; a fence here is always inline, and the converter treats a stray one as a paragraph. Pinned by `test_backtick_run_closes_on_equal_run`, `test_span_may_contain_a_longer_run` (the repo's own line), `test_unclosed_backtick_is_left_alone` and `test_padding_spaces_are_stripped_only_when_symmetric`.

scripts/extract_release_notes.py
shanforge: Agreed on both, and both done.

The quiet return stays — for the reason you give, it is right for the script — and the assertion went into the workflow (bd4e2a7), covering the markdown copy as well as the HTML one, since either being empty is the same silent failure.

`scripts/test_extract_release_notes.py` covers `inline_html`, `to_html`, `find_section` and `strip_link_definitions` in 24 cases, including bullet continuations, heading levels, the list-closing rules, and the two live bugs as named regressions. It runs in the Lint Check job of `ci.yml` (Python and nothing else), so it is enforced rather than available.

You were right that it was the only untested new logic here and that it already had a live bug — writing the tests turned up a second one you had not flagged: the run-matching fix on the other thread is insufficient for the changelog's own line, which the tests now pin.

docs/WHATS_NEW.md
shanforge: You are right, and the automation is what changed — the wording would have been papering over it.

`ci.yml`'s Build job now runs:

```
xcodebuild test ... \
  -only-testing:LogueTests/WhatsNewGateTests \
  -only-testing:LogueTests/AppVersionTests
```

Named suites rather than the whole target, for the reason you identify: `LLMIntegration` runs real inference against a local MLX model, which no hosted runner has. The comment above the step says to add a suite there when one arrives that needs no model, so the list does not quietly go stale.

`scripts/test_extract_release_notes.py` runs in the Lint Check job for the same reason — it needs Python and nothing else.

The doc line now names the job and the constraint instead of asserting the outcome, and the release checklist gives the same command CI uses so the two cannot drift.

docs/WHATS_NEW.md
shanforge: Right on both counts — 13, and 1.1.0 has exactly 13, so it sits *on* its cap. Corrected in 95af4d3, and the doc now says so explicitly, along with what to do when the block needs another card (move it to the release that introduced the feature, rather than raising the cap).

"Fail the build" is now accurate rather than aspirational: both suites run in CI as of bd4e2a7.

docs/WHATS_NEW.md
shanforge: Correct — the shipped resources are `whatsnew-asklogue-1-where` and `whatsnew-asklogue-2-answer`, two frames rather than three. The example named files that do not exist, so anyone using it as a template got names that fail `catalogScreenshotsResolve` — which is exactly the test the doc is teaching them about. Fixed in 95af4d3 to match what shipped.

docs/WHATS_NEW.md
shanforge: Confirmed against the shipped files:

```
695875  whatsnew-asklogue-2-answer.png
344553  whatsnew-asklogue-1-where.png
292606  whatsnew-transcription.png
191254  whatsnew-spaces.png
```

All 900px wide, spanning 190 KB to 700 KB — so the number was never the rule, as you say; width is. Restated that way in 95af4d3, with the observed range named so the next person is not surprised by a 700 KB file that is correctly sized, and a note to prefer a tighter crop since this is download size for every user.

LogueTests/WhatsNewGateTests.swift
shanforge: Confirmed and deleted in cd93305 — the binding was pointless, as you say, since `withScratchDefaults` takes a non-escaping closure.

I also confirmed the CI gap you describe: `swiftformat --lint .` runs with `working-directory: Logue` and SwiftLint against `Logue/`, so `LogueTests/` is neither formatted nor linted, and formatting from the repo root produces code that does not compile.

I have **not** extended CI to cover `LogueTests/` in this PR, deliberately. Running the formatter over it today rewrites 7 files — `LLMTestHarness`, `VerifyPanelLLMTests`, `VocabularyLLMTests`, `DeepLinkRoutingTests`, `DroppedFileImportTests`, `EditorZoomTests`, `ImportFileReadingTests` — none of which this PR touches, and that diff would bury the review. The two files here (`WhatsNewGateTests`, `AppVersionTests`) are both clean under the pinned formatter, so the specific hazard is gone; the general one is worth its own change and I am happy to open it.

LogueTests/WhatsNewGateTests.swift
shanforge: Agreed — fixed in cd93305 with `try #require`, and the four callers plus the helper are now throwing. Same spirit as the project's rule against silent `try?`, as you say: a helper that skips its own body is worse than one that throws, because the suite still reports green.

Thanks for the note on the isolation itself — kept exactly as it was.

LogueTests/WhatsNewGateTests.swift
shanforge: Agreed that it verifies nothing. I removed it rather than kept it (cd93305).

Making it meaningful does need an injectable version, and I did not think that was worth changing `Mode`'s shape in production for — `seenThrough` is read by the view on dismiss, and threading a version through it for a test's benefit trades a real seam for a synthetic one. So the decision is recorded where it is made, on the property, and the file carries a short note saying why the `.discover` branch has no test. The `.whatsNew` branch, which is the half that can be got wrong, keeps both its cases.

Open to the injectable version if you would rather have the coverage than the shape; it is a small change and I would rather you pick.

Logue/Services/WhatsNewGate.swift
shanforge: Confirmed dead, and dropped in aa87380 rather than commented — a case that only tests read is a claim about the app that is not true, and the double-present hazard you name is a real cost of leaving it.

A fresh install is now simply owed nothing at launch, which is what the app already believed. `freshInstallGetsTour` / `freshInstallIgnoresStamp` became assertions that the launch gate stays out of the way, with the reason on them; the tour's real source is `onDismiss` → `presentDiscoverTour()`.

Your second point resolves itself as a side effect: with the tour gone from the gate, the `guard let current` ordering no longer denies anything to a new user, and the onboarding check now comes first regardless.

The coverage gap you identify — the `onDismiss` → handoff → `whatsNew == nil` re-check sequence having no tests, which is where the actual timing hazard lives — is real and still open. It needs the sheet lifecycle rather than a pure function, so it is not reachable from this suite; I have left it rather than pretend otherwise.

Logue/App/AppVersion.swift
shanforge: Fixed in dc96501. Minor as you say, but the failure mode is not small — one rejected component fails the whole parse, which reads as "unknown", and the gate then shows nothing on *every* launch of that build.

One wrinkle worth naming: metadata has to come off before the pre-release suffix, not after. Semver orders them `1.2.3-beta.1+456`, so splitting on `-` first leaves `+456` attached to `beta` and the metadata survives into the core. Both forms are covered.

Logue/Views/WhatsNew/WhatsNewView.swift
shanforge: Correct — fixed in 8b60364. It is now `AppConstants.Delays.whatsNewSequenceCrossfade`, alongside the two this PR did get right. A `TimeInterval` rather than a `Duration`, since it feeds `.easeInOut(duration:)`, which does not take one; noted on the constant so the type difference does not read as an oversight.

Logue/Views/WhatsNew/WhatsNewView.swift
shanforge: Fixed in 8b60364, with one deviation from your suggestion worth flagging.

`Task.detached { urls.compactMap { NSImage(contentsOf: $0) } }.value` sends `[NSImage]` across an actor boundary, and `NSImage` is not `Sendable` — so it either warns now or breaks under stricter checking later. The bytes go off the actor instead and the images are built back on it:

```swift
let loaded = await Task.detached(priority: .userInitiated) {
    urls.map { url -> (URL, Data?) in (url, try? Data(contentsOf: url)) }
}.value
images = loaded.compactMap { url, data in ... }
```

That moves the part your comment is actually about — synchronous file I/O of several hundred KB per frame on the main actor — without an unchecked send. `NSImage(data:)` defers the bitmap decode to draw time regardless, so nothing is gained by forcing it across.

The URL is carried through the tuple so the "unreadable art" log still names the file that failed.

And thanks for the note on the task hygiene — that was deliberate, and it is useful to know it reads that way.

shanforge:
### Note on the runner pin (`312fa08`)

Adding `xcodebuild test` to CI surfaced something the build alone never could: the
Build job ran on `macos-15`, and a macOS 15 runner cannot *run* a test target whose
deployment target is macOS 26.

```
xcodebuild: error: Cannot test target "LogueTests" on "My Mac":
My Mac's macOS 15.7.7 doesn't match LogueTests's macOS 26.0 deployment target.
```

Cross-compiling is fine, which is why `xcodebuild build` was green on that image for
as long as it has been — the mismatch only becomes visible the moment anything tries
to execute a test.

The fix is #53, which pins all three `runs-on` values to `macos-26`, so that commit
is cherry-picked onto this branch. It is the same commit, so once #53 merges a rebase
here drops the duplicate by patch-id rather than conflicting.

**Merge #53 first.** If this lands first, #53 becomes empty.

westerosweb:
@shanforge Pls resolve the conflicts

shanforge:
### Conflicts resolved — and the runner-pin note above is now spent

#53 has merged, so the commit this branch was carrying is on `main` from both sides. As
expected, the workflows merged cleanly (identical content); the only conflict was
`Logue.xcodeproj/project.pbxproj`, which is generated — resolved by running
`xcodegen generate` over the merged `project.yml` rather than picking hunks out of a file
nobody edits by hand.

`origin/main` is now an ancestor of this branch and GitHub reports it mergeable again.

Verified on the merged tree, not just before it:

- `xcodebuild build` succeeds
- 35 tests across `WhatsNewGateTests` and `AppVersionTests` pass
- 24 tests for `scripts/extract_release_notes.py` pass
- SwiftFormat and SwiftLint `--strict` clean on `Logue/`
- Both runner pins in `ci.yml` and the one in `release.yml` still read `macos-26`, and the
  `-only-testing` step survived the merge — worth checking explicitly, since those are the
  two halves that had to meet here.

The PR body is updated: the test count was stale, and it claimed SwiftFormat/SwiftLint
clean on `LogueTests/`, which is not true and not something CI checks. Corrected rather
than left standing.

PR #48, opened 2026-08-01
Stop long meetings losing their transcript at the memory limit

westerosweb:
Fixes the report in discussion #47.

## What a user saw

From #47:

> I've tried recording/transcribing a few meetings, and all the transcriptions seem to stop right at 30 minutes — is this a known limit, configurable, or a function of my system memory somehow?

A hard limit, and not a memory one in the way it looked.

## What was happening

When a recording stopped, the live transcript was replaced by a batch re-transcription (Parakeet TDT, more accurate than the live stream) of the audio accumulated in memory. That buffer was capped at 1800 seconds, and the replacement covered the **whole session** regardless of how much of it the batch pass had actually heard.

So a ninety-minute meeting had its last hour deleted at the moment recording stopped — text that had been on screen the entire time. Memory looked fine because the cap held the buffer at ~115 MB.

## The shape of the fix

Two rules, and most of this PR is making them true everywhere.

**A pass may only speak for the audio it actually heard.** `TranscriptReplacement` bounds the replacement to the stretch of timeline the batch pass covered; whatever the live transcriber produced past that point stays. Nothing is deleted for being past a limit.

**The saved audio file is the meeting's timeline.** A capture source writes only while it is running, so its raw file is shorter than the meeting whenever it joined late or was muted. `CaptureSegmentTimeline` records where each activation belongs — for the microphone and the system tap alike — and the file is composed from those placements whenever the raw one would not already line up. Playback depends on this, and so does the long-recording pass below.

## Removing the length limit

The cap was raised to a memory-derived figure and then removed as a limit altogether.

Neither model needs the recording in one piece. `AsrManager.transcribe(url:)` switches itself to disk-backed chunking above its streaming threshold; Sortformer's whole-buffer call internally walks the audio through the same streaming state machine as its incremental API, carrying speaker identity in `spkcache`/`fifo` between chunks rather than re-deriving it. So when a recording outgrows memory it is read back from its own file 30 seconds at a time and processed to the end. **There is no longer a length at which the transcript or the speaker labels stop.**

The in-memory ceiling now only chooses the route. It is derived from physical memory rather than being a fixed number of minutes — roughly 2.3 h on 8 GB, 4.7 h from 16 GB up.

Reading a file back means trusting its timings, so the route is gated on that trust rather than on a guess. `persistRecordingAudio` reports what it laid down — `onSessionTimeline`, `rawFallback`, or `none` — and the disk route runs only on the first. A composition that failed, a source muted for half the meeting, a stale file: each degrades to keeping the live transcript instead of stamping a plausible transcript at the wrong times over one that had them right.

Where the file cannot be used, each model's result is still kept on its own. A transcript with no speaker labels beats neither.

## Other defects fixed along the way

- **Two capture sources shared one buffer in arrival order.** `AudioTimelineMixer` gives each source its own cursor and sums overlaps, so an hour of meeting is an hour of buffer whether one source is running or two. Previously, enabling the mic during an online meeting filled the buffer at twice real time and drifted every timestamp derived from it, progressively.
- **`elapsedTime` was read from a capture device's clock**, which restarts at zero every time that source is toggled — `SystemAudioCapture` zeroes its own outright. A long meeting with a mid-session mute was saved with the wrong duration. It now measures from the session's own start.
- **One shared `AVAudioConverter` served both sources**, torn down and rebuilt on every buffer as their formats alternated. One per source now.
- **Toggling the mic during an in-person recording left the diarizer with no audio** for the rest of the session, so speaker detection quietly stopped.
- **The playback mix laid the mic file down at zero** however late the mic joined, and across mutes the file is a concatenation — so everything after the first mute played early, by more with each toggle.
- **`AudioRecorder.recordedDuration` read `AVAudioFile.length` while the tap thread wrote to it.** Seconds are counted on the tap thread under a lock, which also survives the input format changing mid-recording.
- **The full audio buffer stayed live across the disk pass**, pinning up to a gigabyte while two CoreML models ran. It is discarded up front on that route.
- **The disk route skipped VAD**, leaving the longest recordings — the ones most likely to hold long silences — as the only ones getting unfiltered output. It now runs per chunk, off the MainActor.

## Verification

| | |
|---|---|
| Full suite | **935 tests in 82 suites pass** |
| New unit tests | 40, across timeline mixing, transcript replacement, capture placement, chunked reading, and the timeline-trust decision |
| 5-hour end-to-end, real models | transcript to 99.9%, speaker labels to 100.0%, 2 speakers held throughout, 6× realtime |
| 6-hour read-back | full coverage, offsets strictly increasing, **≤4 MB resident** |

The 5-hour fixture is two alternating synthesised speakers at 5.01 h against a 4.66 h in-memory ceiling, so it genuinely takes the disk route:

```
Fixture 5.01 h; in-memory timeline holds 4.66 h — the recording does not fit
Processed 5.01 h in 3122 s (6x realtime)
  transcript: 629 segments, last ends at 18022 s (99.9% of the recording)
  speakers:   2 distinct, 1401 segments, last ends at 18032 s (100.0% of the recording)
```

Process RSS peaked around 800 MB early (the memory-mapped source) and fell to ~360 MB for the rest of the run, rather than growing with the recording. Every tenth of the recording has transcript in it, so coverage is not front-loaded.

The two multi-gigabyte fixtures are opt-in (`LOGUE_LONG_AUDIO`, `LOGUE_RUN_MULTIHOUR_TESTS`) so the normal run stays fast and offline.

Build clean; SwiftFormat and SwiftLint `--strict` clean. `RecordingSessionManager` crossed the 800-line limit, so its audio-file operations moved to `RecordingSessionManager+AudioFiles.swift`, and `AudioRecorder`'s two hand-maintained copies of the tap closure became one.

## How long a meeting can run now

- **Live transcript, and the final saved transcript: no limit.**
- **Batch-quality transcript and speaker labels: no limit** — beyond what memory holds it switches to reading the recording from disk.
- That is wall-clock meeting time whether or not the mic is on. Before the mixing fix, enabling the mic in an online meeting halved it.

Chunk-fed diarization is not identical to the whole-buffer call: that one accumulates every chunk's predictions and rebuilds the timeline from all of them at the end, where this commits each chunk as it goes, so expect it to be slightly rougher at the seams. It is the route for recordings the other cannot process at all.

## Not covered here

Package upgrades (FluidAudio 0.13.6 → 0.15.5, Textual → 0.5.0, Sparkle → 2.9.4) and replacing hand-rolled plumbing with FluidAudio equivalents are deliberately left for a separate branch.


Logue/Engine/CaptureSegmentTimeline.swift
```diff
@@ -0,0 +1,61 @@
...
+        copy.sourceStopped(fileDuration: fileDuration)
+        return copy.placements
+    }
+
+    /// Whether the file can stand in for the session timeline as it is.
+    ///
+    /// True only when the source ran once, from the start of the meeting: then the file *is* the
+    /// timeline and can be saved untouched. Anything else — a late join, a mute — means the file is
+    /// shorter than the meeting and has to be laid out before it represents it.
+    static func fileMatchesSessionTimeline(_ placements: [Placement]) -> Bool {
```
shanforge, line 57: **Blocking — this proves the file is *aligned*, not that it *covers* the session, and the disk route treats it as both.**

`fileMatchesSessionTimeline` returns true for any single placement starting at zero. Nothing compares `only.duration` against how long the session actually ran. `persistRecordingAudio` then hands the raw file back as `.onSessionTimeline`, and `RecordingSessionManager+Diarization.swift:176` returns `heardDuration: result.segments.isEmpty ? 0 : nil` — `nil` meaning "heard the whole session", derived only from whether the batch produced text, never from the file's length.

Reachable via mic write failures, which the tap deliberately tolerates (`AudioRecorder.swift:149-151`, "non-fatal, and recording continues"):

1. In-person, 5 h recording. Disk fills at t = 4 h.
2. `try capturedFile?.write` starts throwing, so `writtenSeconds` stops advancing — but `callbackHolder.callback?(buffer)` still fires, so the live transcriber and the mixer keep receiving audio to the end.
3. At stop: `micPlacements == [(fileStart: 0, duration: 4 h, sessionStart: 0)]` — one placement from zero, so `needsComposing == false` and the file is trusted as `.onSessionTimeline`.
4. The mixer overran at ~4.6 h, so `heardDuration != nil` and the disk route is taken — on a 4-hour file.
5. The pass succeeds and reports `heardDuration: nil`, so `TranscriptReplacement.merged` keeps only `existing.filter { $0.startTime < sessionStart }` and deletes **every live segment from 4 h to 5 h**.

That is the same shape as the bug this PR fixes — an hour of on-screen transcript deleted at the moment recording stops — arriving through the door the trust gate left open. The compose route has the same hole (a composed file can be on the timeline yet silent for an hour and still read as full coverage).

`AudioFileChunkReader.duration(of:)` already exists and is currently referenced only from tests. Comparing it against the session duration in `runBatchPass`, and passing a real `heardDuration` when the file falls short, would close this with the tool already in the PR.

`SavedRecordingTrustTests.swift:26` encodes the gap: `unbroken(900)` asserts `!needsComposing` with no notion of how long the session ran.

Logue/Engine/DiarizationManager+LongRecording.swift
```diff
@@ -0,0 +1,204 @@
...
+
+        guard let result else { return nil }
+        let trimmed = result.text.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else {
+            logger.info("Long-recording ASR returned an empty transcript")
+            return nil
+        }
+        guard let timings = result.tokenTimings, !timings.isEmpty else {
+            logger.warning("Long-recording ASR: no token timings — returning one segment")
+            return [TranscriptSegment(text: trimmed, startTime: 0, endTime: result.duration)]
```
shanforge: **Blocking — `result.duration` is always `0` on this route, so this fallback can replace an entire meeting with one zero-length segment.**

Verified in the pinned FluidAudio source rather than inferred:

- `AsrManager.transcribe(_ url:)` routes to `transcribeDiskBacked` above `streamingThreshold` (`AsrManager.swift:348-350`) — which is the whole point of using it here.
- `ChunkProcessor.process` finishes with `processTranscriptionResult(..., audioSamples: [], ...)` (`ChunkProcessor.swift:154-162`).
- `processTranscriptionResult` computes `let duration = TimeInterval(audioSamples.count) / TimeInterval(config.sampleRate)` (`AsrManager+Transcription.swift:109`) → **0**.

The in-memory path passes real samples, so `result.duration` is meaningful there; on this new path it never is.

If this guard fires, the result is a single segment holding a multi-hour transcript with `startTime: 0, endTime: 0`. Because `segments` is then non-empty, `runBatchPass` reports `heardDuration: nil`, and `TranscriptReplacement` discards the entire live transcript in favour of that blob.

Reachability is low — `createTokenTimings` returns `[]` only on a token/timestamp count mismatch — but the value is provably wrong on this route and the blast radius is the whole meeting. Either derive the end from the file (`AudioFileChunkReader.duration(of:)`) or return `nil` here and keep the live transcript.

Logue/Services/RecordingSessionManager+AudioFiles.swift
```diff
@@ -0,0 +1,241 @@
...
+    /// which is only safe when the file is the meeting's timeline. Working that out after the fact —
+    /// from a device clock, or from the file's length — gets it wrong precisely in the cases that
+    /// produce a misaligned file, because those are the same cases that disturb the clock.
+    enum SavedRecording {
+        /// The file is the meeting's timeline. Timings taken from it describe the meeting.
+        case onSessionTimeline(URL)
+        /// Audio was saved, but laying it out failed, so its timings may not line up.
+        case rawFallback(URL)
+        /// Nothing was saved.
+        case none
```
shanforge, line 58: **Blocking — this turns every single-source recording with one mute into a full AAC re-encode, where main did an instant file move.**

`fileMatchesSessionTimeline` is true only for a single placement from zero, so one mic toggle in an in-person recording produces two placements → `needsComposing` → `composeRecording` → `AVAssetExportSession`.

On main the export ran **only** when both sources existed (`RecordingSessionManager.swift:401` on main: `if let sysURL = systemTempURL, let mURL = audioRecorder.tempFileURL`). A mic-only recording took the `moveRecordingFile` branch — a rename, effectively instant.

So a 3-hour in-person recording where the user muted once now re-encodes 3 hours of WAV to AAC inside `await persistRecordingAudio(...)` (`RecordingSessionManager.swift:458`), which sits *before* `updateDuration` (:460) and before the post-recording task is launched (:479). For that whole window — a minute or more — `recordingState` stays `.stopping`, and `startRecording()` guards on `.idle`, so the user cannot start another recording and the meeting shows duration 0 in the list.

Secondary effect worth weighing: the long-recording pass then reads a lossy AAC file back for ASR rather than the original PCM.

A single-source recording whose placements are contiguous in the file needs no re-encode — only a placement offset — so the common mute case could stay a move.

Logue/Engine/AudioTimelineMixer.swift
```diff
@@ -0,0 +1,139 @@
...
+        }
+        if end > capacity {
+            didDropAudio = true
+        }
+
+        let writable = min(end, capacity)
+        if samples.isEmpty {
+            samples.reserveCapacity(min(capacity, Int(sampleRate * 600)))
+        }
+        if samples.count < writable {
```
shanforge, line 141: **Silence between activations is materialised as real samples, on the main actor, and charged against the capacity budget.**

`DiarizationManager` is `@MainActor`, so `processAudioBuffer` → `accumulateBatchBuffer` → `write` puts this append on the main actor.

In an in-person recording the mic is the only source, so while it is muted *nothing* writes. On unmute, `enableMic` calls `beginSource(.microphone, atSessionTime: micJoinedAt)`, and the next write allocates and zeroes the entire gap in one go:

| break | zeros | memory |
|---|---|---|
| 20 min | 19.2 M | ~77 MB |
| 1 h | 57.6 M | ~230 MB |
| 2 h | 115 M | ~460 MB |

Three consequences: a main-actor hitch proportional to the break; on an 8 GB Mac (134 M sample budget) a 2-hour break consumes ~90% of the capacity meant for audio, so real post-break audio starts being dropped after ~50 s; and both models are then made to chew through hours of digital silence.

On main the buffer only ever held real audio, so none of this existed — it is the cost of the (correct) move to a time-addressed timeline. Some representation of the gap is genuinely needed for alignment, so this is a design question rather than an outright bug, but silence probably should not be charged against the capacity budget, and a large gap probably should not be memset on the main actor.

Logue/Engine/AudioFileChunkReader.swift
```diff
@@ -0,0 +1,133 @@
...
+            let offset = producedSamples
+            producedSamples += samples.count
+            return (samples, offset)
+        }
+
+        // Drain whatever the converter is still holding. It buffers a little to do its work, so
+        // without this the last few milliseconds of every recording are quietly dropped.
+        guard !drained else { return nil }
+        drained = true
+        guard let tail = try? convert(feeding: nil), !tail.isEmpty else { return nil }
```
shanforge: **Silent `try?` — and it is swallowing the error in the one place whose whole purpose is not to lose audio silently.**

The comment two lines up says the drain exists because "without this the last few milliseconds of every recording are quietly dropped". If the flush throws, they are dropped exactly as before, with nothing logged.

CLAUDE.md: "Never use silent `try?` for ANY operation" — the only exemption is `try? await Task.sleep`. `next()` already `throws` and every other error path in this file propagates or throws a typed error, so `try` alone would do it.

Logue/App/AppConstants.swift
```diff
@@ -137,6 +137,20 @@ enum AppConstants {
...
         /// Post-recording pipeline timeout in seconds
         static let postRecordingTimeoutSeconds: TimeInterval = 180
+        /// How much of a recording the post-recording batch pass (Parakeet ASR + Sortformer) can be
+        /// handed in one go, expressed as memory rather than as minutes: the session's audio is held
+        /// as 16 kHz mono Float32 (~230 MB per hour) and given to the models whole, so what bounds
+        /// it is what the machine can hold. A fixed number of minutes would be wasteful on a large
+        /// Mac and reckless on a small one. `AudioTimelineMixer.capacity(forPhysicalMemory:)` reads
+        /// these; audio past the limit is dropped and that stretch keeps its live transcript rather
+        /// than losing it — see `TranscriptReplacement`.
+        static let audioBufferMemoryDivisor: UInt64 = 16 // a sixteenth of physical memory
```
shanforge: **Sized from total RAM, not free RAM.**

The derivation is sound — floor-then-cap ordering is right, `UInt64` division cannot trap, and it produces the documented 1.16 h / 2.33 h / 4.66 h. My concern is the input: `ProcessInfo.processInfo.physicalMemory` (`DiarizationManager.swift:112`) is *installed* memory.

On the smallest supported Mac (8 GB) this is `8 GiB / 16 = 512 MB` resident Float32, up from the old fixed ~115 MB, and that buffer is handed to Parakeet and Sortformer **simultaneously** (`RecordingSessionManager+Diarization.swift:196-198`) while an MLX LLM may already be resident.

CLAUDE.md: "Check available memory before loading models. Use `mach_task_basic_info` to verify at least 512MB available." Nothing on this path consults available memory.

shanforge (changes requested):
Reviewed `21fa2ea..1faa6ec` across migration/compatibility, correctness/concurrency, and cleanliness/tests. Verdict: **request changes** — the engineering here is strong and the central fix is right, but three paths reintroduce the PR's own bug class or make a recording unstoppable, and one is a plain regression for the most common in-person recording.

## What holds up

I went looking for a transcript-wipe path and could not find one in the parts this PR set out to fix. `filterWithVAD` and `filterWithFileVAD` both refuse to over-filter to empty; `transcribeBuffer` and `transcribeRecordingFile` both guard empty; the disk route maps an empty result to `nil` with `heardDuration: 0`, which `TranscriptReplacement` treats as "keep everything". That is three independent guards on the same failure, and they hold.

Also verified rather than taken on trust:

- **Persisted-data compatibility is clean.** The only file touched under `Models/` is `TranscriptSegment.swift`, and the change is a doc comment. No field added, removed or retyped; `audioSource` stays optional. `CaptureSegmentTimeline`, `SavedRecording` and `AudioTimelineMixer` are per-session and never persisted, so there is no "missing timeline metadata" case for a 1.0.x meeting. Old recordings are never read under the new assumptions — `processRecordingFile` has exactly one caller, inside `stopRecording`.
- **The trust gate cannot be reached with `rawFallback` or `none`.** `describesSessionTimeline` matches only `.onSessionTimeline`, plus a re-check of the URL and file existence.
- **The timeline arithmetic is sound.** Sample-counted cursors rather than accumulated floats, so no drift over multi-hour runs; `Double`'s mantissa is exact well past 288 M samples; mixing is hard-clamped to ±1; `write` bounds hold on both arrays. `capacity(forPhysicalMemory:)` matches its documented figures (8 GB → 2.33 h, 16 GB+ → 4.66 h).
- **`take()` does not double peak memory**, and the `discardAudioBuffer()` placement genuinely frees the buffer before the models run — nothing else retains it.
- **935 tests in 82 suites pass**, exactly as claimed — I built the branch and ran them. SwiftFormat 0/441 and SwiftLint `--strict` clean against the pinned CI versions.

The four new pure types are cohesive rather than ceremonial: each encodes a rule that was previously implicit and wrong.

## Blocking

1. **`.onSessionTimeline` proves alignment, not coverage** — inline on `CaptureSegmentTimeline.swift`. A file that is on the timeline but *short* is trusted as the whole session, and the disk route then reports `heardDuration: nil`. That drops every live segment past the file's end: the exact loss this PR exists to prevent.
2. **`result.duration` is always `0` on the disk-backed ASR route** — inline on `DiarizationManager+LongRecording.swift`. Verified in FluidAudio's source, not inferred.
3. **The disk pass has no cancellation point** — see below; the app can be left stuck in `.starting`.
4. **Every single-source recording with one mute now does a full AAC re-encode** — inline on `RecordingSessionManager+AudioFiles.swift`. A plain regression against main for in-person recording.

## Non-blocking but worth a look

**The disk pass cannot be interrupted.** `DiarizationManager+LongRecording.swift` wraps both passes in `Task.detached` (which does not inherit cancellation) and neither chunk loop checks `Task.isCancelled`. The 5 s guard in `RecordingSessionManager.swift:226-231` cannot save it: `withTaskGroup` awaits all remaining children before returning, and one child is `await task.value`, so `task.cancel()` is reached only after the thing it would cancel has finished. That block is byte-identical on main, so the ineffective timeout is **pre-existing and not this PR's doing** — but this PR removes the memory cap that kept the work behind it finite. On a 5-hour recording, stopping and immediately starting a new meeting leaves `recordingState == .starting` (and `stopRecording()` a no-op, since `currentMeetingID` is assigned after the wait) until tens of minutes of chunked work completes. CLAUDE.md asks for `Task.isCancelled` between expensive allocations and timeout guards on long loops.

**Memory is sized from physical rather than available RAM.** `ProcessInfo.processInfo.physicalMemory` gives total, not free. On an 8 GB Mac the resident Float32 timeline goes from ~115 MB to 512 MB, and that is handed to Parakeet and Sortformer simultaneously, possibly alongside a resident MLX LLM. CLAUDE.md: "Check available memory before loading models. Use `mach_task_basic_info`…" — nothing on this path does.

**Mute gaps are materialised as real zeroed samples** — inline on `AudioTimelineMixer.swift`.

**Silent `try?` on the converter drain** — inline on `AudioFileChunkReader.swift`.

## Smaller things

- **Three stale `// Extension-visible: +AudioFiles` markers** on `micSegments`, `systemSegments`, `systemRecordedDuration` (`RecordingSessionManager.swift:134-153`). I grepped `+AudioFiles.swift`: it references none of them — placements arrive via the `CaptureSources` parameter. They should be `private`. The same comment runs the other way at `+AudioFiles.swift:208`, which says the core file calls `moveRecordingFile` "when a recording stops" — this PR deleted that call.
- **The system-audio tap closure is duplicated byte-for-byte** at `RecordingSessionManager.swift:403-419` and `:582-598`, and this PR added the new `writtenSeconds` accumulation to *both* copies — in the same commit that extracted `AudioRecorder.installTap` because "[it] previously carried two copies of it and so two places for the two to drift apart". The same treatment would fit here.
- **`RecordingSessionManager+Diarization.swift` went 483 → 579 lines**, past the ~500 guideline for extension files, in the PR that split a file to respect that guideline.
- **`AudioFileChunkReader.duration(of:)` has no production caller** — only tests. Worth noting because it is exactly the tool blocking issue 1 needs.
- **`LongRecordingE2ETests` gating is partial.** `.enabled(if:)` checks only `LOGUE_LONG_AUDIO`, then `#require(fixture)` also demands `LOGUE_LONG_AUDIO_HOURS` — so setting one variable produces a hard failure rather than the skip the doc comment promises.
- **No `CHANGELOG.md` entry, no `WhatsNewCatalog` block, no `CLAUDE.md` update.** This fixes a user-visible 30-minute limit, and CLAUDE.md's own Releasing section makes the changelog the source for both Sparkle's release notes and the GitHub Release body. The Key Files table also now omits five new types, and the Architecture section still calls the pipeline "no chunking" and describes `DiarizationManager` without the disk route. The invariants this PR establishes ("a pass may only speak for the audio it heard", "the saved file is the meeting's timeline") are precisely the hard-won rules that section exists to record.

## Context on the test numbers

The 47 new tests are good — they assert against hand-computed values rather than restating the implementation, and the chunk reader uses a position-encoding ramp so drops or reorders cannot pass. Worth knowing, though: **CI runs no tests at all.** `.github/workflows/ci.yml` has only `Lint Check` and `Build`; there is no `xcodebuild test` step anywhere. So "935 tests pass" is a local result and none of this is a regression gate today. Pre-existing and not for this PR to fix, but it makes the gaps below matter more than they otherwise would — nothing composes a file, and neither `persistRecordingAudio` nor `runBatchPass` is called by any test, so the route-selection and trust decisions are untested. `AudioFileChunkReaderTests` also uses only mono WAV fixtures, while the disk route actually reads the composed AAC `.m4a` with up to two tracks.

westerosweb:
All blocking items addressed in `1bd1968`. Each was verified against the code rather than taken on trust — the findings held up.

## Blocking

**1. `.onSessionTimeline` proves alignment, not coverage.** Correct, and it is this branch's own bug class returning through the gate meant to prevent it. A mic whose writes start failing — which the tap deliberately tolerates — leaves a file that begins at zero, lines up perfectly, and stops early; trusting that deleted every live segment past its end. Coverage is now measured against the session duration, using `AudioFileChunkReader.duration` — which, as you noted, was sitting there with no caller outside tests. A file that falls short passes a real `heardDuration` instead of `nil`, so the live transcript beyond it survives.

**2. `result.duration` is always `0` on the disk route.** Confirmed in the pinned source: `transcribeDiskBacked` → `ChunkProcessor.process` finishes with `audioSamples: []`, and `processTranscriptionResult` derives the duration from that count. The fallback now returns `nil` and keeps the live transcript, which is segmented and correctly timed, rather than replacing a whole meeting with one segment from 0 to 0.

**3. No cancellation point.** Both chunk loops check `Task.checkCancellation()`, and the work runs inside the task tree rather than in a detached task that cannot inherit it. The wait that was meant to bound this is also fixed: you were right that a task group waits for every child and `await task.value` ignores the group's cancellation, so the timeout could only fire after the thing it meant to cut short had already finished. It now waits on the handle with a deadline, then cancels.

**4. Single-source mute forced a full AAC re-encode.** Composition is for *mixing*; a single source that was muted or joined late recorded every sample contiguously, so the gaps are spliced back as PCM instead. No encoder, no lossy copy for the transcriber to read back later, and no minute-long block in `.stopping`. Two sources still compose, exactly as on `main`.

## Non-blocking

- **Silent `try?` on the drain** — propagates now. It was swallowing the error in the one place whose purpose is not to lose audio silently.
- **Sized from installed rather than free memory** — now takes the smaller of a sixteenth of installed and a quarter of free, read via `host_statistics64`.
- **Materialised mute gaps** — this one is *not* fixed. See below.

## Smaller

Stale `Extension-visible: +AudioFiles` markers are private again, and the stale `moveRecordingFile` comment is gone. The system-audio tap closure did exist twice byte-for-byte — one installer now, like the mic one. `+Diarization` was over the guideline at 603 lines; the batch pass moved to its own file, leaving 499 and 117. The long-recording suite gates on both variables, so a half-set environment skips rather than fails.

`CHANGELOG.md` and `CLAUDE.md` are updated — the Key Files table now lists the new types, the Architecture section no longer claims the pipeline does no chunking, and the two invariants are written down.

## Two corrections

- There is no `WhatsNewCatalog` type anywhere in the repo, and `CLAUDE.md` has no Releasing section — `grep` finds neither. The changelog point stood and is done; those two had nothing to update.
- The "~50 s" figure for post-mute audio on an 8 GB Mac does not reconcile. A 2-hour gap leaves 19 M of 134 M samples, which is ~20 minutes. The ~90% capacity figure is right.

## Deferred, with a caveat you should know

The dense timeline still materialises mute gaps as real zeroed samples. Tracked as #50 with the measured costs. Worth flagging: **the available-memory change makes one case worse** — when free memory is low the budget falls to the 256 MB floor (~1.16 h), and a 2-hour mute then exhausts capacity outright rather than merely eating into it. Gaps over a minute are logged so it is visible, but not avoided.

The proper fix feeds the models each activation separately and offsets by its `sessionStart` — no silence in memory or through the models — but Sortformer's speaker state would then need verifying across activations on real audio, and its output mapping back through the placements. That is a design change with a verification cost, and I would rather do it deliberately than bolt it on here.

## Verification

**939 tests in 83 suites pass**; SwiftFormat and SwiftLint `--strict` clean. Four new tests cover the splice and exercise `persistRecordingAudio`, which you correctly identified as having no test at all.

You are also right that **CI runs no tests**, so every number here is local and none of it gates a merge. Left alone as pre-existing, but it deserves its own change.

Logue/Engine/CaptureSegmentTimeline.swift
westerosweb: Fixed in `1bd1968`. You were right, and this is the branch's own bug class arriving through the gate meant to stop it.

Coverage is now measured rather than assumed, using `AudioFileChunkReader.duration` — which, as you spotted, was sitting in the PR with no caller outside tests. `runBatchPass` compares the file against the session and passes a real `heardDuration` when it falls short, so the live transcript past the file's end survives instead of being deleted.

`fileMatchesSessionTimeline` keeps its narrower job — it answers *is this file aligned*, which is what `persistRecordingAudio` needs to decide between moving, splicing and composing. Coverage is a separate question asked at the point that actually depends on it.

Logue/Engine/DiarizationManager+LongRecording.swift
westerosweb: Fixed in `1bd1968`. Confirmed along the same path you traced: `transcribeDiskBacked` → `ChunkProcessor.process` finishes with `audioSamples: []`, and `processTranscriptionResult` divides that count by the sample rate.

Rather than deriving the end from the file, this now returns `nil` and keeps the live transcript. A single segment holding hours of text would replace a correctly segmented, correctly timed transcript with one unplaceable blob — worse than not replacing it at all, even with the right end time.

Logue/Services/RecordingSessionManager+AudioFiles.swift
westerosweb: Fixed in `1bd1968`. Composition is for *mixing*, and a single source needs none.

One source records every sample it heard, contiguously — a mute or a late join only makes the file shorter than the meeting, never incomplete. So the gaps are spliced back in as PCM: no encoder, no lossy copy for the long-recording pass to read back, and no minute-long block in `.stopping` on a long recording. Two sources still compose, exactly as on `main`.

Four tests cover it, including that the output keeps its container and returns sample values untouched — an AAC round-trip would fail both.

Logue/Engine/AudioTimelineMixer.swift
westerosweb: Not fixed — you framed it as a design question and I agree, so it is tracked as #50 with the measured costs rather than bolted on here.

One thing worth adding: **the available-memory change makes one case worse.** When free memory is low the budget falls to the 256 MB floor (~1.16 h), so a 2-hour gap now exhausts capacity outright rather than merely eating into it.

Also, the "~50 s" figure does not reconcile — a 2-hour gap leaves 19 M of 134 M samples, which is ~20 minutes. The ~90% capacity figure is right.

Gaps over a minute are now logged with their size, so the cost is visible rather than silent. The real fix feeds the models each activation separately and offsets by its `sessionStart` — no silence in memory or through the models — but Sortformer's speaker state needs verifying across activations on real audio first, and its output mapping back through the placements.

Logue/Engine/AudioFileChunkReader.swift
westerosweb: Fixed in `1bd1968` — it propagates now. `next()` already throws and every other error path in the file does the same, so this was just inconsistent as well as against the rule.

Logue/App/AppConstants.swift
westerosweb: Fixed in `1bd1968`. The budget is now the smaller of a sixteenth of installed memory and a quarter of what is actually free, read via `host_statistics64` (counting free plus inactive pages, since inactive is reclaimable and excluding it would floor the budget on any machine that has been awake a while).

Flagging a consequence: on a machine that is already tight this lowers capacity toward the 256 MB floor, which interacts badly with materialised mute gaps — noted on that thread and in #50.

shanforge (approved):
Re-reviewed `1faa6ec..317ef60`. **Approving.** All six findings are resolved — five fixed in `1bd1968`, one tracked as #50 by agreement — and I verified each in the code rather than taking the replies for it.

## First, a correction of my own

My "~50 s of post-break audio" figure in the mixer thread was wrong, and your number is right: a 2-hour gap leaves 19 M of a 134 M-sample budget, which is **~19.8 minutes**, not seconds. I checked the arithmetic before writing this. The ~90% capacity figure was the accurate half of that comment. Thanks for catching it rather than working around it.

## Verified fix by fix

**Alignment vs coverage** — `runBatchPass` now measures `AudioFileChunkReader.duration(of:)` against `sessionDuration` (plumbed from `finalElapsedTime - sessionStart`) and passes a real `heardDuration` when the file falls short. I re-walked my original scenario: 5 h session, mic writes failing from 4 h → `coversSession == false` → `heardFromFile == 4 h` → `TranscriptReplacement` keeps every live segment past it. The hour survives.

What I like more than the fix is that it **fails safe**: `duration(of:) ?? 0` makes an unreadable file report zero coverage, which keeps the entire live transcript rather than trusting a file it could not measure. Splitting the question in two — `fileMatchesSessionTimeline` answers *aligned*, `runBatchPass` answers *covers* — is the right seam, and it means the compose route inherits the same protection.

**`result.duration` on the disk route** — returns `nil` and keeps the live transcript. You took the stronger of the two options I offered, and the reasoning holds: a correctly segmented, correctly timed transcript beats one unplaceable blob even with the right end time.

**The AAC re-encode** — this is the change I scrutinised hardest, because replacing a compose with a hand-written splice trades a known cost for new code. It holds up. Format and container come from the source, gaps are written as silence and activations copied straight through in 64 K-frame blocks, placements are sorted before splicing, it runs off the main actor, and failure degrades splice → compose → `rawFallback` with the audio never discarded. Two sources still compose, exactly as on `main`.

The tests are the reason I am comfortable with it: asserting exact sample values at exact offsets (`out[Int(rate * 4)] == 0` for the muted stretch, `0.25` preserved through the splice) is a check an AAC round-trip could not pass, so the lossless claim is pinned rather than asserted.

**Silent `try?`** — propagates. **Total vs free RAM** — `min(installed/16, available/4)` via `host_statistics64`, counting free plus inactive. And thank you for flagging the interaction with #50 in both places rather than letting the two changes quietly compound.

**#50** — right call. I raised it as a design question and it is one; a fix that feeds each activation separately needs Sortformer's speaker state verified across activations on real audio, which is not this PR's job. Logging gaps over a minute makes the cost visible in the meantime.

## Beyond what I asked for

Cancellation checks in both chunk loops with the work moved into the task tree; the three `Extension-visible` markers back to `private`; the byte-duplicated system tap closure down to one; `+Diarization` split at 500 lines via `+BatchPass`; the E2E suite gating on both variables so a half-configured environment skips instead of failing.

And CLAUDE.md now records both invariants — "a pass may only speak for the audio it actually heard" and "the saved audio file is the meeting's timeline", including the sentence that coverage means *both* aligned and long enough. That is the part most likely to stop this class of bug returning, and it is worth more than any individual fix here.

## Verification

| | |
|---|---|
| Local suite on `317ef60` | **951 tests in 84 suites pass** (built and ran it myself) |
| SwiftFormat 0.62.1 | 0/443 files require formatting |
| SwiftLint 0.65.0 `--strict` | 0 violations |
| CI | Build + Lint green |
| Merge of `main` | brings #49 in byte-identical — I diffed both files against `origin/main` |

One standing caveat, unchanged and not yours to fix here: CI still runs no tests, so those 951 are a local result. Worth its own issue.

PR #49, opened 2026-08-01
fix(command-center): let the Ask Logue island be dismissed again

### LINKED ISSUE #46: Ask Logue dialogue never closes, remains on screen over other apps
### What happened?

When I open the 'Ask Logue' dialogue (cmd-shift-w), the input bar remains until I close or hide the app, even when I switch to another app.

### Steps to reproduce

1. open Ask Logue
2. observe input field/bar
3. cmd-tab to switch to a different app
4. note that Ask Logue dialogue is still present over other apps
5. cmd-tab back to Logue
6. cmd-shift-w does nothing to hide the dialogue (or anything else)

### Logue version

1.0.0 (7)

### macOS version

macOS 26.4 (Tahoe)

### Mac model

MBP, 2026 M5 Pro 48GB

### Screenshots or logs

_No response_

shanforge:
Fixes #46.

## What was reported

The Ask Logue bar stayed on screen over every other app, and `Cmd+Shift+W` did
nothing to put it away. Separately: clicking outside it dismissed it sometimes
and did nothing other times, and it could reach a state where quitting Logue was
the only way to get rid of it.

## Why it happened

**The shortcut rebuilt the island instead of closing it.** `showChatPanel` tore
the panel down and immediately built a new one, so pressing the shortcut on a
showing island just blinked it.

**That rebuild then stranded the panel.** Dismissal is animated, and the
completion handler ran ~0.15s later against whatever `panel` pointed at *by
then* — the replacement. It cleared the state describing a live window, and from
that moment every route out died at once: `dismissPanel` returned early on the
nil reference, so Esc, the close button and the click-outside handler were all
dead, and each further press stacked another orphan. This is both the "never
closes" in the title and the "only quitting the app clears it".

**Esc and click-outside could not see their events.** Both were watched with
global monitors only, and a global monitor never receives events routed to our
own app. Creating the panel makes it key and activates Logue, so Esc was
guaranteed not to work right after summoning it, and clicking Logue's own window
beside the island did nothing while clicking another app worked.

**Nothing watched activation for chat.** The app-active observers were only
installed for the recording island. The chat island is created at the floating
level with `hidesOnDeactivate` off and `canJoinAllSpaces` on — right for a
recording island, wrong for a prompt bar — so it sat above everything until
Logue was quit or hidden.

## What changed

- Dismissal is atomic: `relinquishPanel()` clears every piece of state describing
  the panel and hands it back to be animated away, so a dying panel can never be
  mistaken for a live one.
- The shortcut dismisses an island that is up in front, and *raises* one that was
  sent behind other apps rather than destroying it.
- Esc and click-outside are watched globally *and* locally, so they behave the
  same whichever app is frontmost. Local monitors are chat-only; the recording
  island keeps the behaviour it had.
- Losing frontmost status is now meaningful to the chat island: an empty one is
  discarded, matching what a click outside it does.

## What the island protects, and from what

The two are deliberately not the same promise:

- **A conversation** must survive a stray click off the island. Unchanged.
- **An unsent prompt** survives switching apps — you are coming back to it — but
  not Esc, the close button, the shortcut or a click off the island, which are
  all decisions to put the island away and have always discarded it.

Collapsing those into one flag was tried and reverted: it left a draft-only
island with no way out at all, and turned the transparent region above the pill
into a dead zone. The container now asks whether a click is its own before
claiming it during hit testing, so clicks it will not act on pass through.

## Tests

The decisions that were wrong lived inside the window code, where no test could
reach them. They are now `CommandCenterChatRule`, which knows nothing about
`NSPanel` — the same pure-rules pattern as `ExternalChangePlan` and
`SpaceFolderAdoption`.

- 10 rule tests over the trigger, focus-loss and content matrices
- Full non-LLM suite green: 900 tests in 77 suites
- SwiftFormat and SwiftLint `--strict` run locally against the pinned CI
  versions (0.62.1 / 0.65.0): clean

## Checked by hand, not covered by tests

The AppKit half — window level changes, monitor lifetime, panel identity — is
not reachable from the test target. Worth a reviewer's eye on the recording
island, which shares `dismissPanel` and the activation observers with chat.

Known and deliberately not addressed here: a transient deactivation such as a
first-run microphone permission prompt will dismiss an empty island, and key
codes are still spelled three ways across `CommandCenterController`,
`ShortcutManager` and `AppDelegate`. Both deserve their own change.

https://claude.ai/code/session_01565d5yT4e5oTCNdxdWvtVP

### DIFF
--- a/Logue/CrossApp/CommandCenterChatRule.swift
+++ b/Logue/CrossApp/CommandCenterChatRule.swift
@@ -21,10 +21,11 @@ enum CommandCenterChatRule {
     /// What losing frontmost status should do to a chat island, or `nil` when the
     /// showing panel is not one.
     enum FocusLoss: Equatable {
-        /// Nothing has been said to it, so nothing is lost by closing it.
+        /// Empty, so nothing is lost by closing it.
         case dismiss
-        /// It holds a conversation that only exists in the view's state. Keep it,
-        /// but stop it covering the app being switched to.
+        /// It holds a conversation, or a prompt typed but not sent, and both live
+        /// only in the view's state. Keep it, but stop it covering the app being
+        /// switched to.
         case sendBehindOtherApps
     }
 
@@ -34,8 +35,8 @@ enum CommandCenterChatRule {
         return .replace
     }
 
-    static func focusLoss(mode: CommandCenterMode?, chatHasMessages: Bool) -> FocusLoss? {
+    static func focusLoss(mode: CommandCenterMode?, chatHasContent: Bool) -> FocusLoss? {
         guard case .chat = mode else { return nil }
-        return chatHasMessages ? .sendBehindOtherApps : .dismiss
+        return chatHasContent ? .sendBehindOtherApps : .dismiss
     }
 }
--- a/Logue/CrossApp/CommandCenterChatRule.swift
+++ b/Logue/CrossApp/CommandCenterChatRule.swift
@@ -1,12 +1,5 @@
 import Foundation
 
-/// What a Command Center trigger, or Logue losing frontmost status, means for the
-/// chat island.
-///
-/// Kept free of AppKit so the matrix is testable without an NSPanel. The controller
-/// owns the panel and only asks what to do with it — issue #46 was three separate
-/// wrong answers to these questions, none of which a test could have reached while
-/// they lived inside the window code.
 /// What the chat island is currently holding.
 ///
 /// The two are kept apart because they earn different protection. Deliberately
@@ -23,6 +16,13 @@ struct CommandCenterChatContent: Equatable {
     }
 }
 
+/// What a Command Center trigger, or Logue losing frontmost status, means for the
+/// chat island.
+///
+/// Kept free of AppKit so the matrix is testable without an NSPanel: the controller
+/// owns the panel and only asks what to do with it. Issue #46 was several wrong
+/// answers to these questions, none of which a test could reach while they lived
+/// inside the window code.
 enum CommandCenterChatRule {
     /// What pressing the Ask Logue shortcut should do.
     enum Trigger: Equatable {
--- a/Logue/CrossApp/CommandCenterChatRule.swift
+++ b/Logue/CrossApp/CommandCenterChatRule.swift
@@ -10,8 +10,12 @@ import Foundation
 enum CommandCenterChatRule {
     /// What pressing the Ask Logue shortcut should do.
     enum Trigger: Equatable {
-        /// The chat island is already up — the shortcut puts it away.
+        /// The chat island is up in front — the shortcut puts it away.
         case dismiss
+        /// The island was sent behind other apps when Logue lost focus. The
+        /// shortcut is how you ask for it back, so bring it forward rather than
+        /// destroying it along with whatever was typed into it.
+        case raise
         /// Another island is up and has to go first.
         case replace
         /// Nothing is up.
@@ -29,10 +33,14 @@ enum CommandCenterChatRule {
         case sendBehindOtherApps
     }
 
-    static func trigger(mode: CommandCenterMode?, isShowingPanel: Bool) -> Trigger {
+    static func trigger(
+        mode: CommandCenterMode?,
+        isShowingPanel: Bool,
+        isChatBehindOtherApps: Bool
+    ) -> Trigger {
         guard isShowingPanel else { return .present }
         if case .chat = mode {
-            return .dismiss
+            return isChatBehindOtherApps ? .raise : .dismiss
         }
         return .replace
     }
--- a/Logue/CrossApp/CommandCenterChatRule.swift
+++ b/Logue/CrossApp/CommandCenterChatRule.swift
@@ -7,6 +7,22 @@ import Foundation
 /// owns the panel and only asks what to do with it — issue #46 was three separate
 /// wrong answers to these questions, none of which a test could have reached while
 /// they lived inside the window code.
+/// What the chat island is currently holding.
+///
+/// The two are kept apart because they earn different protection. Deliberately
+/// putting the island away — clicking off it, Esc, the close button, the shortcut
+/// — has always destroyed an unsent prompt, and a conversation is what a click
+/// off the island must not throw away. Switching apps is not a decision about the
+/// island at all, so that leaves either alone.
+struct CommandCenterChatContent: Equatable {
+    var hasMessages: Bool
+    var hasDraft: Bool
+
+    var isEmpty: Bool {
+        !hasMessages && !hasDraft
+    }
+}
+
 enum CommandCenterChatRule {
     /// What pressing the Ask Logue shortcut should do.
     enum Trigger: Equatable {
--- a/Logue/CrossApp/CommandCenterChatRule.swift
+++ b/Logue/CrossApp/CommandCenterChatRule.swift
@@ -31,7 +31,9 @@ enum CommandCenterChatRule {
 
     static func trigger(mode: CommandCenterMode?, isShowingPanel: Bool) -> Trigger {
         guard isShowingPanel else { return .present }
-        if case .chat = mode { return .dismiss }
+        if case .chat = mode {
+            return .dismiss
+        }
         return .replace
     }
 
--- a/Logue/CrossApp/CommandCenterChatRule.swift
+++ b/Logue/CrossApp/CommandCenterChatRule.swift
@@ -1,0 +1,41 @@
+import Foundation
+
+/// What a Command Center trigger, or Logue losing frontmost status, means for the
+/// chat island.
+///
+/// Kept free of AppKit so the matrix is testable without an NSPanel. The controller
+/// owns the panel and only asks what to do with it — issue #46 was three separate
+/// wrong answers to these questions, none of which a test could have reached while
+/// they lived inside the window code.
+enum CommandCenterChatRule {
+    /// What pressing the Ask Logue shortcut should do.
+    enum Trigger: Equatable {
+        /// The chat island is already up — the shortcut puts it away.
+        case dismiss
+        /// Another island is up and has to go first.
+        case replace
+        /// Nothing is up.
+        case present
+    }
+
+    /// What losing frontmost status should do to a chat island, or `nil` when the
+    /// showing panel is not one.
+    enum FocusLoss: Equatable {
+        /// Nothing has been said to it, so nothing is lost by closing it.
+        case dismiss
+        /// It holds a conversation that only exists in the view's state. Keep it,
+        /// but stop it covering the app being switched to.
+        case sendBehindOtherApps
+    }
+
+    static func trigger(mode: CommandCenterMode?, isShowingPanel: Bool) -> Trigger {
+        guard isShowingPanel else { return .present }
+        if case .chat = mode { return .dismiss }
+        return .replace
+    }
+
+    static func focusLoss(mode: CommandCenterMode?, chatHasMessages: Bool) -> FocusLoss? {
+        guard case .chat = mode else { return nil }
+        return chatHasMessages ? .sendBehindOtherApps : .dismiss
+    }
+}
--- a/Logue/CrossApp/CommandCenterController.swift
+++ b/Logue/CrossApp/CommandCenterController.swift
@@ -500,36 +500,40 @@ class CommandCenterController: ObservableObject {
         }
     }
 
-    /// Shared cleanup after a panel closes. Resets panel state and optionally
-    /// tears down recording observers if no recording is actively running.
+    /// Gives up ownership of the showing panel and hands it to the caller to
+    /// animate away. Every piece of state describing it is cleared here, before
+    /// returning — a dismissal takes 150–300ms to animate, and for that whole
+    /// window anything reading `panel` or `currentMode` would otherwise be
+    /// answered about a panel that is already leaving. That stale answer is what
+    /// made the shortcut, a menu-bar click and the dismiss notification act on a
+    /// dying panel instead of presenting a new one (issue #46).
     ///
-    /// `dismissed` is the panel whose close animation just finished. Dismissal is
-    /// animated, so a new panel can be created before this runs — clearing state
-    /// then would strand the live panel with no reference to it, leaving it on
-    /// screen with nothing able to close it (issue #46).
-    private func finalizeDismiss(_ dismissed: CommandCenterPanel, preserveRecordingState: Bool = false) {
-        guard panel === dismissed else { return }
+    /// Returns `nil` when there is nothing showing.
+    private func relinquishPanel() -> CommandCenterPanel? {
+        guard let dismissed = panel else { return nil }
+
+        // Keep recording state if a recording is still running (allows re-show from menu)
+        let keepRecordingState = RecordingSessionManager.shared.isRecording && activeRecordingMeetingID != nil
+
         panel = nil
         isVisible = false
         currentMode = nil
         chatHasContent = false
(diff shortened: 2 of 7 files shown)


Logue/CrossApp/CommandCenterChatRule.swift
```diff
@@ -0,0 +1,44 @@
...
+        /// switched to.
+        case sendBehindOtherApps
+    }
+
+    static func trigger(mode: CommandCenterMode?, isShowingPanel: Bool) -> Trigger {
+        guard isShowingPanel else { return .present }
+        if case .chat = mode {
+            return .dismiss
+        }
+        return .replace
```
westerosweb, line 61: **Critical — the shortcut destroys a demoted island and the draft it was protecting.**

Trace the case `.sendBehindOtherApps` exists for:

1. Type a prompt, don't send, switch to Chrome.
2. `handleAppResignedActive` → `.sendBehindOtherApps` → `panel.level = .normal`. The island is now behind Chrome's window, i.e. invisible.
3. Press the Ask Logue shortcut to bring it back. Chrome is still frontmost, so the **global** hotkey monitor fires → `showChatInterface()` → `toggleChatPanel()`.
4. `trigger(mode: .chat, isShowingPanel: true)` → `.dismiss`. The island and the draft are destroyed, and nothing appears on screen. Logue was never activated, so `handleAppBecameActive` never runs and the island is never re-promoted.
5. `ShortcutManager.handleKeyEvent` debounces at 500 ms, so the reflexive second press is swallowed. The third gives an empty island — draft gone.

So the draft survives the app switch and is then thrown away by the most natural gesture for coming back to it, and the intermediate press reads as "the shortcut does nothing" — the exact symptom of #46. The demoted island *is* resurrectable to floating, but only by activating Logue some other way (Dock, main window, ⌘Tab); the shortcut path cannot reach it.

The rule needs a third answer rather than a fifth special case in the controller:

```swift
static func trigger(mode: CommandCenterMode?, isShowingPanel: Bool, isChatBehindOtherApps: Bool) -> Trigger {
    guard isShowingPanel else { return .present }
    if case .chat = mode { return isChatBehindOtherApps ? .raise : .dismiss }
    return .replace
}
```

`.raise` maps to `panel.level = .floating` + `orderFrontRegardless()` + `makeKey()` + activate. The controller can supply `isChatBehindOtherApps` as `panel?.level == .normal` (or `!NSApp.isActive`), so the rule stays AppKit-free — plus the two new matrix rows in the test suite.

Logue/CrossApp/CommandCenterController.swift
```diff
@@ -382,26 +410,34 @@ class CommandCenterController: ObservableObject {
...
+            if event.keyCode == Self.escapeKeyCode {
                 Task { @MainActor in self?.dismissPanel() }
             }
         }
 
+        escLocalMonitor = NSEvent.addLocalMonitorForEvents(matching: .keyDown) { [weak self] event in
+            guard event.keyCode == Self.escapeKeyCode, let self, panel?.isKeyWindow == true else { return event }
+            dismissPanel()
+            return nil
+        }
```
westerosweb: **Important — this monitor is installed for the recording island too, which changes its behaviour.**

The click monitors below are gated on `if case .chat = mode`; this one isn't. `createPanel` calls `panel.makeKey()` and `NSApp.activate(...)`, so immediately after a recording starts the recording island *is* the key window — and Esc now collapses it while Logue is frontmost, which it never did before (the global monitor only fired when another app was frontmost).

It's recoverable, and arguably consistent, but the PR says the recording island should be unaffected and this is an unmentioned change to a shared path. If it's deliberate, say so in the comment and the changelog; if not, move `escLocalMonitor` inside the existing `if case .chat = mode` block.

Logue/CrossApp/CommandCenterController.swift
```diff
@@ -412,28 +448,54 @@ class CommandCenterController: ObservableObject {
...
+        guard let window = event.window else { return event.locationInWindow }
+        return window.convertPoint(toScreen: event.locationInWindow)
+    }
+
+    /// Puts an empty chat island away when a click lands anywhere but on it. An
+    /// island holding a conversation or an unsent prompt stays.
+    private func dismissIfClickMissedPanel(_ screenPoint: NSPoint) {
+        guard let panel, !chatHasContent, !panel.frame.contains(screenPoint) else { return }
+        dismissPanel()
+    }
```
westerosweb, line 474: **Important — after this change, one keystroke creates an invisible 740×600 dead zone.**

Two separate rules decide what a click in the transparent area means and they now disagree. `TransparentContainerView.hitTest` returns `self` for the whole empty region *because `onClickEmptyArea` is non-nil*, so those clicks never reach the window underneath — and `onClickEmptyArea` no-ops whenever `chatHasContent` is true. The chat panel is 740×600 while the pill is only ~50 px of it at the bottom, so after a single typed character the region above the pill swallows clicks and does nothing with them. The user sees no island there, clicks the document behind it, and nothing happens. Previously that click at least dismissed the empty island.

Pre-existing wart, but this change materially aggravates it. The clean fix also unifies the two rules: compare clicks against the **hosting view's** frame in screen coordinates rather than `panel.frame` here, and let `hitTest` return `nil` for the empty region so clicks pass through. That leaves one rule — "a click outside the pill" — instead of two.

Logue/CrossApp/CommandCenterController.swift
```diff
@@ -412,28 +448,54 @@ class CommandCenterController: ObservableObject {
...
+    /// animated, so a new panel can be created before this runs — clearing state
+    /// then would strand the live panel with no reference to it, leaving it on
+    /// screen with nothing able to close it (issue #46).
+    private func finalizeDismiss(_ dismissed: CommandCenterPanel, preserveRecordingState: Bool = false) {
+        guard panel === dismissed else { return }
         panel = nil
         isVisible = false
         currentMode = nil
-        chatHasMessages = false
+        chatHasContent = false
```
westerosweb: **Important — the identity guard fixes the catastrophic case, but the transition is still re-enterable.**

`dismissPanel()` leaves `panel` and `currentMode` describing a panel that is already going away for the 150–300 ms of the animation, so anything consulting them in that window gets a stale answer. Concretely: with an empty island up, clicking the menu bar status item fires `clickLocalMonitor` → `dismissPanel()` (the main-actor `Task` drains during menu tracking), and choosing "Ask Logue" inside 150 ms has `toggleChatPanel` see `panel != nil` / `.chat` and dismiss again instead of presenting — a swallowed activation. The 500 ms hotkey debounce masks the keyboard route but not the mouse or notification routes.

Making the transition atomic is the stronger answer: set `panel = nil` / `currentMode = nil` synchronously at the top of `dismissPanel()`, keep only the local reference for the animation, and reduce this to the recording-state part. That removes the need for the identity guard entirely rather than leaving a guard every future call site has to remember.

Logue/Views/CrossApp/CommandCenterChatView.swift
```diff
@@ -45,6 +45,12 @@ struct CommandCenterChatView: View {
...
 
+    /// Whether the island is holding anything the user would miss — a sent
+    /// conversation, or a prompt typed but not sent yet.
+    private var hasContent: Bool {
+        !messages.isEmpty || !inputText.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+    }
+
     var body: some View {
         VStack(spacing: 0) {
             if !messages.isEmpty {
```
westerosweb, line 57: **Important — an island holding only a draft has no visible way out.**

The close (✕) button lives inside `messagesPanel`, which is only rendered when `messages` is non-empty. Combine that with the other two changes: click-outside is now suppressed by `chatHasContent`, and the Esc local monitor requires `panel?.isKeyWindow == true` while the global one can't see Logue's own events. So with a prompt typed but not sent and the main window focused, there is no affordance, no Esc, and only the keyboard shortcut — which per the other comment destroys the draft.

That's a thinner escape hatch than a PR about "it can't be dismissed" should leave behind. Minimum fix: make ✕ unconditional (move it into `promptPill`, or render the header whenever `hasContent`).

Logue/Views/CrossApp/CommandCenterChatView.swift
```diff
@@ -45,6 +45,12 @@ struct CommandCenterChatView: View {
 
     private let pillWidth: CGFloat = 740
 
+    /// Whether the island is holding anything the user would miss — a sent
+    /// conversation, or a prompt typed but not sent yet.
+    private var hasContent: Bool {
+        !messages.isEmpty || !inputText.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+    }
```
westerosweb: Minor — the draft-preservation guarantee is narrower than the description reads. Focus loss goes out of its way to keep an unsent prompt, but Esc, the shortcut and ✕ all close the panel, which destroys the hosting view and this `@State`. If a draft is worth preserving it belongs outside the view (controller-held, restored into the field on the next open); if it isn't, `hasContent` gating click-outside is doing a lot of work — including the dead zone above — for something the user can lose by pressing Esc.

LogueTests/CommandCenterChatRuleTests.swift
```diff
@@ -0,0 +1,65 @@
+import Foundation
+@testable import Logue
+import Testing
+
+/// The chat island's lifecycle rules — issue #46.
+///
+/// Reported as "the Ask Logue bar never closes, and sits over every other app".
+/// Each of the three causes has a test here: the shortcut rebuilt the island
+/// instead of closing it, and losing frontmost status meant nothing at all to a
+/// chat island, whether or not anything had been said to it.
```
westerosweb: Minor — "each of the three causes has a test here" isn't quite accurate, and the gap is in the risky direction. These 8 tests cover two 4-line pure functions, a couple of them restating the implementation (`focusLossIgnoresIdle`, `focusLossIgnoresRecording`). The fixes most likely to be broken by a future edit — panel identity in `finalizeDismiss`, and monitor add/remove symmetry — aren't reachable from here at all.

Either soften the comment, or extract the testable seam: a small `PanelLifecycle` value owning the `panel`/`currentMode`/`isDismissing` transitions could be asserted without AppKit, which would also pin down the re-entrancy fix. Missing matrix rows either way: `focusLoss(mode: .recording, chatHasContent: true)` and `trigger(mode: nil, isShowingPanel: true)` (currently `.replace`, which is right, but unpinned).

westerosweb (commented):
Reviewed `21fa2ea..74c43e1`. Verdict: **merge with fixes** — one path re-creates the reported symptom and loses a draft with it; the rest is sound.

### What holds up

- The root-cause fix is right: `finalizeDismiss` compares by reference identity and both animation completions capture the local `panel` strongly with `self` weak, so cleanup can never land on a replacement.
- Monitor lifetime is airtight — all four monitors are stored and removed in `removeMonitors()`, which is the *first* statement of both dismiss paths, so a nil panel can't strand one.
- No observer double-registration: `setupAppActiveObservers()` tears down first, so the new call site in `toggleChatPanel` is safe across repeated opens.
- The recording island really is untouched on the rule path — `focusLoss` returns `nil` for `.recording` *and* for `nil` mode, falling through to the byte-identical old guard. One exception, inline below.
- Project standards are clean on this diff: `[weak self]` throughout, block-based observers with stored tokens, no silent `try?`, no force unwraps, no inline delay literals, file and function lengths inside limits.

### Critical

The shortcut destroys a demoted island *and* its unsent draft — see the inline comment on `CommandCenterChatRule.trigger`.

### Important

1. Esc's local monitor is not gated on chat mode, so it now collapses the recording island while Logue is frontmost.
2. A draft-only island has no visible way out: no ✕, click-outside suppressed, Esc dead unless the panel is key.
3. Widening `chatHasContent` turns the transparent region above the pill into a click-swallowing dead zone.
4. `dismissPanel()` leaves `panel`/`currentMode` describing a panel that is already going away for 150–300 ms.

### Minor

- `.dismissCommandCenter` is observed but never posted anywhere in the tree, so one of the listed dismissal routes only looks present. Worth wiring up or deleting rather than leaving it half-there.
- `focusLoss` → `.dismiss` fires on any transient deactivation, including a TCC prompt. The island's own mic button can trigger the microphone/speech permission dialog on first use, resign active, and dismiss the island out from under the flow — worth a check on a machine that hasn't granted mic access.
- Key codes are now spelled three ways: `escapeKeyCode` here, raw `34` in `AppDelegate`, raw `61` in `ShortcutManager`. Extracting the new one was right; a shared `KeyCodes` namespace would finish the thought.

shanforge:
Worked through all seven. Every one reproduced against the code — nothing dismissed as noise.

### Critical — demoted island destroyed with its draft (`a17bbbb`)

Confirmed exactly as traced. Pressing the shortcut from another app never activates Logue, so `handleAppBecameActive` never fires, the island stays demoted, and the trigger reads "already showing" → `.dismiss`. Draft gone, nothing on screen, debounce swallowing the second press. That is #46's symptom reintroduced by its own fix.

Took the suggested shape: `.raise` as a fourth answer, `isChatBehindOtherApps` supplied by the controller as `panel?.level == .normal` so the rule stays AppKit-free, plus the matrix rows.

### Important 1 — Esc's local monitor ungated (`95d8ec9`)

Correct, and it contradicted this PR's own claim that the recording island was untouched. Moved inside the existing `if case .chat = mode` block. Not deliberate — the recording island keeps exactly the Esc behaviour it had.

### Important 4 — re-enterable transition (`1102bf9`)

Took the stronger answer rather than keeping the guard. `relinquishPanel()` clears every piece of state describing the panel and hands it back for the caller to animate away, so there is no window where a dying panel reads as live. `finalizeDismiss` and the identity guard are gone — the invariant holds by construction now, with nothing for a future call site to remember.

### Important 2 + 3 + the Minor on draft preservation (`e83f2cc`, `c80577a`)

These three turned out to be one mistake, and the fix is not the one suggested. I had widened a single flag to mean "holds anything", which is what created the dead zone *and* removed the last way out of a draft-only island. But as the Minor comment says, the draft guarantee was incoherent anyway — Esc, ✕ and the shortcut all destroy it.

So the flag was wrong, not the button. Split into `CommandCenterChatContent { hasMessages, hasDraft }`:

- Clicking off the island, Esc, ✕, the shortcut — deliberate "put it away" gestures, which have always discarded an unsent prompt. Gated on `hasMessages`, exactly as before this PR.
- Switching apps — not a decision about the island. Gated on `!isEmpty`, so it keeps either.

Click-outside works again for a draft-only island, so the missing ✕ is back to its pre-existing state rather than a new gap. Separately fixed the container to ask *whether* a click is ours before claiming it in `hitTest`, instead of claiming everything and no-oping later — that removes the dead zone at its root, for messages too.

### Minor — dead `.dismissCommandCenter` (`475534e`)

Declared, observed, never posted anywhere in the tree. Deleted, and dropped from the PR description, which had counted it as a working way out.

### Minor — test comment overclaimed (`a17bbbb`)

Softened; it now says what these cover and names what it cannot reach (panel identity, monitor symmetry). Added both missing rows plus the `.raise` pair and the content matrix — 10 rule tests, 900 total.

### Not taken

**TCC prompt dismissing an empty island.** Real risk and I can't rule it out, but I have no machine without mic access to check on, and I'd rather not guess at a fix for a path I can't reproduce. Left alone and worth its own issue.

**Shared `KeyCodes` namespace.** Agreed, but it means touching `ShortcutManager` and the inline-assistant hotkey in `AppDelegate` — unrelated features in a bug-fix PR. Better on its own.

SwiftFormat 0/436 and SwiftLint `--strict` clean against the pinned CI versions; 900 tests in 77 suites green.

westerosweb (approved):

PR #52, opened 2026-08-02
Let the panes follow the window instead of fixed numbers

westerosweb:
Three fixes to how the main window divides its width. They arrived as one report, and the last two turned out to share a root cause.

## The editor column stopped at 720

`DocumentWorkspaceView` clamped the whole `BlockEditorView` — scroll view included — to a flat 720/1100 and centred it. On a large display the column never grew, and the scroller rode the column's right edge rather than the pane's, so it floated in the middle of the window beside the text.

The column now grows as a share of its pane. `EditorContentWidth` is the whole rule:

```
.scaling(mode) → min(cap, max(pane × fraction, min(pane, base)), pane)
.fixed(w)      → min(w, pane)          // focus mode keeps its own column
```

| mode | fraction | base | cap |
| --- | --- | --- | --- |
| normal | 0.70 | 720 | 900 |
| wide | 0.90 | 1100 | 1400 |

The `max(…, min(pane, base))` term matters: without it a 1000pt pane would have gone 720 → 700, so widening the window would have narrowed the text on laptop-sized windows. With it the column is never narrower than it was before at any pane width.

The clamp also moves *inside* the `ScrollView` — outside the existing 24pt insets, so the text measure is unchanged — which is what puts the scroller on the pane's edge.

## The inspector could only reach 520pt

Raised to 900, with a second limit derived from the window: the panel stops where the content pane would have to give up its reading measure (720). The tighter of the two wins, because neither works alone — a fixed ceiling is cramped on a 32" display and still crushes the editor on a 13" one, while a purely window-relative limit allows a ~2700pt panel on an ultrawide.

`SidebarWidthLimit` holds that rule for both sidebars. The navigation column goes 300 → 400 on the same reasoning but with a **static** ceiling: measuring the window there means writing state during the split view's first layout, and that split view persists any collapse it reports.

## The window's sidebar toggle did nothing

The visibility binding discarded every "list is visible" report and recorded only collapses. Two bugs from one rule:

- Pressing the toggle reported "visible", the report was dropped, nothing happened. AppKit merely restored its own saved column width, which read as the panel jumping wider.
- Dragging the sidebar shut reported "collapsed", which *was* recorded — and nothing could undo it. The only way back was the View menu.

The giveaway was the button's accessibility label reading **"Show Sidebar"** while the sidebar was plainly visible: AppKit and SwiftUI had drifted apart because half the reports were being thrown away.

Both directions are now honoured, telling a real report from SwiftUI's echo by comparing against the stored mode:

```swift
guard listIsVisible != current.showsList else { return nil }  // echo
return listIsVisible ? current.withList : current.withoutList
```

`current` must be read fresh from `UserDefaults` rather than from the caller's `@AppStorage` — the wrapper caches, and that stale read is the original bug the old asymmetry was working around.

## Closing the left sidebar also closed the right

Exposed by the fix above. Collapsing landed on `editorOnly`, which drops the inspector too — because "no list, inspector still up" was not a state the three layout modes could express. There are four combinations of (list, inspector) and the enum modelled three.

It is now the fourth case, `editorAndInspector`, which the View menu picks up on ⌘4 automatically. Hiding the list leaves the inspector exactly where it was.

## Verification

960 → 973 tests, all passing; SwiftFormat and SwiftLint `--strict` clean. The new rules are pure types, so they are tested directly rather than through a view:

- `EditorContentWidthTests` — never narrower than before, never wider than the pane, monotonic, both caps reached, degenerate panes clamped
- `SidebarWidthLimitTests` — content keeps its measure at every window size, both limits bind where expected, chosen width survives a shrink
- `DocumentIconTests` — echo suppression, the hide/show round trip, and that neither direction disturbs the inspector

Also checked in the running app: with both panes up, collapsing the left leaves the right in place, and the toggle round-trips.


LogueTests/DocumentIconTests.swift
```diff
@@ -112,50 +112,79 @@ struct EditorLayoutModeTests {
...
-    func collapseFromAllPanelsDropsTheInspector() {
-        #expect(EditorLayoutMode.allPanels.showsInspector)
-        let afterCollapse = EditorLayoutMode.modeAfterVisibilityReport(listIsVisible: false)
-        #expect(afterCollapse == .editorOnly)
-        #expect(afterCollapse?.showsInspector == false)
+    /// The regression the echo test exists for. SwiftUI echoes the split view's visibility back
+    /// through the binding right after a menu item sets a mode. A report that agrees with the
+    /// stored mode is that echo, and acting on it is how every ⌘3 pressed from editor-only used
+    /// to land on editor-and-list.
+    @Test("An echo of the stored mode changes nothing", arguments: EditorLayoutMode.allCases)
```
shanforge: _(Anchored here because the lines it concerns sit above the diff.)_

Three tests above the diff (lines 59–84) were not updated for the fourth case. All three still pass, because each enumerates cases by hand rather than iterating `allCases` — so the full suite stays green while two of the names now assert something the code contradicts.

**Line 72 — `inspectorVisibility`**
```swift
@Test("Only the all-panels layout shows the inspector")
```
No longer true: `editorAndInspector.showsInspector` is also `true`. The name states an invariant the PR deliberately breaks, and the body never mentions the new case.

**Line 79 — `listVisibility`**
```swift
@Test("Editor-only hides the list, the other layouts show it")
```
Also no longer true — `editorAndInspector` hides the list too, which is the whole point of it existing.

**Line 59 — `shortcutMapping`**
```swift
@Test("Each shortcut number maps to a distinct layout")
func shortcutMapping() {
    #expect(EditorLayoutMode(shortcutNumber: 1) == .editorOnly)
    #expect(EditorLayoutMode(shortcutNumber: 2) == .editorAndList)
    #expect(EditorLayoutMode(shortcutNumber: 3) == .allPanels)
}
```
Missing `#expect(EditorLayoutMode(shortcutNumber: 4) == .editorAndInspector)`. `everyLayoutHasAShortcut` proves 4 resolves to *something*, but nothing pins it to `editorAndInspector` specifically, so swapping the 3 and 4 arms would pass.

Suggested: fold the first two into `allCases`-driven checks so a fifth case can't slip past them the same way.

Separately, and pre-existing rather than yours: `EditorLayoutModeTests` and now `SidebarWidthLimit`-adjacent layout tests live in a file called `DocumentIconTests.swift`. This PR roughly doubles what's in there under that misleading name — a reasonable moment to split it into `EditorLayoutModeTests.swift`.

Logue/Views/Components/UnifiedSidebarView.swift
```diff
@@ -13,14 +13,23 @@ struct UnifiedSidebarView<Tool: ToolbarTool, PanelContent: View>: View {
...
 
     private let defaultWidth: CGFloat = 320
-    private let minWidth: CGFloat = 260
-    private let maxWidth: CGFloat = 520
+    private let limit = SidebarWidthLimit.inspector
+
+    /// The width actually rendered. Held separate from `currentWidth` so that shrinking
+    /// the window narrows the panel without forgetting the width the user chose — grow
+    /// the window back and the panel returns to it.
+    private var effectiveWidth: CGFloat {
```
shanforge, line 34: The doc comment claims more than the code delivers:

> shrinking the window narrows the panel without forgetting the width the user chose — grow the window back and the panel returns to it

True for a **window resize**, which leaves `currentWidth` untouched — and `chosenWidthSurvivesAShrink` tests exactly that path. Not true once a **drag** happens in the shrunken window: `onChanged` assigns the *clamped* value back into `currentWidth`, and `onEnded` persists it to `panelWidths["_shared"]`. The remembered wider width is then gone for good.

I think that's the behaviour you want — `dragStartWidth = effectiveWidth` says a drag re-chooses from where the handle actually is, which is right. It's the comment that overstates. Something like "a window resize does not overwrite the chosen width; a drag does" would match.

Minor related note: `currentWidth` is restored in `.onAppear` only. If two workspaces ever share the same `panelWidths` binding in one window, the second won't observe the first's change.

Logue/Models/EditorContentWidth.swift
```diff
@@ -0,0 +1,36 @@
...
+    func resolved(forPaneWidth paneWidth: CGFloat) -> CGFloat {
+        // A pane can measure zero — or briefly negative — during window setup and
+        // sidebar animations; a negative frame width traps in SwiftUI.
+        let pane = max(0, paneWidth)
+
+        switch self {
+        case let .fixed(width):
+            return min(width, pane)
+        case let .scaling(mode):
+            let grown = max(pane * mode.widthFraction, min(pane, mode.baseContentWidth))
```
shanforge, line 32: This line is the whole PR, and it's easy to misread as redundant — worth the comment it already has.

Without `min(pane, base)` the column would go **720 → 700** at a 1000pt pane, so widening the window would narrow the text on laptop-sized windows. I checked the boundary: the fraction only overtakes the base at `pane = base / fraction`, i.e. ~1029pt for normal and ~1222pt for wide. Below that the base holds, above it the column grows — so there's no discontinuity and no regression at any width.

The `max(0, paneWidth)` guard above is also load-bearing rather than defensive noise: a negative frame width traps in SwiftUI, and panes do briefly measure negative during sidebar animations. `degeneratePaneIsClamped` covers it.

shanforge (approved):
## Approving

The three fixes are correct, and the decision to extract `EditorContentWidth`, `SidebarWidthLimit` and `withoutList`/`withList` as pure types is what makes them reviewable at all — the rules are readable in isolation and tested without a view.

### Verified independently

I did not take the PR description on trust. Checked out `a012e83`, regenerated the project, built and ran the three suites:

```
Test run with 35 tests in 3 suites passed
Suite "EditorContentWidth" passed
Suite "SidebarWidthLimit" passed
Suite "EditorLayoutMode" passed
** TEST SUCCEEDED **
```

Specific claims I re-derived rather than assumed:

- **"Never narrower than before at any pane width."** Holds. `max(pane × fraction, min(pane, base))` — at pane 1000, normal gives `max(700, 720) = 720`, exactly the old flat value; without the second term it would have been 700. At pane 500 it gives 500, which is what the old `.frame(maxWidth: 720)` produced in a 500pt pane. Monotonic because both arguments to `max` are non-decreasing in `pane`, and `min` with constants preserves that.
- **`maxContentWidth` changed meaning** from *the* width (720/1100) to a *ceiling* (900/1400). That is the kind of rename that silently changes behaviour at a call site nobody revisited, so I grepped the tree: exactly one consumer remains (`EditorContentWidth.resolved`), and `normalContentWidth`/`wideContentWidth` have no stragglers. Clean.
- **⌘4 has no collision** anywhere in `Logue/`, and the View menu builds from `1 ... allCases.count` through `init(shortcutNumber:)`, so the fourth case is bound automatically as the description says. `everyLayoutHasAShortcut` compares `Set(reachable) == Set(allCases)`, so an unbound future case fails rather than going quiet.
- **No layout loop** in `measuringWorkspaceWidth()`. The published width is the workspace container's, which the window decides; the inspector reads it but cannot influence it. Applying it to the workspace `HStack` rather than the `NavigationSplitView` avoids the state-write-during-first-layout hazard the comment describes.
- **`everyPaneCombinationExists`** asserting `combinations.count == 4` is the test that actually pins the fourth case down — it would fail if `showsList`/`showsInspector` were wrong for `editorAndInspector`. Good choice.

The `guard listIsVisible != current.showsList` echo rule with `current` read from `EditorLayoutMode.stored` rather than the cached `@AppStorage` is the right fix, and the doc comment explaining *why* the previous asymmetric version existed is worth more than the code.

### Not blocking, but worth a follow-up commit

Three pre-existing tests were not revisited when the fourth case landed — see the inline comment. They still pass, because they enumerate cases by hand instead of iterating `allCases`, which is exactly why the 973-test run did not surface them. The new behaviour is genuinely covered elsewhere, so this is naming hygiene rather than a coverage hole.

Happy for these to land in a follow-up rather than hold the fix.

LogueTests/DocumentIconTests.swift
westerosweb: All three fixed in cf2e837, and you were right that the hand-enumeration is the reason the suite stayed green — that's what I changed, not just the names.

**`inspectorVisibility` / `listVisibility`** now derive the expectation from the case rather than listing which layouts qualify:

```swift
@Test("A layout shows the inspector exactly when its name says so", arguments: EditorLayoutMode.allCases)
func inspectorVisibility(mode: EditorLayoutMode) {
    let expected = mode == .allPanels || mode == .editorAndInspector
    #expect(mode.showsInspector == expected)
}
```

A fifth case now enters both tests automatically and fails until the expectation names it.

**`shortcutMapping`** is parameterised over pinned pairs, so the swap you describe fails:

```swift
arguments: [(1, .editorOnly), (2, .editorAndList), (3, .allPanels), (4, .editorAndInspector)]
```

Since a pinned list can itself go stale, I added `shortcutPairingsCoverEveryLayout` asserting `allCases.count == 4` — a fifth case fails there rather than silently going unpinned.

**File split** done: the layout suite moved to `EditorLayoutModeTests.swift`. `DocumentIconTests.swift` is back to 55 lines of icon tests.

Full suite 973 → 974, and the four affected suites verified in isolation:

```
✔ Suite "DocumentIcon" passed
✔ Suite "EditorContentWidth" passed
✔ Suite "EditorLayoutMode" passed
✔ Suite "SidebarWidthLimit" passed
✔ Test run with 43 tests in 4 suites passed
```

Logue/Views/Components/UnifiedSidebarView.swift
westerosweb: Correct on both counts — the comment was the thing that was wrong, not the code. Reworded in cf2e837 to say which path preserves the chosen width and which replaces it:

> Held separate from `currentWidth` so a *window resize* does not overwrite the width the user chose: shrinking narrows the panel on screen, and growing the window back returns it to the chosen width. A *drag* does overwrite it — `onChanged` stores the clamped value and `onEnded` persists it — which is intended, because a drag re-chooses the width from wherever the handle actually is.

That also makes it clear `chosenWidthSurvivesAShrink` covers the resize path only, which is what it actually tests.

On the `.onAppear` restore: real today, but not reachable. `panelWidths` is separate `@State` in `DocumentWorkspaceView` and `MeetingWorkspaceView`, and `contentArea` renders one or the other, never both — so no two `UnifiedSidebarView`s share a binding in a window. I've left it rather than add a `.onChange` for a case that can't currently arise; it becomes a real bug the moment a split view puts a document and a meeting side by side, which is worth remembering if that lands.

Logue/Models/EditorContentWidth.swift
westerosweb: Thanks — and your crossover figures are the useful part, because they're the thing the code doesn't state. `base / fraction` gives ~1029pt for normal and ~1222pt for wide, which is exactly where each column starts growing. No code change here; nothing to fix.

Worth adding that the boundary is also where the two terms are equal, so the piecewise function is continuous there — that's why `growthIsMonotonic` finds no step when it sweeps 200 → 3000 in 20pt increments.

On the negative-pane guard: it came from a real trap during a sidebar animation, not from caution. `EditorContentWidth` is `CoreGraphics`-only and has no SwiftUI dependency, so `degeneratePaneIsClamped` pins the contract at the type rather than relying on a view test to catch it.

PR #53, opened 2026-08-09
ci: pin every runner to macOS 26 instead of tracking macos-latest

shanforge:
Lint ran on `macos-latest` while Build and the release job pinned `macos-15`. That alias is
repointed to a new major macOS on GitHub's own schedule, so the jobs drift onto different
runners without anything in this repo changing — and the first sign of it would be a lint
failure nobody could reproduce locally.

**Revised after review** ([#discussion_r3763831165](https://github.com/bitwize-ai/Logue/pull/53#discussion_r3763831165)): the drift was not hypothetical, and this PR now pins *forward* rather than back.

## The split already existed

On the last `main` run before this branch ([31033559846](https://github.com/bitwize-ai/Logue/actions/runs/31033559846)), the Lint job log reads `Image: macos-26-arm64` while Build ran macOS 15. The alias had already moved; the two jobs were already on different major OSes.

The first version of this PR would have closed that gap by moving Lint *backwards* onto
macOS 15 — on a project whose `project.yml` sets `deploymentTarget.macOS: "26.0"` and
`xcodeVersion: "26.0"`, and with nothing scheduled to catch macOS 15's retirement. Same
determinism, but onto N-1.

So all three pins now say `macos-26`: the OS the app targets, and the image Lint has been
running on in practice.

## All three, not two

There are three `runs-on` pins, and the comment now names each of them so a future bump
moves them together:

| Job | File |
| --- | --- |
| Lint Check | `.github/workflows/ci.yml` |
| Build | `.github/workflows/ci.yml` |
| Build, Sign, Notarize, Release | `.github/workflows/release.yml` |

Leaving the signing/notarizing release build on a stale image would have reintroduced
exactly the split this change exists to remove.

## No cost change

`macos-26` is a standard GitHub-hosted runner — [the runner-images table](https://github.com/actions/runner-images#available-images) lists `macos-latest`, `macos-26` and `macos-26-xlarge` against macOS 26 Arm64, and only the `-large`/`-xlarge` variants are the paid ones. Per [GitHub's docs](https://docs.github.com/en/actions/reference/runners/github-hosted-runners):

> Use of the standard GitHub-hosted runners is free and unlimited on public repositories.

This repo is public, and the timing API confirms it directly for this image: run
31033559846 reports `"MACOS": {"total_ms": 0}`, including the Lint job that ran on
`macos-26-arm64`.

## Xcode on the new image

Both build jobs already select the newest installed Xcode rather than the image default
(the default was too old for the macOS 26 SDK and did not support `xcodebuild
-downloadComponent`). `macos-26-arm64` ships Xcode 26.0.1 through 26.6 with no Xcode 27
preview installed, so that step resolves to 26.6 — newer than the 26.3 the jobs were
picking up on macOS 15.

## Verification

- YAML parses; all three jobs resolve to `macos-26`.
- CI runs against this PR, which exercises the changed runner directly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,16 @@ permissions:
 jobs:
   lint-check:
     name: Lint Check
-    runs-on: macos-latest
+    # Pinned for the same reason the tool versions below are: `macos-latest` is an alias
+    # GitHub repoints to a new major macOS on its own schedule, which would move this job
+    # off the runner the Build job uses without anything in the repo changing. It already
+    # had: the alias now serves macOS 26 while Build was pinned to 15.
+    #
+    # Pinned forward to 26 rather than back to 15, because 26 is what the app targets
+    # (`project.yml`: deploymentTarget 26.0, xcodeVersion 26.0) and what this job has
+    # been running on in practice. Bump all three `runs-on` pins together and
+    # deliberately: this job, Build below, and release.yml.
+    runs-on: macos-26
 
     steps:
       - name: Checkout Repository
@@ -73,7 +82,8 @@ jobs:
 
   build:
     name: Build
-    runs-on: macos-15
+    # One of the three pins — see the Lint Check job above.
+    runs-on: macos-26
     timeout-minutes: 30
 
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,9 +52,10 @@ env:
 jobs:
   build-and-release:
     name: Build, Sign, Notarize, Release
-    # GitHub-hosted runner (matches CI). Swap to a self-hosted runner if your
-    # required Xcode/SDK isn't available on the hosted image.
-    runs-on: macos-15
+    # GitHub-hosted runner. The third of the three pins — see the Lint Check job in
+    # ci.yml for why they are pinned and why to 26. Swap to a self-hosted runner if
+    # your required Xcode/SDK isn't available on the hosted image.
+    runs-on: macos-26
     timeout-minutes: 90
 
     steps:


westerosweb (commented):
Two notes on the runner pin — both about the reasoning rather than the YAML, which is green on this SHA.

.github/workflows/ci.yml
```diff
@@ -14,7 +14,11 @@ permissions:
 jobs:
   lint-check:
     name: Lint Check
-    runs-on: macos-latest
+    # Pinned for the same reason the tool versions below are: `macos-latest` is an alias
+    # GitHub repoints to a new major macOS on its own schedule, which would move this job
+    # off the runner the Build job uses without anything in the repo changing. Bump both
+    # together and deliberately.
+    runs-on: macos-15
```
westerosweb: The premise here is inverted: the drift isn't hypothetical, it already happened.

On the last `main` run before this PR (run 31033559846), the Lint Check job log reads `Image: macos-26-arm64` while Build ran `macos-15`. `macos-latest` had already been repointed to macOS 26, so the two jobs were already on different major OSes.

That changes what this PR does. It isn't preventing a future split — it's reverting an existing one by moving Lint *backwards* from macOS 26 to macOS 15, on a project whose `project.yml` sets `deploymentTarget.macOS: "26.0"` and `xcodeVersion: "26.0"`.

Worth making that a deliberate call rather than an unnoticed side effect. Pinning all three jobs to `macos-26` gets the same determinism, but pinned *forward* onto the OS the app actually targets and the image the alias already serves.

`macos-15` does work today — the image carries Xcode 26.3, which is why Build is green. But it's now N-1, and there's no dependabot config or scheduled bump anywhere to catch its retirement, so the change trades silent drift for eventual silent breakage. Either direction is defensible; it just shouldn't be accidental.

.github/workflows/ci.yml
```diff
@@ -14,7 +14,11 @@ permissions:
 jobs:
   lint-check:
     name: Lint Check
-    runs-on: macos-latest
+    # Pinned for the same reason the tool versions below are: `macos-latest` is an alias
+    # GitHub repoints to a new major macOS on its own schedule, which would move this job
+    # off the runner the Build job uses without anything in the repo changing. Bump both
+    # together and deliberately.
```
westerosweb: There are three `runs-on` pins, not two: this job (`ci.yml:21`), the Build job below (`ci.yml:80`), and `release.yml:57`.

"Bump both together" names only the Build job in this file. Followed literally, someone bumps two of the three and leaves the signing/notarizing release build on a stale image — reintroducing exactly the split this change exists to remove. ("both" also reads ambiguously against the preceding sentence, which is about the tool versions.)

```suggestion
    # off the runner the Build job uses without anything in the repo changing. Bump all
    # three `runs-on` pins together and deliberately: this job, Build below, release.yml.
```

.github/workflows/ci.yml
shanforge: Verified and accepted — you are right, and the PR was arguing from a false premise.

Confirmed the drift on run 31033559846: the Lint job (92399940153) carries label `macos-latest` and its log reads `Image: macos-26-arm64`, while Build carries `macos-15`. The alias had already moved.

So this pins forward. All three `runs-on` values are now `macos-26`, which is the image Lint was already getting and the OS `project.yml` targets (deploymentTarget 26.0, xcodeVersion 26.0). Pinning to 15 would have moved that job onto N-1 with nothing scheduled to catch its retirement, which is the trade you flagged.

Two things I checked before moving Build and the release job forward with it:

- **Xcode.** Both jobs run `ls -d /Applications/Xcode*.app | sort -V | tail -1` rather than trusting the image default, so the concern would be a beta sorting last. `macos-26-arm64` ships 26.0.1 → 26.6 and no Xcode 27 preview, so it resolves to 26.6 — newer than the 26.3 they were getting on macOS 15.
- **Cost.** `macos-26` is the standard runner (only `-large`/`-xlarge` are the paid variants), and the timing API for that same run reports `"MACOS": {"total_ms": 0}` — including the Lint job that ran on this exact image. So the no-cost claim now rests on evidence from the image being adopted rather than the one being left behind.

Commit message and PR body rewritten so neither still claims the drift was hypothetical.

.github/workflows/ci.yml
shanforge: Correct on both counts — three pins, and "bump both" named only one of the other two. Taken, with the wording adjusted since the pin moved to `macos-26`:

```yaml
    # Pinned for the same reason the tool versions below are: `macos-latest` is an alias
    # GitHub repoints to a new major macOS on its own schedule, which would move this job
    # off the runner the Build job uses without anything in the repo changing. It already
    # had: the alias now serves macOS 26 while Build was pinned to 15.
    #
    # Pinned forward to 26 rather than back to 15, because 26 is what the app targets
    # (`project.yml`: deploymentTarget 26.0, xcodeVersion 26.0) and what this job has
    # been running on in practice. Bump all three `runs-on` pins together and
    # deliberately: this job, Build below, and release.yml.
```

The other two now carry a back-reference so the count is discoverable from either end rather than only from here — `# One of the three pins — see the Lint Check job above.` on Build, and on `release.yml` the existing comment gained `The third of the three pins — see the Lint Check job in ci.yml`.

Your point about the release job was the operative one: leaving the signing/notarizing build on a stale image while moving the other two would have rebuilt the exact split this change exists to remove, and it is the one job whose breakage would surface at release time rather than on a PR.

PR #57, opened 2026-08-12
Tasks: a first-class task list, action-item triage, and library panels

westerosweb:
## Why

Logue extracts action items from meetings and then has nowhere to put them.

`ActionItem` is a field on `MeetingNote`, so it cannot exist without a meeting. No UI anywhere creates one — only the summariser writes them. It has no priority, no tags, no repetition. And it lives in the always-encrypted meeting store, so it is never a file the user can edit.

A user records a meeting, Logue correctly pulls out "Send the revised deck to Priya by Friday", and that sentence's entire future is a checkbox on one detail screen. **The app's best output has no destination.**

This adds the destination, the path from a meeting to it, and the surfaces that make both worth using.

## What this adds

**`TaskItem`** — a task the user owns, stored the way documents are: plain `.md` with frontmatter in `~/Logue/Tasks/` when markdown storage is on, encrypted JSON otherwise. It follows the *document* mode rather than carrying one of its own, because two independent switches would let someone end up with plaintext tasks beside encrypted documents.

**Natural-language capture** — `Send the deck tomorrow #launch !`, with a live preview of what was understood so the syntax teaches itself. A `+` in the toolbar (⌘⇧N) puts the cursor there, because a text field is not what people look for when they want to add something.

**Repetition** — completing a repeating task rewrites the same task with its due date advanced from the *old due date*, not from today, so a weekly task completed three days late does not drift.

**A Tasks surface** — capture, counted filter chips, search over title, tags and notes, and the app's list chrome.

**Editing** — double-click a title to rename in place; an inspector panel for due date, priority, tags, recurrence and notes. This closes a gap where capture was the only moment a task could be described: triage could already set a due date and add a tag, so the model could edit a task and its owner could not.

**Promotion** — a meeting action item becomes a real task, keeping a link back to where it was decided. A copy that reuses the action item's identifier, so re-promotion is a lookup rather than a second bookkeeping field, and "Add all to Tasks" stays idempotent for the people who press it twice.

**Action Items becomes a triage inbox** — not a second dated to-do list. `ActionItem` gains `isDismissed`, so an extracted item can be rejected without claiming it was done; the chips become Inbox / Dismissed / All; and each row offers the two decisions an inbox is for. Promoted and dismissed items leave, so the inbox can actually reach empty.

**Triage** — the on-device model reviews open tasks and proposes changes. It never writes; the user applies one at a time.

**Library panels** — Action Items and Templates leave the left sidebar and become on-demand panels of All Meetings and All Documents, opened from a button in each surface's header. Neither earned a permanent row: an inbox whose resting state is empty, and a gallery you visit when starting a document, are lenses on a library rather than siblings of it.

Design docs: `docs/specs/2026-08-12-tasks-and-triage.md`, `docs/specs/2026-08-12-action-item-inbox-plan.md`, `docs/specs/2026-08-13-library-panels-design.md`.

## Deliberate decisions

**A new entity, not a document type.** A `WritingDocument` carries body, score, review grade, fact checks, PII findings, vocabulary suggestions and chat messages — a very heavy object for one sentence, and every task would appear in document lists, the link graph and search.

**A task and an action item stay separate.** Promotion is a copy, not a move. A meeting is a record of what was said; editing that record because the user later reprioritised would be rewriting history. The meeting instead shows, read-only, which of its action items became tasks — so the two relate without either owning the other.

**Tags, not "projects".** Logue already has two grouping concepts; a third would mean three ways to say the same thing. `#launch` becomes a tag.

**A trailing `!` means high priority.** A single `!` is high rather than a step on a scale, because a person typing it while capturing a task means urgent. Low priority is set from the UI — nobody reaches for punctuation to say something matters less.

**Editing routes through the same sanitiser as capture.** A title reaches both a filename and an LLM prompt, and a second, laxer entrance is how a sanitisation boundary quietly stops being one.

**Triage's safety gate is the parser, not the prompt.** `TriagePatch` is a struct with four optional fields rather than a dictionary, so "rewrite the title" is unrepresentable rather than something a validator must remember to reject. A suggestion is dropped unless its task was in the batch and still open, its kind is known, and its one patch field matches that kind. Past dates refused; tags outside the tag charset refused; a status other than `done` refused, so triage cannot reopen finished work; a `duplicate` never carries a patch, because deciding which of two tasks dies is the user's call.

**Templates keeps its grid.** The gallery uses adaptive 200–280pt columns over 55 templates, which collapses to one column in a panel. The panel is a purpose-built compact list, with the full grid one click away behind *Browse all* rather than deleted.

**Library panels are not held to the editor's reading measure.** `SidebarWidthLimit.inspector` reserves 720pt for the pane beside it, which is right for a document and wrong for a list — it capped these panels barely above their own default width. A list stays usable at 400, so the panel may take the rest.

## Six bugs worth calling out

Each was found by tracing or by running the app, and each is covered by a regression test where the behaviour is testable.

1. **Task files would have been adopted as documents.** `importAll` walks `snapshot.files` directly, so filtering `FolderSnapshot` was not enough. A task file carries `_logue_task_id`, not `_logue_id`, so it fell through to `unidentifiedFiles` — which the scan *adopts*, stamping document frontmatter onto it once it settles. Every task in the library would have become a document about two seconds after being written.

2. **The folder name collided with a real space.** A library in use already had `~/Logue/Tasks` as a space holding about a hundred documents. Writing our marker there would have made the space read as *vanished*, which routes to `trashDocuments(inSpace:)`. Now `_space.md` always wins, `prepare()` refuses rather than sharing, and the folder name steps over an occupied one.

3. **`/var` is a symlink to `/private/var`.** `appendingPathComponent` and `contentsOfDirectory` spell the same file differently, so a filename-changed check read same-file as different-file and trashed the file it had just written — a task vanished on every re-save.

4. **Prompt injection through tags.** Titles are JSON-encoded and Foundation *happens* to escape `/`, so `</tasks>` came out escaped — an implementation detail, not a guarantee. Tags are joined as plain text and `allTags` reads from hand-editable files. Both now strip angle brackets.

5. **The panel's resize handle could not be grabbed.** A 1pt rectangle with the 8pt grab area in an `.overlay` — and hit-testing does not reach outside a view's own bounds, so those points were never live. The handle is now an 8pt view with the hairline drawn inside it. `UnifiedSidebarView` had the same construction, so it now uses `ResizableEdge` too — which means this branch changes the drag handle on the two surfaces that sidebar serves, `MeetingWorkspaceView` and `DocumentWorkspaceView`. Both need a click-through: the handle should be grabbable across an 8pt band rather than a 1pt hairline, the hairline should still be 1pt, and a width dragged there should persist across a restart.

6. **The panel clipped its own contents.** `.fixedSize()` on the filter menus sized the controls row to the longest meeting title, which overflowed the panel frame and pushed the filter chips out of view.

## Not stranding anyone

`SidebarSelectionMigration` maps the two retired sidebar values to their host surface with the panel open — without it, anyone whose app last sat on Action Items or Templates would launch into nothing. Both keep their command-palette entries: with the rows gone, the palette is the discovery path.

## Verification

- **1347 tests in 108 suites, zero failures** (`scripts/test-no-llm.sh`), run in CI as a step of
  the Build job at every head on this branch.
- Lint clean under SwiftFormat + SwiftLint `--strict`.
- Signed build succeeds and runs; `main` merged in with no behavioural conflicts.
- The folder rules are executed by their tests, not merely accompanied by them: reverting the
  resolve-trash-forget order in `TaskFolderLocator.retire` fails three of the eight cases in
  `TaskFolderLocatorTests`. Checked by performing the revert. Four earlier rounds of review each
  found a defect in that logic while it lived in statics over `DocumentStorage.markdownRootURL`
  and `UserDefaults.standard`, where the revert left every suite green.

`scripts/test-no-llm.sh` is new because `-skip-testing:` takes a class name, never a directory — `-skip-testing:LogueTests/LLMIntegration` is silently ignored and the model-backed suites run anyway.

**Not verified by hand:** the two library panels and the task inspector have not been observed rendering under automation. Logue cannot be an Accessibility client, so a panel that opens from a toolbar button is out of reach; they were exercised by eye during development.

**Deliberately out of scope:** a CLI (belongs with the MCP server), a daily journal note, adopting hand-created task files, reminders for tasks, filing a task into a space, and live external-edit pickup for task files — they load on launch, unlike documents.


shanforge (changes requested):
This adds a real task entity with its own storage, natural-language capture, triage, promotion from action items, and moves Action Items and Templates into on-demand library panels. The model layer and the pure rules are careful work and well tested. The storage layer is not finished: the tasks folder is located by recomputing a name rather than by the identity marker it writes, which re-opens the two bugs the PR body says it fixed, and the inspector writes the previously-selected task's text onto the newly-selected one on a plain click.

## Blocking

**`Logue/Views/Tasks/TaskInspectorPanel.swift:60`** — selecting a different task in the list writes the previous task's title and notes onto the newly-selected task.

`.onChange(of: task.id) { commitDrafts(); loadDrafts() }` runs the commit *before* the drafts are refreshed, and the zero-parameter `onChange` closure is re-created on the update that already carries the new value — so inside it `task` is B while `titleDraft`/`notesDraft` still hold A's text. `commitDrafts()` (line 279) then evaluates `sanitised != task.title` as A's title vs B's title (true), builds `updated` from B, and calls `onChange(updated)` → `store.update(...)` → `TaskStorage.save`. Click task A, then click task B: B is now titled with A's title and carries A's notes. In markdown mode `TaskFolderStore.save` also renames B's file to A's title and trashes B's original file. No edit is required for this — merely selecting two tasks in turn is enough, because `loadDrafts` ran on the first selection. `TaskListView` renders the panel from a stable `if let selectedTask` branch, so the `@State` survives the swap and the handler does fire.

Fix: capture the drafts against the task they were loaded from — either commit in `.onChange(of: task.id) { old, new in }` against the *old* id, or hold `draftTaskID` alongside the drafts and refuse to commit when it does not match `task.id`.

**`Logue/Services/MarkdownStorageMigrator+SpaceFolders.swift:89`** — creating a space named "Tasks" writes `_space.md` into the live tasks folder, which re-opens bug 1 from the PR description.

`createFolder(for:in:)` derives the path from the space name (`SpaceFolderLayout` disambiguates only against sibling *spaces*), `createDirectory(withIntermediateDirectories: true)` succeeds silently on the existing `~/Logue/Tasks`, and `writeIdentity` drops `_space.md` in it. Nothing on the space-creation side consults `TaskFile.folderName` or the marker. From that moment: `FolderSnapshot.taskFolders` (line 88) excludes the folder because space identity deliberately wins, so every task `.md` — and `_tasks.md` itself — becomes a `documentFile`; `importAll` finds no `_logue_id` in them so they go to `unidentifiedFiles`; `MarkdownFolderScan.plan` adopts settled unidentified files, and `MarkdownStorageMigrator.adopt` writes `MarkdownDocumentFile.render(content)` back over each file. `DroppedFileImport` uses `purpose: .readingAFileWeAlreadyHold`, whose `refusesUnreasonableFiles` is false, so even a notes-free task file is adopted. `status`, `priority`, `due`, `recur`, `completed_count` and the task identity are gone from the only copy on disk, and the marker is overwritten so the folder can never be recognised again. The watcher fires on writes to `~/Logue`, so no further user action is needed. `TaskFolderIsolationTests.existingSpaceWinsOverTaskMarker` covers the opposite ordering only; `TaskFolderStore.prepare()` also only refuses when the space got there first.

Fix: refuse in `createFolder`/`moveFolder` when the target directory already holds `TaskFile.folderMarkerFilename`, and disambiguate the space's directory component past it — the same "step over an occupied name" rule the tasks side already applies to spaces.

**`Logue/Services/TaskStorage.swift:35`** — the tasks folder is located by recomputing a path from a fixed name, never by the `_logue_tasks_folder` identity, so the task list silently empties whenever that name resolves somewhere else.

`tasksFolderURL` builds `root + TaskFile.folderName` (plus a numeric suffix) and only ever asks whether that path is already a space. `TaskFile.markerIdentifier(in:)` has exactly one caller in the tree — `FolderSnapshot.taskFolders` — so the marker is used for *exclusion* and never for *location*. Two reachable failures:

- The user renames `~/Logue/Tasks` to `~/Logue/My Tasks` in Finder, which the marker file they are reading explicitly invites: "Do not remove this file: it is how Logue recognises the folder, so that renaming the folder keeps working." `tasksFolderURL` still returns `~/Logue/Tasks`, `TaskFolderStore.exists` is false, `loadAll()` returns `[]`, and every task disappears from the app. The renamed folder is still excluded from the document library by its marker, so the files are invisible everywhere in the app; the next save recreates an empty `~/Logue/Tasks`.
- With no Finder action at all: a user who already had a space named "Tasks" got tasks written to `~/Logue/Tasks 2`. They rename that space to anything else, `moveFolder` moves `~/Logue/Tasks` away, `~/Logue/Tasks` is now free, `tasksFolderURL` flips back to it, `loadAll()` finds nothing, and `Tasks 2` is orphaned.

This contradicts `TaskFile.swift:26-30` ("a folder is found by its identity, never by recomputing a path from its name. Renaming `Tasks/` in Finder must keep tasks working") and the CLAUDE.md invariant of the same name. `TaskFolderStoreTests.lookupSurvivesRename` renames an individual task *file*, not the folder, and `TaskStorage` has no test suite at all.

Fix: resolve the folder from the snapshot the way `SpaceFolderMap` resolves a space — first marked directory wins, name only as the creation-time default.

## Major

**`Logue/Services/TaskStorage.swift:180`** — `reEncryptFromFolder()` discards every write result and cannot tell an unreadable folder from an empty one, so switching out of markdown mode can trash the folder while tasks that never reached encrypted storage are shown as present.

`let fromFolder = folderStore.loadAll(); for task in fromFolder { writeEncrypted(task) }` — `writeEncrypted` is `@discardableResult` and its `Bool` is dropped; the merge below then re-adds `fromFolder` over the encrypted set regardless. If the writes fail (full disk, encryption error — `writeEncrypted` only logs) or `loadAll()` returns `[]` because the folder could not be listed, `switchToEncrypted` still sets `restoredTasks`, `PrivacySettingsTab` calls `taskStore.replaceAll(with: storage.restoredTasks)` (memory only), and `retireFolderAfterReEncryption` (DocumentStorage.swift:373) verifies `hasEncryptedCopy` for **documents only** — so `~/Logue` goes to the Trash with the tasks' only copy inside. Tasks created during markdown mode have no prior encrypted copy at all, so on the next launch `loadEncrypted()` cannot find them. The documents path guards both halves of this (`SwitchError.folderIncomplete`, `hasEncryptedCopy`); the tasks path guards neither.

Fix: count the successful `writeEncrypted` calls, refuse the switch when any failed or when `folderStore.exists` is true but `loadAll()` came back short, and extend `retireFolderAfterReEncryption` to check the tasks too. While you are there, `restoredTasks` (line 476) is a mutable side channel on the singleton read by exactly one caller — returning it makes the temporal coupling checkable.

**`Logue/Services/DocumentStorage.swift:484`** — "Clear all example data" trashes the whole `~/Logue` root, including `Tasks/`, and no reset path clears tasks in encrypted mode.

`DocumentStore.clearAllData()` (DocumentStore.swift:526) calls `clearMarkdownFolderContents()`, which calls `migrator.retireRoot()` — `retireFile(rootURL)` on `~/Logue`, taking `Tasks/` and every task file with it — then recreates an empty root. Reached from GeneralSettingsTab.swift:256, whose dialog says "This will remove all sample documents, meetings, and spaces", and from SeedDataBannerView.swift:84. Nothing clears or reloads `TaskStore`, so the list on screen still shows the tasks until relaunch. The mirror gap: `TroubleshootingActions` "clear everything" (lines 89-91) never touches `Application Support/Logue/tasks`, so in encrypted mode the tasks survive a full reset. Recoverable from the Trash in the first case, but only if the user notices.

Fix: give `TaskStore`/`TaskStorage` a `clearAllData()` and call it from both paths, or exclude `Tasks/` from `retireRoot` when it is not what the user asked to clear.

**`Logue/Services/MarkdownStorageMigrator.swift:504`** — `importAll` has a contents-level guard for space files but none for task files, and the helper written for exactly that is never called.

The loop over `snapshot.documentFiles` escapes on `SpaceFile.isSpaceFile(filename:) || SpaceFile.isSpaceFile(contents:)`, then hands anything without a `_logue_id` to `unidentifiedFiles` — which the scan adopts and rewrites. `TaskFile.isTaskFile(contents:)` exists at TaskFile.swift:152 with the doc comment "a second check, in case it was renamed" and has no callers anywhere. Any task file that reaches this loop — because the folder was colonised (blocking finding above), because `_tasks.md` was deleted or renamed, or because a user copied a task file up into `~/Logue` to look at it — has its frontmatter overwritten with `_logue_id` and loses its status, priority, due date, tags and recurrence.

Fix: `continue` on `TaskFile.isTaskFile(contents: contents)` in the same guard as the space check. It is one line and it makes the folder-level exclusion a belt rather than the only strap.

**`Logue/Services/TaskFolderStore.swift:145`** — `save(_:)` re-walks the folder three times and re-parses every file in it on every single write, and `exportAll` calls `save` in a loop.

`save` calls `prepare()`, then `url(forTaskID:)` (line 133), which loops `taskFileURLs()` doing `read(at:)` + `TaskFile.parse` for **every** `.md` in the folder, then `currentFilenames()`, which lists the directory again. `exportAll` (line 209) then runs that per task. Turning on plain-markdown storage with 300 tasks is on the order of 45,000 file reads and parses, synchronously on the main actor (`TaskStorage` is `@MainActor`). The same per-task cost is paid on a multi-line `TaskStore.capture`, which calls `add` → `save` per line.

Fix: build the id→URL map once (`loadAll()` already walks it) and pass it into `save`, or have `exportAll` walk once and write against that index.

## Minor

**`Logue/Services/TaskFolderStore.swift:177`** — `currentFilenames()` swallows the directory listing with `try?` and returns `[]`, against CLAUDE.md's "Never use silent `try?` for ANY operation. This includes file I/O". Every other read in the file uses `do/catch` with `Self.logger.error`. The empty set also means `DocumentFilename.filename(forTitle:id:avoiding:)` sees no collisions, so a listing failure that still permits a write lands a new task on an existing task's filename and `save` reports success.

**`Logue/Engine/TaskTriage.swift:206`** — the rule for what a tag may contain is written out four times: `$0.isLetter || $0.isNumber || $0 == "-" || $0 == "_"` at TaskTextParser.swift:108, TaskEdit.swift:26, TaskTriage.swift:208 and TaskTriage.swift:337. `maxTagLength = 32` is declared at TaskTextParser.swift:30 and again at TaskTriage.swift:84 (TaskEdit correctly references the parser's). Widening the charset in one place gives you a tag capture accepts, the inspector rejects, and the triage sanitiser silently strips — and the prompt-injection guard is the copy most likely to be missed.

**`Logue/Engine/TaskTriage.swift:154`** — `dayFormatter` is a shared `private static let` whose `timeZone` is mutated in place by both `userPrompt` (line 154) and `validFutureDate` (line 329). Both are non-isolated statics, so nothing stops two callers from mutating and reading it at once; `DateFormatter` is only safe to read concurrently, not to mutate. Production reaches it only from `@MainActor`, so this is latent rather than firing today. Build the formatter from the calendar's time zone locally, or make it a computed property.

**`Logue/Views/Library/LibraryPanelContainer.swift:47`** — the drag machinery is a verbatim copy of `UnifiedSidebarView.resizeHandle` (Logue/Views/Components/UnifiedSidebarView.swift:75-119): same `dragStartWidth`/`dragStartX`, same delta math, same `Transaction`, same `NSCursor` push/pop, same comment word for word. The PR body is explicit that only this copy got the hit-target fix. Extracting one `resizableEdge` modifier would let the other surfaces pick the fix up when you decide to give it to them, instead of guaranteeing the two drift.

**`Logue/Views/Sidebar/CategorySidebarView.swift:12`** — `@Environment(TemplateStore.self) private var templateStore` is dead after this diff removed the Templates row; it appears nowhere else in the file. Delete it rather than leave the view carrying a dependency it never reads.

**`scripts/test-no-llm.sh:18`** — seven of the 23 `-skip-testing:` entries name source files, not suites: `GrammarClarityLLMTests VocabularyLLMTests RewritePanelLLMTests ReviewPanelLLMTests MeetingLLMTests VerifyPanelLLMTests DocumentLLMTests` are the `.swift` filenames in `LogueTests/LLMIntegration/`; the 16 suite types they contain (`DocumentTitleLLMTests`, `DocumentChatLLMTests`, …) are already listed on the lines above. They skip nothing, and the list looks more complete than it is — add a suite to `DocumentLLMTests.swift` assuming "that file is already skipped" and you get a model-backed run in the fast suite.

**`Logue/Views/Library/LibrarySurfaceView.swift:9`** — the file declares no `LibrarySurfaceView`; its contents are `struct LibraryPanelToggle` and the `View` extension. `LibraryPanelContainer.swift` next door follows the one-type-per-file-named-after-it convention.

**`Logue/Models/TaskItem+ActionItem.swift:95`** — `merging` builds and discards an entire `TaskItem`, passing a freshly invented `UUID()` as the source meeting id, only to reach the `private` `carriedNotes` helper 60 lines above it in the same file. Making `carriedNotes` `fileprivate` and calling it directly removes both the fabricated identifier and the throwaway object.

## Comments to delete

- `Logue/Views/Tasks/TaskInspectorPanel.swift:21` — "Not resizable yet — the drag machinery lives in `UnifiedSidebarView`, which is built for tabbed multi-tool panels." This PR added `LibraryPanelContainer`, a tab-less single-panel resizable container carrying its own copy of that machinery, which is exactly what this panel would reuse. The stated reason is no longer true.
- `Logue/Views/Sidebar/CategorySidebarView.swift:259` — `// MARK: - Action Item Counts` now heads a section whose only member is `hasOverdueTasks`, which reads `taskStore.openTasks`. `pendingActionItemCount`, `hasOverdueItems` and `actionItemsAccessibilityLabel` were all removed from it by this diff.
- `Logue/Engine/TaskTriage.swift:204` — "Matching `validTag` on the way back keeps one definition of what a tag may contain." There are four definitions; the two in this file disagree with the comment on their face.

## Compatibility with main

Persisted-data compatibility is clean. `ActionItem.isDismissed` decodes with `decodeIfPresent ?? false` and the rest of the meeting decoder is unchanged, so meetings written by 1.0.x still load. `TaskItem` has a hand-written `init(from:)` with `decodeIfPresent` and a default on every non-identity field. No frontmatter key on documents or `_space.md` was renamed or removed, and the two retired sidebar values are mapped from string literals by `SidebarSelectionMigration` rather than from enum raw values, so an install that last sat on Action Items or Templates lands somewhere real. The security rules hold: triage sanitises titles and tags (angle brackets stripped), wraps them in `<tasks>`, truncates via `LLMEngine.maxInputChars`, routes through `LLMEngine.complete` with an explicit `maxTokens`, and its button is gated on `LLMEngineStatus.shared.isBusy`; no `url.absoluteString` logging, no `[0]` on FileManager arrays, no force unwraps or casts in the new code; entitlements untouched. Recording is not touched at all.

What is not safe to merge is the interaction between the new `Tasks/` folder and the existing markdown-folder machinery. Anyone with plain-markdown storage on — the users who already trust `~/Logue` with their library — can lose task data three ways that need no unusual action: create a space called "Tasks", rename the tasks folder the marker file invites them to rename, or press "Clear Examples". The inspector defect needs nothing but two clicks and does not require markdown mode at all. `TaskStorage` has no test suite, so `tasksFolderURL`, `reEncryptFromFolder`, `loadEncrypted` and `writeEncrypted` — the half that moves the user's data between modes — are entirely unexercised; that gap is where three of the four findings above live.

westerosweb:
All findings addressed in c7441d1. Each was verified against the code before changing anything; the three blocking traces were accurate.

## Blocking

**Inspector writing the previous task's text onto the newly-selected one** — confirmed exactly as described: the zero-parameter `onChange(of: task.id)` closure reads `task` as B while the drafts still hold A's, so `commitDrafts` built `updated` from B with A's title. Took the second of the two suggested fixes: the drafts now carry a `draftTaskID` set by `loadDrafts`, and `commitDrafts` refuses when it does not match `task.id`. The focus-loss commit still saves genuine edits before the selection changes; anything uncommitted at the moment of the swap is dropped rather than misapplied, which seemed the right trade against holding a second copy of the outgoing task that could go stale behind the discrete controls.

**Space named "Tasks" colonising the folder** — confirmed; `MarkdownStorageMigrator+SpaceFolders.swift` contained no reference to `TaskFile` at all. `createFolder` and `moveFolder` now route their destination through `avoidingTaskFolder`, which steps past any directory carrying `TaskFile.folderMarkerFilename` and picks the first free `<name> N`. Stepping over rather than refusing, so naming a space "Tasks" still works — the mirror of the rule the tasks side already applies. Only the colliding case changes; every other path is returned untouched.

**Folder located by name rather than identity** — confirmed, and it contradicted the invariant `prepare()`'s own docstring states. Added `TaskFolderStore.markedFolder(in:)`, which scans the root for a directory carrying the marker; `tasksFolderURL` consults it first and falls back to the name only as the creation-time default. Ambiguity keeps the folder actually named `Tasks` and logs rather than guessing. Three tests cover it, including the Finder-rename case the marker file invites.

## Major

- **`reEncryptFromFolder`** now returns a `ReEncryptResult` carrying the tasks, a count of failed writes, and whether the folder read back empty because it was unreadable (`TaskFolderStore.taskFileCount` returns `nil` when it cannot be listed, which is not the same as `0`). `retireFolderAfterReEncryption` checks tasks on the same terms as documents — `taskReEncryptionIsComplete` plus a per-task `hasEncryptedCopy` — so the folder cannot be trashed while a task's only copy is inside it.
- **Reset paths** — added `TaskStore.clearAllData()`, which empties the list, removes the encrypted copies and trashes the folder, and wired it into both `GeneralSettingsTab` and `TroubleshootingActions`. That closes both halves: the markdown path no longer leaves the list on screen, and the encrypted path no longer leaves tasks behind entirely.
- **`importAll`** — `continue`s on `TaskFile.isTaskFile(contents:)` in the same guard as the space check. That helper had no callers anywhere, as noted.
- **`save` / `exportAll`** — added `FolderIndex` and `write(_:index:)`. The folder is walked once per batch rather than three times per task, and `exportAll` threads one index through. `url(forTaskID:)` remains for single lookups; `currentFilenames()` is gone.

## Minor

- Tag charset unified behind `TaskTextParser.isTagCharacter` (was four copies), and the duplicate `maxTagLength` in `TaskTriage` removed.
- `dayFormatter` is built per call from the calendar instead of having its `timeZone` mutated in place by two non-isolated statics.
- Extracted `ResizableEdge`, now used by both `LibraryPanelContainer` and `UnifiedSidebarView` — so the hit-target fix is no longer confined to one copy, and the workspace inspectors pick it up.
- `currentFilenames()` no longer swallowed a listing failure as an empty set; that path is gone entirely, and `makeIndex()`/`taskFileCount` return `nil` on failure so callers cannot read it as "no files".
- Dead `@Environment(TemplateStore.self)` removed from `CategorySidebarView`.
- `scripts/test-no-llm.sh` trimmed to the 16 real suite types; verified each remaining entry is a declared `struct` in `LogueTests/`.
- `LibrarySurfaceView.swift` renamed to `LibraryPanelToggle.swift`.
- `merging` calls `carriedNotes` directly (now `fileprivate`), so no throwaway `TaskItem` and no fabricated meeting identifier.

All three stale comments deleted, and the `// MARK: - Action Item Counts` heading renamed to match what is left under it.

## Verification

Build succeeds; **1291 tests in 102 suites pass** (four new, at the `TaskFolderStore` layer); SwiftLint `--strict` clean; SwiftFormat clean at `0/484 files require formatting`.

## Not done

`TaskStorage` still has no test suite. `tasksFolderURL`, `loadEncrypted` and `writeEncrypted` read singletons and the real Application Support directory, so covering them needs an injectable root that does not exist yet — a change worth making deliberately rather than folded into this PR. The new coverage sits one layer down in `TaskFolderStore`, which is where the folder-identity logic now lives. Happy to do the injection as a follow-up if you would rather have it before merge.

Logue/Services/MarkdownStorageMigrator+SpaceFolders.swift
```diff
@@ -87,11 +87,41 @@ extension MarkdownStorageMigrator {
...
+    /// `createDirectory(withIntermediateDirectories: true)` succeeds silently on a directory
+    /// that is already there — so `_space.md` would land inside it. From that moment the
+    /// snapshot resolves the folder in the space's favour, every task file reads as an
+    /// unidentified document, and the scan adopts and rewrites them: status, priority, due
+    /// date and the task's identity gone from the only copy on disk.
+    ///
+    /// Steps over rather than refusing, so naming a space "Tasks" still works — the same rule
+    /// the tasks side already applies when a space got to the name first. Only the colliding
+    /// case behaves differently; every other path is returned untouched.
+    func avoidingTaskFolder(_ directory: URL) -> URL {
```
shanforge: **Major** — stepping the folder aside makes its name diverge from the space's name, and nothing else in the folder machinery knows that. Two confirmed consequences.

(1) With `~/Logue/Tasks` marked, creating a space named "Tasks" lands it at `Tasks 2`. The watcher is created with `kFSEventStreamCreateFlagIgnoreSelf`, so our own `_space.md` write raises no event; at the next scan (app activation) `SpaceFolderAdoption.renames` compares the claimed folder name against the space name, they differ, and `adoptSpaceRename` renames the user's space to "Tasks 2". Renaming it back runs `moveFolder` → here → `Tasks 3` (because `Tasks 2` is the source and so not free), physically relocating the folder and every document in it, and the next scan renames the space "Tasks 3". The user can never keep a space named "Tasks", and each attempt moves files on disk.

(2) In the window before that first scan the space is still named "Tasks" while its folder is `Tasks 2`. Creating a sub-space in it takes the fallback in `SpaceFolderMap.components` — the child has no claim yet — which derives the whole path from *ancestor names*: `Tasks/Q3`. Only the leaf is inspected here, and `Q3` carries no marker, so `createDirectory(withIntermediateDirectories: true)` creates it inside the live tasks folder and `_space.md` lands there. `FolderSnapshot.spaceFiles` and `documentFiles` exclude everything under a task folder, so that space and every document filed into it drop out of every scan permanently — logged as `unwalkable` and left alone. `moveFolder` has the same hole through its name-derived destination.

Fix: derive a child's path from its parent's *claimed* components rather than the ancestor name chain, and apply this check to every component of the destination rather than the leaf — or keep the space's name in step with the folder it was actually given at the moment it is given.

Logue/Services/TaskStorage.swift
```diff
@@ -0,0 +1,272 @@
...
+    func reEncryptFromFolder() -> ReEncryptResult {
+        let fromFolder = folderStore.loadAll()
+        var failed = 0
+        for task in fromFolder where !writeEncrypted(task) {
+            failed += 1
+        }
+        // An empty read from a folder that exists and holds task files is a listing failure,
+        // not an empty task list. Told apart here because afterwards the folder is gone.
+        // `nil` (unlistable) counts as unreadable; only a confirmed zero means truly empty.
+        let unreadable = fromFolder.isEmpty && folderStore.exists && (folderStore.taskFileCount ?? 1) != 0
```
shanforge: **Major** — this tells an unreadable folder from an empty one, but not from a *partial* one, which is what the previous review asked for ("refuse when `loadAll()` came back short").

Ten task files, three of which fail `read(at:)` or `TaskFile.parse` — hand-edited frontmatter, a permissions error. `loadAll()` compactMaps them away and returns seven; all seven encrypt, so `failed == 0`; the result is non-empty, so `unreadable` is false; `isComplete` is true, `retireFolderAfterReEncryption` passes, and `retireRoot()` trashes `~/Logue` with those three files inside. Tasks created during markdown mode have no encrypted copy at all, so they exist only in the Trash. `taskFileCount` is already being read on this line and would have answered ten against seven.

The converse also fires: zero tasks and one non-task `.md` in the folder — a state `nonTaskFileIgnored` declares supported — makes `taskFileCount` 1, so the switch reports a failure and refuses to retire the folder.

Fix: compare `fromFolder.count` against a count of files that actually parse as tasks, treating `nil` as unreadable, instead of testing for emptiness.

Logue/Services/DocumentStorage.swift
```diff
@@ -364,6 +376,14 @@ final class DocumentStorage {
             throw SwitchError.reEncryptionIncomplete(count: missing.count)
         }
 
+        // Tasks are checked on the same terms and for the same reason. Without this the
+        // documents half is guarded and the tasks half is not, so a failed task write sends
+        // the folder — holding the only copy of those tasks — to the Trash anyway.
+        let missingTasks = restoredTasks.filter { !TaskStorage.shared.hasEncryptedCopy(of: $0.id) }
+        guard taskReEncryptionIsComplete, missingTasks.isEmpty else {
+            throw SwitchError.reEncryptionIncomplete(count: max(missingTasks.count, 1))
```
shanforge: **Minor** — a task-side failure is reported to the user as a document failure with a count nobody measured. When `taskReEncryptionIsComplete` is false and `missingTasks` is empty, this throws with `count: 1` and `PrivacySettingsTab` shows "1 document(s) had not finished writing to encrypted storage… Your documents are safe in both places — try turning the setting off again" — no document failed, and the setting is already off, since `mode` flipped inside `switchToEncrypted` before this runs. The user is pointed at a control that no longer exists and never told it is their tasks that need attention.

Fix: give `SwitchError` a task-shaped case carrying the real count (or zero-and-unreadable), worded for tasks.

Logue/Views/Components/ResizableEdge.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    private static let hitWidth: CGFloat = 8
+
+    @State private var dragStartWidth: CGFloat?
+    @State private var dragStartX: CGFloat?
+
+    var body: some View {
+        Rectangle()
+            .fill(Color.clear)
+            .frame(width: Self.hitWidth)
```
shanforge, line 39: **Minor** — the extraction changes the layout of a surface that predates this PR. On `main` the workspace inspector's handle is a 1pt rectangle with the 8pt grab area in an `.overlay`, and an overlay does not expand its parent, so it occupies exactly 1pt in the enclosing `HStack` with the hairline flush against the panel. Here the 8pt hit area *is* the view, so opening any inspector in the document or meeting workspace now costs the editor 7 extra points and the hairline sits about 3.5pt off the panel edge. Picking up the hit-target fix is the point of the extraction, but this is a visible change to two shipped surfaces that the PR body does not mention — worth a before/after screenshot, or keep the handle 1pt wide and give it `.contentShape` padding instead.

Logue/Views/Tasks/TaskInspectorPanel.swift
```diff
@@ -0,0 +1,338 @@
...
+                .padding(16)
+            }
+        }
+        .frame(width: panelWidth)
+        .background(AppThemeConstants.surfaceBackground)
+        .onAppear(perform: loadDrafts)
+        // Switching rows must not carry the previous row's half-typed text across. The
+        // commit is guarded by `draftTaskID`, so anything not already saved by the
+        // focus-loss commit below is dropped rather than written onto the new task.
+        .onChange(of: task.id) {
```
shanforge, line 69: **Minor** — the commit here can never write. `task` already holds the new value when this fires, and `draftTaskID` is only reassigned by `loadDrafts()` on the line below, so the guard in `commitDrafts()` always fails on this path. That is what closes the blocking defect, but it leaves a call that reads as a save and is not one — and if SwiftUI delivers the selection change before the focus change (both come from the same mouse-down and the order is not guaranteed), a half-typed title is dropped rather than saved, with no indication.

Fix: either delete the call and let the comment say plainly that unsaved text is dropped, or use `.onChange(of: task.id) { old, _ in }` and commit the drafts against `old`.

Logue/Services/TaskFolderStore.swift
```diff
@@ -0,0 +1,376 @@
...
+    ///
+    /// The marker file tells the user that renaming the folder is safe — "it is how Logue
+    /// recognises the folder, so that renaming the folder keeps working" — so the lookup has
+    /// to honour that. Resolving by name instead means a folder renamed in Finder reads as
+    /// missing, every task vanishes from the app, and the next save quietly recreates an empty
+    /// one beside the real files, which that same marker keeps out of the document library.
+    ///
+    /// Ambiguity is not resolved by guessing: two marked folders keep the one actually named
+    /// `TaskFile.folderName` if either is, and the situation is logged.
+    static func markedFolder(in root: URL) -> URL? {
```
shanforge: **Minor** — this searches only the root's immediate children, while `FolderSnapshot.taskFolders` recognises the marker at any depth. Drag `~/Logue/Tasks` into `~/Logue/Work` in Finder: this returns `nil`, `tasksFolderURL` falls back to `~/Logue/Tasks`, which no longer exists, `loadAll()` returns `[]` and the task list empties — while the snapshot still finds the marker at `Work/Tasks` and keeps those files out of the document library too. The tasks are visible in Finder and reachable from nowhere in the app, and the next save recreates an empty folder beside them. Same shape as the rename bug this fix closes, one Finder gesture over.

Fix: enumerate recursively (bounded by the depth the snapshot already uses) so a moved folder is found the way a renamed one is.

shanforge (changes requested):
All three blocking findings are closed, and closed at the cause rather than the symptom: the inspector refuses a commit whose drafts belong to another task, the tasks folder is now located by its marker with the name demoted to a creation-time default, and space-folder creation *and* the previously unflagged rename path both step over a marked tasks folder. Sixteen of the eighteen prior findings are genuinely fixed. Two are half-fixed, and the step-over rule introduced to close blocking finding 2 has a consequence of its own: it puts a space's folder under a name that nothing else in the folder machinery derives.

## Fixes verified

- **Blocking — `TaskInspectorPanel.swift:60`** cross-task write — **fixed**. `draftTaskID` is set in `loadDrafts` and `commitDrafts` refuses when it does not match `task.id`, so the A→B click can no longer build `updated` from B. (See the minor note on the now-dead call inside the `task.id` handler.)
- **Blocking — `MarkdownStorageMigrator+SpaceFolders.swift:89`** `_space.md` written into the live tasks folder — **fixed** for the reported path, and for the rename variant the last review did not name. Both `createFolder` and `moveFolder` route their destination through `avoidingTaskFolder`. (See Major 1 for what the step-over costs.)
- **Blocking — `TaskStorage.swift:35`** folder located by name — **fixed**. `TaskFolderStore.markedFolder(in:)` resolves by marker first, ambiguity is broken by rule rather than by guessing, and three of the four new tests fail at `db480a13` because the API did not exist.
- **Major — `TaskStorage.swift:180`** re-encryption discards write results — **partially fixed**. Failures are now counted and `retireFolderAfterReEncryption` refuses on them, but the review asked for a refusal when `loadAll()` "came back short" and the new check only detects when it came back *empty*. See Major 3.
- **Major — `DocumentStorage.swift:484`** "clear all example data" — **partially fixed**. `TaskStore.clearAllData()` exists and is wired into `TroubleshootingActions.swift:92` and `GeneralSettingsTab.swift:259`; the third call site, named by file and line in the last review, was not. See Major 2.
- **Major — `MarkdownStorageMigrator.swift:504`** no task-file guard in `importAll` — **fixed**. The `continue` sits in the same run of guards as the space check and ahead of the `unidentifiedFiles` branch, so `TaskFile.isTaskFile(contents:)` finally has a caller.
- **Major — `TaskFolderStore.swift:145`** triple folder walk per write — **fixed**. `makeIndex()` builds the id→URL map and the taken-name set in one listing, `write(_:index:)` keeps it consistent, and `exportAll` builds it once for the batch.
- **Minor — `TaskFolderStore.swift:177`** silent `try?` in `currentFilenames` — **fixed**. The function is gone; `makeIndex()` uses `do/catch` and returns `nil`, which both `save` and `exportAll` treat as a refusal rather than as an empty folder.
- **Minor — `TaskTriage.swift:206`** tag charset written four times — **fixed**. `TaskTextParser.isTagCharacter` is the single definition and all four sites call it; `TaskTriage.maxTagLength` is deleted.
- **Minor — `TaskTriage.swift:154`** shared mutable `DateFormatter` — **fixed**. Built per call from the passed calendar's zone; nothing mutates shared state.
- **Minor — `LibraryPanelContainer.swift:47`** duplicated drag machinery — **fixed**. One `ResizableEdge`, both call sites, `project.pbxproj` regenerated. (See Minor 2 for the layout it changes.)
- **Minor — `CategorySidebarView.swift:12`** dead `@Environment(TemplateStore.self)` — **fixed**.
- **Minor — `scripts/test-no-llm.sh:18`** filename entries in `-skip-testing:` — **fixed**. The seven filenames are gone, the sixteen real suite types remain, and the rule is stated in a comment above the loop.
- **Minor — `LibrarySurfaceView.swift:9`** file named after a type it does not declare — **fixed** by rename.
- **Minor — `TaskItem+ActionItem.swift:95`** throwaway task around a fabricated meeting id — **fixed**. `carriedNotes` is `fileprivate` and called directly.
- **Comments** — all three **fixed**: the inspector's width comment no longer claims the machinery lives in `UnifiedSidebarView`, the MARK reads "Task Counts", and the "one definition of what a tag may contain" comment is now true.

## Major

**`Logue/Views/Home/SeedDataBannerView.swift:84`** — the third "Clear Examples" path still does not clear tasks. (Outside this PR's diff, so it is here rather than inline.)

With plain-markdown storage on, pressing "Clear Examples" on the Overview calls `documentStore.clearAllData()` → `clearMarkdownFolderContents()` → `retireRoot()`, which trashes the whole `~/Logue` root — `Tasks/` and every task `.md` inside it — then recreates an empty root. `TaskStore` is never told, so the list on screen still shows every task; the next edit recreates `~/Logue/Tasks` holding only that one task, and on the next launch the rest are only in the Trash. In encrypted mode the mirror gap is intact too: this button leaves `Application Support/Logue/tasks` untouched, so tasks survive a "clear all example data" here while the settings button removes them. Two of the three sibling paths were updated in the fix commit; add `TaskStore.shared.clearAllData()` to this one so the same button behaves the same way in both places.

**Test coverage for the fixes themselves.** The fix commit changed exactly one test file (`LogueTests/TaskFolderStoreTests.swift`, four tests). Three of those genuinely fail at `db480a13`; the fourth, `exportWritesEveryTask`, passes at both SHAs and duplicates `exportAllWritesEverything`, so the `FolderIndex` rewrite it was added for is not actually pinned — a batch with two same-titled tasks, or a retitle inside a batch, is what would exercise `index.filenames`. Nothing covers `avoidingTaskFolder`, the `importAll` task-file guard, `reEncryptFromFolder`/`taskFileCount`, or `clearAllData` — the four fixes with the most data-loss surface, and three of them sit on types that already take their root as a parameter and are trivially testable against a temporary directory, which is how the rest of this PR's storage tests are written. `TaskFolderIsolationTests.taskFilesAreNotAdoptable` looks like it covers the new `importAll` guard but does not: its task file sits inside a marked folder, so `documentFiles` excludes it before the loop is reached, and the test passes unchanged at `db480a13`. Approval needs tests that fail without these fixes.

## Minor

Everything else is anchored inline.

## Compatibility with main

No change here to persisted data. The fix commit renames no `Codable` field, no frontmatter key, no `_logue_id`/`_logue_task_id`/`_space.md` shape; `TaskItem.init(from:)` still uses `decodeIfPresent` with a default on every non-identity field, so meetings and documents written by 1.0.x still decode. No new concurrency hazard: nothing new touches `LLMEngine`, no new `Task` closure or polling loop, and the one shared mutable `DateFormatter` was removed rather than added. No new security-rule trip — no force unwrap or cast, no `url.absoluteString` logged, no `[0]` on a FileManager array, the one silent `try?` was deleted, prompt delimiters and the context-window check unchanged. `project.pbxproj` was regenerated for the added and renamed files.

The one behaviour change to code that predates this PR is the resize handle: `UnifiedSidebarView` now delegates to `ResizableEdge`, which occupies 8pt of layout where the old handle occupied 1pt. That is a shipped surface in both the document and meeting workspaces, changed by a de-duplication refactor, and it is the reason the screenshot ask below is not optional.

CI at `c7441d1f`: Build pass, Lint Check pass. Note that CI runs only `WhatsNewGateTests` and `AppVersionTests`, so green says the target compiles and lints — it does not mean any of this PR's task suites ran. Still no screenshots or recording on the PR, so the "shown to work" gate is unmet independently of the findings above.

shanforge:
**Correcting my own earlier review.** I wrote that "neither CI job runs `scripts/test-no-llm.sh`, so nothing on this PR has executed the tests." That is wrong, and I should not have published it.

`.github/workflows/ci.yml` carries a **`Test (everything that needs no model)` step inside the Build job**. `gh pr checks` lists jobs, not steps, which is what I read — the step has been there at every head on this branch and concluded `success` each time. The tests on this PR do execute in CI, and always have. Nothing was needed to fix it.

Everything else in that review stands, and the outstanding items have now been addressed in `5ef950d`.

shanforge (commented):
Zero blocking and zero major findings, CI green, and the rule that had been rewritten wrong in four consecutive rounds is now executed by something. This is a comment rather than an approval only because the PR still carries no visual evidence — and the body says outright that the task inspector and the two library panels "have not been observed rendering under automation".

## What changed since the round on `5ef950d`

Thirteen minors, all addressed in `9fdddfe`. Four of them were in the test file added the round before, which is where this PR's recurring failure mode lives — a fix landing somewhere nothing could execute — so those are worth naming:

- **The falsification instruction named a mutation that does nothing.** It said to move the `defer` in `retire` above `let folder`. Swift runs a `defer` at scope exit wherever it is registered, so hoisting it is a no-op; anyone following the instruction would have concluded the case was dead and deleted it. The mutation that discriminates is replacing the `defer` with a straight-line `forget()` above the resolve — which is what the code used to be. Corrected, and re-checked by performing it: three of eight cases go red.
- **`aMissingRootKeepsTheKnownFolder` passed with the guard it is named for deleted.** It hid the root behind an injected `fileExists`, but `resolve(in:)` reaches the filesystem through `TaskFolderStore`, which never saw the fake — so the walk found the folder either way. It now removes the scratch root for real; deleting the guard makes it answer `<root>/Tasks`, confirmed. The `fileExists` injection went with it: nothing else passed one, and a seam whose only user did not work is surface rather than a seam.
- The restore step in `aRestoredFolderIsNotStillOurs` was inert — nothing resolved after it, so the final assertion could not be affected by it and merely repeated the previous case. It now resolves and checks the folder is adopted afresh.
- `aFailedTrashDoesNotForget` asserted the opposite of its own body. `xcodebuild` and Xcode's navigator key on the symbol, not the display string, so someone triaging a red run would have read an inverted contract and "fixed" `retire` by moving the forget out of the `defer` — reintroducing the defect two other cases exist to catch.

One latent bug came out of the craft tier and is worth more than its tier: **`retire` held a non-recursive `NSLock` across the caller's closure.** Not reachable today — the only call site passes `FileManager.trashItem` — but a closure that asked this locator anything would deadlock, and a log line naming `TaskStorage.tasksFolderURL` would be enough. It deadlocks *while deleting the user's tasks folder*. Nothing about the ordering needed one critical section: the point is that the resolve happens before the forget, not that nothing runs in between. The lock is now taken and released by `folderURL` and `forget()` separately.

Also: the step-aside to `Tasks 2` and the exhaustion of that loop were silent and the new file had no logger at all, so a user with a space named `Tasks` had nothing anywhere answering "why is there a `Tasks 2`"; `invalidate()`'s doc claimed a Finder rename arrives through the scan, which the two comments below it correctly deny — it read as licence to delete the existence check that actually catches a rename; `retireRootAndForgetTasks(using:)` lost its parameter, which had re-opened the seam the extraction closed (a caller passing a migrator over a different root would retire one folder and forget the marker for another); `locator` is private, since internal exposed `retire` and `forget` to the whole app; and the `mode` `didSet` comment no longer claims to cover the value read back at launch, which property observers do not see.

## Compatibility with main

Clean. This round adds no `Codable` field, frontmatter key, marker format or enum raw value, so meetings, documents and tasks written by 1.0.x still decode. `AppConstants.UserDefaultsKeys.lastTaskFolderMarker` keeps its value and stays absent-tolerant.

## Verification

CI at `9fdddfe`: Build pass 7m27s, Lint Check pass 35s. The Build job's `Test (everything that needs no model)` step ran **1347 tests in 108 suites, all passing** — worth stating plainly because an earlier review of mine claimed the tests never execute in CI. That was wrong: the step is inside the Build job, and `gh pr checks` lists jobs, not steps. `scripts/test-no-llm.sh` being a deny-list rather than an `-only-testing:` allow-list is why the number is 108 suites and not a handful; the two other open PRs still use allow-lists and run 17 and 7 suites respectively, so their green means considerably less than this one's.

Every behavioural claim in this review was checked by reverting the fix and watching the named cases fail, not by reading.

## What is needed to approve

Only the click-through — every other gate is met.

1. **The three surfaces nobody has seen render**: the task inspector, the triage panel, and the two library panels, populated. A screenshot of each.
2. **An inspector edit does not clobber its neighbours.** Change a task's title only, its notes only, then both. Correct: each persists and priority, due date, tags and status are untouched.
3. **A focused field does not write back stale text.** Open a task, put the cursor in the notes field, and change that task from elsewhere (or let a folder scan change it). Correct: the field updates or stays put, but nothing the user did not type is saved. This is the one finding static review could not settle in either direction — three rounds have tried.
4. **A folder renamed in Finder keeps working.** Enable plain-markdown storage, let `~/Logue/Tasks` appear, rename it to `Errands` in Finder, then tick a checkbox in the app. Correct: the change lands in `Errands`. Wrong: a new empty `Tasks` appears and the list empties.
5. **A space named Tasks is not colonised.** Create a space called `Tasks` first, then enable markdown storage. Correct: tasks go to `Tasks 2`, the space keeps every document, and the log says why.
6. **A restored folder is not re-adopted.** In markdown mode use Clear Examples, confirm the tasks folder is in the Trash, then restore it by hand. Correct: the app does not adopt it again.
7. **Old data still decodes** — launch against tasks and meetings written by 1.0.x.
8. **The shared sidebar handle, on both surfaces it changed.** `UnifiedSidebarView` now uses `ResizableEdge`, so a meeting workspace *and* a document workspace both changed. Correct: the divider is grabbable across an 8 pt band, the hairline is still 1 pt, and a dragged width survives a restart.

When those come back, re-checking that nothing else moved is quick and the approval is cheap.

shanforge (approved):
Approving. Zero blocking and zero major findings after twelve rounds, CI green, tests that I confirmed discriminate — and the visual gate is now met: I drove this branch's build by hand, and the reviewer confirmed the click-through.

The PR body said the task inspector and the library panels "have not been observed rendering under automation". They have now.

## What I exercised on a build of `9fdddfe`

- **The Tasks surface renders**, including its empty state: "Nothing Here" with the quick-add prompt, the filter chips (All / Today / Overdue / Upcoming / No Date / Completed), sort and search.
- **Natural-language capture parses correctly.** Typing `Send the deck tomorrow #launch !` produced a task titled "Send the deck" with a due date of **Aug 19 2026** (tomorrow, resolved against the real clock), priority **High** from the `!`, and the tag **launch** — with the syntax stripped from the title rather than left in it. A second task with `#docs` and no date landed with no due date. The filter counts moved with them: Upcoming 1, No Date 1.
- **The inspector renders in full** — title, due date with stepper and Clear, the Low/Medium/High control reflecting the parsed priority, tag chips with removal, the repeat control, notes, and delete.
- **An edit touches only what was edited.** Typing notes on the first task left its title, due date, priority and tag exactly as they were, and the note appeared on that task's row in the list.
- **A focused field does not write back into the task that replaces it.** This is the finding three consecutive rounds could not settle in either direction, so it is the one I went after specifically. With the cursor still in the notes field, I switched rows. The note committed to the task it was loaded from — visible on "Send the deck" — and the inspector for "Draft the release notes" came up with **its own** empty notes, its own title, its own tag and no due date. Nothing bled across. Switching back showed the note still there.

That is `TaskTextCommit` refusing a commit whose `taskID` does not match, observed rather than reasoned about.

## What I did not exercise

The plain-markdown folder paths — rename in Finder, the `Tasks 2` step-aside, and restore-from-Trash — were not clicked, because they need storage-mode switching against a real library. I am comfortable approving over that: those are exactly the rules `TaskFolderLocator` now owns, they are driven directly by `TaskFolderLocatorTests`, and I verified that suite discriminates by reverting the resolve/forget order and the missing-root guard and watching the named cases turn red. That is stronger evidence than a single click-through would have been.

## Verification

CI at `9fdddfe`: Build pass, Lint Check pass, and the Build job's test step ran **1347 tests in 108 suites**, all passing. This branch is also the one that switched CI from an `-only-testing:` allow-list to `scripts/test-no-llm.sh`, which is why that number is 108 suites rather than a handful — worth carrying to the other branches.

PR #58, opened 2026-08-13
Automatic capture, recording resilience, and a transcript that stops moving

westerosweb:
Recording no longer asks the user what to capture, survives things going wrong mid-meeting, and stops rearranging the transcript under the reader once speakers are identified.

Driven by customer feedback: the capture controls confused people, meetings were lost to crashes and dropped headsets, and the transcript rewrote itself a minute after recording stopped.

Design doc: `docs/specs/2026-08-12-automatic-capture-and-recording-resilience.md`.

## Capture decides for itself

The recording-mode picker, the "Online Meeting Detected" prompt, and the mic/system-audio toggles are gone. The microphone always runs; `SystemAudioArmingMonitor` watches Core Audio and arms the ScreenCaptureKit tap about a second after anything starts playing, then leaves it armed for the session. `RecordingMode` becomes a description of what happened rather than a question asked up front, and the second `SpeechTranscriberEngine` — which existed only for "started on system audio, added mic later" — is deleted.

A mute button remains. It mutes; it is not a mode.

## Surviving the meeting

- **Device loss.** `CaptureDeviceMonitor` holds a grace period sized by connection type (3s Bluetooth, 2s wired), then falls back to the default device, placing the gap correctly on the timeline. Recording never stops and nothing is asked.
- **Crash recovery.** In-progress audio moved out of the temporary directory, which was flagged delete-on-reboot. A `RecordingCheckpoint` every 30s carries the transcript and both capture timelines — the part that is otherwise only resolved at stop — and `RecordingRecoveryService` silently rebuilds an interrupted session at launch, bounded by `TranscriptReplacement`.
- **Screen recording access** is preflighted every time and requested once per launch. Persisting the refusal meant a lapsed grant could never be recovered from, and the app quietly stopped recording the far side of every call.

## Bugs found and fixed along the way

- **The microphone was silent.** `setVoiceProcessingEnabled(true)`, added here for echo cancellation, silenced the input outright — every timeline window logged `RMS 0.00000` while the level meter kept moving off the system tap. Reverted. `TimelineContributionLog` now reports what each source actually contributes, so this cannot happen unnoticed again.
- **Tap buffers were read after they expired.** Audio handed to an async consumer without copying; the engine reuses that memory the moment the callback returns.
- **A VAD filter could delete a transcript.** It removed two of three already-transcribed segments as "silence". It now refuses to remove half or more, and says so.
- **Sub-word tokens were split across lines**, turning "Hundreds" into "H" / "undreds".

## The transcript stops moving

Identifying speakers used to re-cut everything: a different number of blocks, text merged and split, the reader's place lost. Now blocks are decided by time alone and the speaker appears in the gutter as `S1 0:26`, so diarization only ever adds labels.

The post-recording pass is still more accurate word-for-word, so its words are poured into the live segments via `TranscriptRealignment` — every line keeps its identity, start and end, and only the text improves. Lines are merged into whole sentences as they arrive, and realigned lines are snapped to sentence ends.

Also: bookmarks belong to a timestamp rather than the whole transcript, AI highlights are anchored to the line their words appear in rather than the model's guessed timestamp, and drag-to-reassign-speaker is removed now that diarization is accurate enough.

## Testing

1064 tests across 100 suites. New pure units under test: arming debounce, transcription gate, pre-roll buffer, device-loss policy, checkpoint round-trip, recovery bounds, VAD filter, timeline conversion, detached buffers, gutter marks, realignment, sentence merging, speaker labels, highlight anchoring.

Verified by hand on device: automatic arming, Bluetooth detection, silent recovery of an interrupted session, and microphone capture before and after the voice-processing revert.

## Known gaps

- Echo cancellation is gone with voice processing, so the mic may re-hear remote participants through speakers. Headphones avoid it.
- Reassigning a single line to a different speaker is no longer possible; renaming a speaker still is.
- `SpeakerShortLabel` reduces a one-word name to one letter, so two named speakers can collide on the same initial.


shanforge (changes requested):
This PR removes the capture-mode question, adds device-loss and crash recovery, and stops diarization re-cutting the transcript. The design is right and most of the new pure units are clean, but four defects in the paths the PR exists to add will lose or expose user data as written: realignment ignores the session offset, crash recovery runs before the meeting store has loaded (and then deletes the audio it was meant to rebuild), the VAD over-filter fix was applied to only one of the two identical filters, and the checkpoint writes the transcript to disk in plaintext.

## Blocking

**`Logue/Services/RecordingSessionManager+Diarization.swift:158`** — `TranscriptRealignment.realign(live:words:)` is passed the whole meeting's segments together with batch word timings that are relative to *this session's* audio buffer, with no `sessionStart` shift.

`startRecording(for:)` sets `timeOffset = lastEnd + 1.0` when a meeting already has segments, so a resumed session's live lines are stamped at 301s+ while `lastBatchWords` (built from `result.tokenTimings` in `DiarizationManager+BatchASR.swift:52`) runs 0…T2. `indexOfSegment(containing:)` therefore drops every session-2 word into session-1's lines and overwrites their text, while session-2's lines keep the live draft. Record 5 min, stop, hit record again on the same meeting, stop — the first five minutes now read as the second session. The removed `replaceTranscript(..., sessionStart:, heardDuration:)` did the shift; the sibling call `applySortformerUpdates(normalized, for:, sessionStart: sessionStart)` eight lines below still does.

Fix: shift each `TimedWord` by `sessionStart` before realigning (or bound `live:` to `segments.filter { $0.startTime >= sessionStart }` and shift), and add a `TranscriptRealignmentTests` case with a non-zero offset and a prior session's lines present — every existing case uses offset 0.

**`Logue/Services/RecordingSessionManager+Recovery.swift:33`** — recovery reads `MeetingStore.shared.meetings` before the store has finished loading, concludes the meeting no longer exists, and hard-deletes the interrupted session's audio and checkpoint.

`MeetingStore.init()` enqueues `Task { await loadFromDiskAsync() }`, which suspends at `try await Task.detached { … }.value` (`MeetingStore+Persistence.swift:143`) while it reads and decrypts every meeting file. The `.task` in `LogueApp.swift:218` is enqueued after it on the same actor, so it runs during that suspension with `meetings == []`. `recoverSession` reads the checkpoint fine, fails this guard, logs "belongs to a meeting that no longer exists", and calls `InProgressRecordingStore.clear(meetingID:)` — which `removeItem`s the whole working directory. Every interrupted recording is destroyed at the exact launch meant to rebuild it, and the comment above the `.task` ("After the store has loaded") is not enforced by anything. `MeetingStore.isLoaded` exists for this and is never consulted.

Fix: await `isLoaded` (poll with a bounded `ContinuousClock` deadline, per the polling rule in CLAUDE.md) before the recovery loop, and never `clear()` on a "meeting not found" result until the store is known to be loaded.

**`Logue/Engine/DiarizationManager+LongRecording.swift:146`** — the "a VAD filter could delete a transcript" fix was applied to `BatchTranscriptFilter` only; `filterWithFileVAD` still keeps the old `filtered.isEmpty ? segments : filtered` guard, so the bug the PR claims to have fixed survives on the disk path.

`BatchTranscriptFilter.filter` refuses when `kept.count * 2 < segments.count`; `filterWithFileVAD` has no such check. This path is taken by every recovery (`RecordingSessionManager+Recovery.swift` → `processRecordingFile`) and by every recording that outgrows the in-memory mixer. On the recovery route the filtered output goes straight into `replaceTranscript(for:with:sessionStart:heardDuration:)`, which drops every existing segment inside the covered window and substitutes the batch result — so a VAD that hears only the loud stretch of a quiet meeting deletes the majority of the recovered transcript, which is exactly the failure the PR body describes. Nothing tests `filterWithFileVAD`; `BatchTranscriptFilterTests` only exercises the pure enum.

Fix: route this filter through `BatchTranscriptFilter.filter` with the regions already computed, and log `distrustedVAD` the same way `filterWithVAD` does.

**`Logue/Engine/RecordingCheckpoint.swift:40`** — the checkpoint writes `meeting.segments` — the full transcript — to Application Support as plaintext JSON, bypassing the AES-256-GCM encryption every other meeting write goes through.

`try JSONEncoder().encode(checkpoint)` then `data.write(to:options: .atomic)` under `<Application Support>/Logue/InProgress/<meetingID>/checkpoint.json`, refreshed every 30s for the whole duration of every recording, where the equivalent write in `MeetingStore+Persistence.swift:30` is `try EncryptionManager.encryptCodable(meeting)`. After a crash the plaintext file persists on disk until a recovery pass consumes it. For an app whose data layer is documented as encrypted at rest, this puts an unencrypted copy of every transcript on disk during every meeting.

Fix: `EncryptionManager.encryptCodable(checkpoint)` on write and the matching decrypt on read.

## Major

**`Logue/Services/RecordingSessionManager+DeviceLoss.swift:43`** — a capture-device drop silently un-mutes a microphone the user muted, because mute is expressed only as `isMicActive == false`.

`setMicMuted(true)` stops the tap and sets `isMicActive = false`; nothing else records the intent, and `MeetingWorkspaceView.swift:258-264` drives the button from that same flag. `deviceMonitor` keeps running (only `teardownAudioPipeline` stops it). A Bluetooth headset renegotiating while muted produces `.keepWaiting` (no-op) then `.fallBackToDefault`, which calls `resumeMicrophoneCapture()` whose only gate is `guard !isMicActive else { return }` — it passes, the tap is reinstalled, `isMicActive = true`, and a muted microphone is recording and transcribing with the UI showing it live. Same via `.resumeSameDevice`.

Fix: add an explicit `isMicMuted` flag set by `setMicMuted`, return early from `resumeMicrophoneCapture()` while it is set, and while you are there collapse the two copies of the resume sequence — `resumeMicrophoneCapture()` repeats the unmute branch of `setMicMuted` step for step and they have already drifted (unmute failure sets `errorMessage`, device-loss failure sets `captureNotice`).

**`Logue/Services/MeetingStore+Transcript.swift:27`** — `restoreRecoveredSession` replaces the meeting's segments with the checkpoint's, discarding transcript that was already saved correctly.

The transcript is persisted every 3s during recording (`appendSegment` → `scheduleDebouncedSave` → `AppConstants.Delays.liveRecordingSaveDebounce = 3s`), while checkpoints are written every 30s. After a crash the meeting file on disk holds *more* lines than the checkpoint, and `meetings[index].segments = segments` throws the newest ones away — up to ~27s of correct, already-durable transcript deleted by the recovery feature.

Fix: merge by segment id / keep whichever set reaches further, rather than assigning wholesale.

**`Logue/Services/RecordingSessionManager+Recovery.swift:13`** — recovery checks `recordingState == .idle` only at the top of the loop, then performs minute-long awaits during which a new recording can start and be damaged.

`startRecording` guards on `recordingState`, which recovery never sets, so ⌘R (or `pendingAutoRecord`) can begin a session while recovery is inside `persistRecordingAudio` (splice/compose are `await`ed and slow) or `reprocessRecoveredAudio`. Three concrete collisions on the shared `RecordingSessionManager.shared`: `persistRecordingAudio` ends by calling `audioRecorder.clearTemporaryFile()` (splice path, and `moveIntoPlace`'s `defer`), which `removeItem`s the *live* recording's `tempFileURL` and nils it — at stop `sources.micURL` is nil and the meeting saves with no audio; `recoverSession`'s final `InProgressRecordingStore.clear(meetingID:)` deletes the working directory, which is the same directory if the user re-records the recovered meeting; and `reprocessRecoveredAudio` sets and clears `isDiarizing`/`diarizationStage` underneath the live session.

Fix: take the recording state for the duration of recovery (a dedicated `.recovering` state or a guard flag `startRecording` respects), and re-check it after every await.

**`Logue/Engine/DiarizationManager+LongRecording.swift`** — the disk transcription route never populates `lastBatchWords`, so the PR's headline realignment silently does not happen for long recordings.

`lastBatchWords` is assigned only in `transcribeBuffer` (`DiarizationManager+BatchASR.swift:47,52`). `transcribeRecordingFile` never touches it, so it stays `[]`, and the realign block at `RecordingSessionManager+Diarization.swift:154` is guarded by `!diarizer.lastBatchWords.isEmpty`. A meeting that overruns `AudioTimelineMixer` therefore gets speakers and no text improvement, with no log line saying so — where on main it reached `replaceTranscript`. The divergence between short and long recordings is untested and unstated.

Fix: build `lastBatchWords` from `result.tokenTimings` on this route too (they are already in hand at `transcribeRecordingFile`), or log explicitly that realignment was skipped.

**`Logue/Views/Meeting/TranscriptTimelineView.swift:254`** — the block cache is keyed on `segments.count`, `searchText` and the speaker-label hash, so a change that alters only segment *text* never re-renders — which this PR makes the normal case.

`let key = "\(segments.count)|\(searchText)|\(speakerLabelHash)"` guarded by `guard key != blockCacheKey`, and `SpeakerBlock` holds copies of the segments. Live: the transcriber finalises "The agent helps assemble" (count 0→1, renders), then the continuation is folded into it by `TranscriptSentenceMerge` in `appendSegment` — count unchanged, labels unchanged, none of the four `onChange` triggers fire — so the volatile text disappears and the finalised continuation is invisible until a later segment happens not to merge. The post-recording realignment case is masked in practice, because `speakerColors` changes when diarization adds speakers and `forceRecomputeBlocks()` fires; it is not masked when diarization returns no speakers.

Fix: include a cheap text signature in the key (e.g. hash of `segments.count` plus each `id` and `text.count`), or drop the cache and let `computeSpeakerBlocks()` run from the segments directly.

**`Logue/Engine/TimelineAudioConverter.swift`** — a near-verbatim copy of `DiarizationManager.convertBufferToFloatArray` with no production caller, whose tests report the shipped conversion path as covered when it is not.

`git grep TimelineAudioConverter` at head returns its own declaration and four hits in `LogueTests/TimelineAudioConversionTests.swift`. `DiarizationManager.swift:367` still calls `convertBufferToFloatArray`. The two have already diverged in both directions: the copy adds `+ 64` frames of capacity ("a buffer one frame too small makes it stop early") which the shipped one lacks, and the shipped one sets `converter.primeMethod = .none` which the copy lacks — so the tests do not describe the converter that runs. The file's own doc says it was "Lifted out of `DiarizationManager`", which it was not.

Fix: make `DiarizationManager` use it (per source, keeping `primeMethod`), or delete it and its tests.

**`Logue/Engine/TranscriptSentenceMerge.swift:29`** — the "never merge across speakers" guard is inert in the only place `shouldMerge` runs, and the test that covers it hand-sets labels the live path never has.

`MeetingStore.appendSegment` (`MeetingStore.swift:422`) is the sole caller and runs during recording, where segments arrive from `SpeechTranscriberEngine` with `speakerLabel` nil — speakers are assigned only after recording stops. So the guard is always `nil == nil`. With one transcriber now fed by both mic and system tap, a local utterance with no terminal punctuation followed 0.4s later by a remote reply satisfies all four checks and becomes one line carrying one speaker id and one `audioSource`. Blocks are cut by time and diarization only adds labels, so nothing can split it again — and drag-to-reassign was removed in this same PR, so there is no manual repair either.

Fix: also require `previous.audioSource == next.audioSource` before merging, and make `TranscriptSentenceMergeTests.speakersAreNeverMerged` exercise nil-labelled segments.

## Minor

**`Logue/Services/RecordingSessionManager+Recovery.swift:70`** — `guard let directory = try? InProgressRecordingStore.directory(for:)` is a silent `try?` on a directory-creating call, which CLAUDE.md forbids for any operation other than `Task.sleep`. The failure it swallows is exactly the one the `.task` comment says this code must tolerate (a container that moved under `SandboxContainerMigrator`), and it degrades to "recovered meeting with no audio" with nothing in the log. Use `do/catch` and log.

**`Logue/Services/RecordingSessionManager+Diarization.swift:136`** — `let heardDuration = pass.heardDuration` is bound and never read after the `replaceTranscript` call was deleted; the only other occurrence in the file is a comment claiming it still bounds the pass. This is a compiler warning. Either apply it (bound `realign` to the heard stretch) or drop the binding and the comment together. The old data-loss mode is genuinely gone here — `realign` never blanks a line — so this is dead code rather than a live bug.

**`Logue/Services/MeetingStore.swift:422`** — sentence merging can fold the first line of a resumed recording into the last line of the previous session: `timeOffset = lastEnd + 1.0` puts the gap at 1.0s, inside `TranscriptSentenceMerge.maximumGap` of 2s. It only fires when the previous session left every `speakerLabel` nil (diarization found nothing), but the merged line then straddles a session boundary that `TranscriptReplacement`'s `startTime < sessionStart` filter, playback seek and bookmark anchoring all treat as session 1. Skip the merge when `next.startTime >= sessionStart`.

**`Logue/Engine/TranscriptRealignment.swift:59`** — `words(fromTokens:)` accepts `\u{2581}` as a word-start marker but `flush()` joins the pieces verbatim (line 49) and nothing anywhere strips it; `trimmingCharacters(in: .whitespacesAndNewlines)` does not remove it. If Parakeet ever emits the marker form this writes literal `▁` into the transcript with no word separation, and if it never does the branch is dead. `TranscriptRealignmentTests.sentencePieceMarker` asserts only `joined.count == 1`, never the text. Strip the marker in `flush()` and assert the resulting string.

**`Logue/Engine/TimelineContributionLog.swift:62`** — the dropped-conversion warning carries a 17-space run mid-sentence, so it prints as `… from microphone                 — conversion produced nothing`. The sibling message at lines 39-42 uses `\`-continuation correctly.

**Dead code left behind by the removals.** `Logue/Services/ConferencingAppDetector.swift:4` has no callers at all after both were deleted in this diff; `Logue/Services/MeetingStore.swift:212` `reassignSegmentSpeaker` has none after drag-to-reassign was removed; `Logue/Views/Meeting/TranscriptTimelineView.swift:15` `SpeakerBlock.formattedStartTime` has none now that the block header deliberately prints no time; `Logue/Views/Meeting/TranscriptSpeakerBlockView.swift:31` `showBookmarkAdded` is written at 389 and 392 and never read, so the confirmation `AppConstants.Delays.bookmarkConfirm` is named for never appears, and `isHovered` (22, 366-367) drives nothing now that `hoveredGroupID` handles hover chrome.

**Test coverage.** `LogueTests/RecordingRecoveryTests.swift` does not touch any of the new recovery code — its five cases call `TranscriptReplacement.merged`, `RecordingCheckpoint.coveredDuration` and `CaptureSegmentTimeline`, all pre-existing or pure. Nothing constructs an in-progress directory with audio plus a checkpoint and runs `recoverInterruptedSessions()`, so none of the four decisions that determine whether an interrupted meeting survives is verified. `MicrophoneSpeechGate.admit` — whose failure mode is a silently empty transcript — has no test either, only its two sub-units.

## Comments to delete

- `Logue/Views/Meeting/TranscriptTimelineView.swift:57-59` — `/// Groups consecutive segments by speaker. Starts a new block when the speaker changes / or there's a >15s gap between segments from the same speaker. / Segments without a speaker label merge into the previous block…` sits six lines above the new comment saying splitting on speaker was removed, and the only branch left is `if timeGap`.
- `Logue/Views/Meeting/TranscriptTimelineView.swift:352` — `// Renders a group of consecutive segments from the same speaker as a card.` is a tombstone under `// MARK: - Speaker Block View` for a type that moved to `TranscriptSpeakerBlockView.swift`, and it describes a grouping rule this diff removed.
- `Logue/Views/Meeting/TranscriptSpeakerBlockView.swift:57-61` — the five-line "What the gutter shows for a line…" block floats between declarations, separated by a blank line from the `groupView` doc it appears to introduce; the behaviour it describes is `gutterSpeaker(for:mark:revealSpeaker:)`, 135 lines below.
- `Logue/Services/RecordingSessionManager+Diarization.swift:141-152` — the 14-line comment states the same point twice ("it re-cut every sentence: a different number of blocks…" then "its sentence boundaries are its own. Adopting them re-cut the transcript…") and makes two claims the code contradicts: "the transcript never changes under anyone" immediately above `updateSegments(realigned, for: meetingID)`, and "`heardDuration` still bounds how much of the session that timeline may speak for" above a binding with no readers.

## Compatibility with main

Persisted-data compatibility is clean: `MeetingNote.wasRecovered` is `decodeIfPresent ?? false` with the key added to `CodingKeys`, no keys were renamed, no enum cases changed, and `CaptureSegmentTimeline.Placement` gaining `Codable` is additive — meetings written by 1.0.1 still decode here and meetings written by this build still decode on 1.0.1. The behavioural compatibility is where it is not safe to merge: an existing user who resumes a recording loses the earlier session's transcript (blocking 1), an existing user who crashes mid-meeting has the crash audio deleted at the next launch (blocking 2) and, once that is fixed, can have the recovered transcript majority-deleted by the unguarded VAD (blocking 3); and every user of this build gets an unencrypted transcript on disk for the duration of every recording (blocking 4). None of these is caught by the suite as written. Fix the four, add the offset and recovery tests, and the rest is mergeable.

westerosweb:
All nine addressed. Every claim verified against the code first; all nine were real, so no pushback. 1148 tests pass.

## Blocking

**Realignment ignored the session offset** — `realign(live:words:sessionStart:)` now shifts each word onto the meeting's timeline, and the call site passes the same `sessionStart` the sibling `applySortformerUpdates` already used. Two tests added: a resumed session at 301s whose earlier lines must not change, and a zero-offset equivalence check.

**Recovery ran before the store loaded, then deleted the audio** — recovery now waits for `MeetingStore.isLoaded` behind a bounded deadline, and the "meeting no longer exists" branch refuses to delete anything unless the library is known to be loaded. Both were needed: the wait alone would still delete on a timeout.

**Over-filter fix reached only one of the two filters** — `filterWithFileVAD` now routes through `BatchTranscriptFilter.filter` with the regions it already computed, and logs `distrustedVAD` the same way. The existing `BatchTranscriptFilterTests` now cover both paths, since they exercise the shared rule.

**Checkpoint wrote the transcript in plaintext** — now `EncryptionManager.encryptCodable` on write and the matching decrypt on read. Added a test that writes a checkpoint containing a known sentence and asserts the bytes on disk do not contain it, then that reading it back returns it.

## Major

**Device loss silently un-muted the microphone** — added `isMicMuted` as explicit intent, separate from `isMicActive` which only tracks whether the tap is running. `resumeMicrophoneCapture()` returns early while muted. The two copies of the resume sequence are collapsed into that one method, which was where they had already drifted.

**Recovery discarded newer saved transcript** — `restoreRecoveredSession` now merges by segment id and only when the checkpoint reaches further than what is already saved, so the ~27s the debounced save had already persisted survives.

**Recovery could be trampled by a new recording** — added a `.recovering` state held for the whole pass, which `startRecording` refuses to cut across. That closes all three collisions: `clearTemporaryFile()` deleting the live recording's audio, the working-directory delete, and `isDiarizing` being driven underneath a live session.

**Disk route never populated `lastBatchWords`** — it now builds them from `result.tokenTimings` exactly as the in-memory route does, so long recordings and recoveries get the realignment too, and clears them on the no-timings path.

**Block cache missed text-only changes** — the key now includes a signature of each segment's id and text length. This was the sharpest catch: sentence merging made count-stable updates the normal case, so the live transcript could stop updating mid-recording.

Logue/Views/Meeting/TranscriptTimelineView.swift
```diff
@@ -250,7 +263,11 @@ struct TranscriptTimelineView: View {
     }
 
     private func recomputeBlocksIfNeeded() {
-        let key = "\(segments.count)|\(searchText)|\(speakerLabelHash)"
+        // Includes a signature of the text, not just the count. Merging a continuation into the
+        // line before it leaves the count unchanged, and realignment rewrites text without touching
+        // the count at all — both were invisible to a key built from counts alone, so the transcript
+        // stopped updating mid-recording.
+        let key = "\(segments.count)|\(searchText)|\(speakerLabelHash)|\(segmentTextHash)"
```
shanforge, line 277: **Major** — the key now discriminates on text, but nothing calls the function that computes it when only text changed, so the finding this was written for still reproduces exactly as before. The triggers are unchanged at lines 258-262: `onAppear`, `segments.count`, `searchText`, `speakerColors`, `speakerLabelHash`. `segmentTextHash` appears nowhere else in the file.

Live recording: the transcriber finalises "The agent helps assemble" (count 0→1, block renders), then the continuation is folded into that line by `TranscriptSentenceMerge` inside `appendSegment` — count unchanged, `searchText` unchanged, `speakerColors` unchanged, `speakerLabelHash` unchanged because labels are nil until diarization runs. `body` re-evaluates, no `onChange` fires, `cachedBlocks` still holds the pre-merge copies, and the continuation stays invisible. Same for post-recording realignment when diarization returns no speakers.

Add `.onChange(of: segmentTextHash) { recomputeBlocksIfNeeded() }` to the modifier list. While you are there: hashing `text.count` rather than the text means a same-length rewrite (a corrected word) still collides — worth `combine(segment.text)` given realignment's whole job is swapping words for better ones of similar length.

Logue/Services/RecordingSessionManager.swift
```diff
@@ -88,72 +96,112 @@ final class RecordingSessionManager {
...
     var isMicActive = false
+
+    // Extension-visible: +DeviceLoss
+    /// Whether the user has muted the microphone.
+    ///
+    /// Distinct from `isMicActive`, which only says whether the tap is running. A device dropping
+    /// also stops the tap, so without recording the *intent* separately a headset renegotiating
+    /// while muted would be treated as capture to restore, and a muted microphone would quietly
+    /// start recording again.
+    private(set) var isMicMuted = false
```
shanforge, line 132: **Major** — nothing ever clears this. `setMicMuted` is the only writer; `startRecording`, `startMicrophoneRecording` and `stopRecording`'s defer (which resets `recordingState`, `isCapturingSystemAudio` and `isMicActive`) all leave it alone. So a mute leaks into every later meeting of the same app run.

Meeting A: mute the mic, press Stop while still muted — the flag stays true. Meeting B: start recording (mic active), and the Bluetooth headset renegotiates mid-meeting. `handleMicrophoneLoss(.keepWaiting)` stops the tap, the device returns as `.resumeSameDevice` or `.fallBackToDefault`, and `resumeMicrophoneCapture()` returns at its `guard !isMicMuted` with only a log line. The microphone is dead for the rest of meeting B — the exact silent capture loss the device-loss handling exists to prevent — while `.fallBackToDefault` has already set `captureNotice = "Switched to …"`, which is now untrue. Only a manual click of the mic button recovers it.

Set `isMicMuted = false` alongside `isMicActive = true` in `startMicrophoneRecording`, and in `stopRecording`'s defer.

Logue/Engine/RecordingCheckpoint.swift
```diff
@@ -0,0 +1,62 @@
...
+        try data.write(to: url, options: .atomic)
+    }
+
+    static func read(meetingID: UUID) -> RecordingCheckpoint? {
+        do {
+            let directory = try InProgressRecordingStore.directory(for: meetingID)
+            let url = directory.appending(component: fileName)
+            guard FileManager.default.fileExists(atPath: url.path) else { return nil }
+            let data = try Data(contentsOf: url)
+            return try EncryptionManager.decryptCodable(RecordingCheckpoint.self, from: data)
```
shanforge: **Major** — encrypting the checkpoint added a new class of read failure, and the branch that consumes a failed read hard-deletes the recording. Every error here is caught and returned as `nil`, and `recoverSession` reads `nil` as "a session that crashed inside its first thirty seconds" and calls `InProgressRecordingStore.clear(meetingID:)`, which `removeItem`s the whole working directory — mic file, system file and all. Not trashed; removed.

Before this commit the only way to reach that branch was a corrupt file written atomically. Now it is also reached whenever `decryptCodable` throws: the key is stored `kSecAttrAccessibleWhenUnlockedThisDeviceOnly`, so restoring the Mac from a backup or migrating to a new machine leaves `getOrCreateKey` generating a fresh one, `AES.GCM.open` fails, and the audio of a crash-interrupted meeting is destroyed at the first launch. Anyone dogfooding this branch who crashed under the pre-fix build hits the same path with a plaintext checkpoint.

Distinguish absent from unreadable: return `nil` only when the file does not exist, and on a decrypt/decode error keep the directory (the same instinct as the `isLoaded` guard added below it). The commit was careful to add exactly that guard for the meeting-not-found branch and not for this one.

Logue/Services/RecordingSessionManager+Recovery.swift
```diff
@@ -0,0 +1,205 @@
...
+            // thirty seconds. There is nothing to rebuild from, so it is cleared rather than kept.
+            logger.info("Discarding an interrupted session with no checkpoint")
+            InProgressRecordingStore.clear(meetingID: meetingID)
+            return
+        }
+
+        guard MeetingStore.shared.meetings.contains(where: { $0.id == meetingID }) else {
+            // Only safe to conclude this because the library is known to have loaded — see
+            // `waitForMeetingStore`. Deleting on an unloaded store destroys good recordings.
+            guard MeetingStore.shared.isLoaded else {
```
shanforge: **Major** — `isLoaded` is not evidence that the library was actually read. `loadFromDiskAsync` sets it on all four terminal paths, two of which are failures: the `guard !loaded.isEmpty` branch (every meeting file failed to decrypt → `meetings = []`) and the `catch` (which substitutes seed meetings). Both satisfy this guard and both satisfy `waitForMeetingStore()`.

Walk it: a machine restored from backup, or a partial `SandboxContainerMigrator` run that moved `InProgress` but not the meetings directory. The library loads as empty or as seed data, `isLoaded` is true, this guard passes, the `meetings.contains` check above fails, and `InProgressRecordingStore.clear` removes the crashed recording's audio — the outcome the whole `waitForMeetingStore` fix exists to prevent, reached through the flag the fix chose as its evidence.

`MeetingStore.loadedSeedData` is set true on exactly those two failure paths and false on the success path — gate the deletion on `isLoaded && !loadedSeedData && !meetings.isEmpty`, so "we read nothing" never reads as "the meeting was deleted".

Logue/Services/RecordingSessionManager.swift
```diff
@@ -251,20 +295,23 @@ final class RecordingSessionManager {
 
     // swiftlint:disable:next function_body_length
     func startRecording(for meeting: MeetingNote) async {
-        guard recordingState == .idle else { return }
+        guard recordingState == .idle else {
+            if recordingState == .recovering {
```
shanforge: **Major** — the refusal is silent, and two of the three ways a user starts a recording destroy the request rather than retrying it. Neither `errorMessage` nor `captureNotice` is set here, and `.recovering` is not surfaced by `startStopButton` the way `isDiarizing` is (`isDiarizing` is only set inside `reprocessRecoveredAudio`, i.e. *after* the composition step), so throughout the compose phase the toolbar renders a live, enabled "Record" button.

Relaunch after a crash mid-long-meeting: `.recovering` is held across `persistRecordingAudio` (splice/compose of a multi-hour file) and `diarizer.initialize()`, minutes at a time. (1) The user picks New Meeting — `MeetingWorkspaceView`'s `.task(id:)` sets `store.pendingAutoRecord = nil` *before* awaiting `startRecording`, so the request is consumed and never retried and automatic capture, this PR's headline feature, silently does nothing. (2) The menu-bar item calls `startRecording` and then unconditionally shows the recording panel, so the user gets an island for a session that never started. (3) The in-app button just doesn't change state.

Set `captureNotice = "Finishing an interrupted recording…"` when the press is refused, and add a `.recovering` branch to `startStopButton` so the control reads as busy rather than ready.

Logue/Engine/TranscriptRealignment.swift
```diff
@@ -0,0 +1,201 @@
...
+    ///   zero while its lines are stamped minutes in. Without this shift every word of session two
+    ///   lands in session one's lines and overwrites them.
+    static func realign(
+        live: [TranscriptSegment],
+        words: [TimedWord],
+        sessionStart: TimeInterval = 0
+    ) -> [TranscriptSegment] {
+        guard !live.isEmpty, !words.isEmpty else { return live }
+
+        let words = sessionStart == 0 ? words : words.map { word in
```
shanforge, line 91: **Major** — the shift fixes the reported case, but `live:` is still the whole meeting and the nearest-segment fallback has no distance cap, so a second session can still overwrite the first's text — now all of it into one line.

Record 5 minutes and stop (segments end at 300s). Record again: `timeOffset = 301`. Session two produces no live segments — the speech gate admits nothing, or the transcriber errors, which is `MicrophoneSpeechGate`'s documented failure mode — so `meeting.segments` is still session one only. The batch pass succeeds, so `batchSegments` is non-nil and `lastBatchWords` is non-empty and the realign block runs. Every word shifts to 301…301+T2, past the end of every segment in `live`; `indexOfSegment(containing:)` finds no container and the loop at lines 170-186 returns the nearest, which is session one's last line for all of them. That line's text becomes the entire concatenated second session, and five minutes of correct transcript reads as the wrong session.

Bound the input instead of only shifting the words: realign against `live.filter { $0.startTime >= sessionStart }` (mapping the results back by id), so a word can only ever land in a line from its own session.

Logue/Engine/TimelineAudioConverter.swift
```diff
@@ -0,0 +1,79 @@
+import AVFoundation
+import Foundation
+import os.log
+
+/// Converts capture buffers to the 16 kHz mono the session timeline is built from.
+///
+/// Lifted out of `DiarizationManager` so it can be tested without a microphone. It is worth testing:
+/// the saved recording is written straight from the capture tap, so a conversion that quietly yields
+/// silence leaves an audible file and an empty transcript, and nothing in between reports a fault.
+final class TimelineAudioConverter {
```
shanforge, line 10: **Major** — unchanged from the last review. `git grep TimelineAudioConverter` at this head returns this declaration and four hits in `LogueTests/TimelineAudioConversionTests.swift`; `Logue/Engine/DiarizationManager.swift:367` still calls its own private `convertBufferToFloatArray`.

So the suite whose doc comment says it guards "a conversion that quietly yields silence" guards a type that never runs. Introduce a regression in the shipped converter's capacity or priming and all four tests stay green. The two have already diverged in both directions: this copy adds `+ 64` frames of margin the shipped one lacks, and the shipped one sets `converter.primeMethod = .none`, which this lacks — the two settings that decide the exact failure mode being tested.

Either make `DiarizationManager` call it (keeping `primeMethod = .none`), or delete the file and its tests.

Logue/Engine/TranscriptSentenceMerge.swift
```diff
@@ -0,0 +1,45 @@
...
+    static func endsSentence(_ text: String) -> Bool {
+        let trimmed = text.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard let last = trimmed.last else { return false }
+        return ".!?…".contains(last)
+    }
+
+    /// Whether `next` should be folded into `previous`.
+    static func shouldMerge(previous: TranscriptSegment, next: TranscriptSegment) -> Bool {
+        // Never merge across speakers: a reply is a new line however it is punctuated.
+        guard previous.speakerLabel == next.speakerLabel else { return false }
```
shanforge, line 60: **Major** — unchanged from the last review; the guard is still inert wherever `shouldMerge` actually runs. `MeetingStore.appendSegment` is the sole caller and runs during recording, where `SpeechTranscriberEngine` constructs every segment with `speakerLabel: nil`. So this is always `nil == nil` and always passes. With one transcriber fed by both the mic and the system tap, a local utterance ending without terminal punctuation followed 0.4s later by a remote reply satisfies all four checks and becomes one line carrying one id, one speaker and one `audioSource`; blocks are cut by time only, diarization adds labels without splitting, and drag-to-reassign was removed in this PR, so nothing can undo it.

One correction to the fix suggested last time: `previous.audioSource == next.audioSource` will not discriminate as written. `SpeechTranscriberEngine` leaves it nil, and `startRecording` stamps `.system` on *every* segment when `recordingMode == .onlineMeeting` — so both sides are equal in both modes. The source has to be tagged per buffer before that comparison can mean anything.

Logue/Services/MeetingStore+Transcript.swift
```diff
@@ -0,0 +1,47 @@
...
+    /// looking at moves.
+    func updateSegments(_ segments: [TranscriptSegment], for meetingID: UUID) {
+        guard let index = meetingIndex(for: meetingID) else { return }
+        meetings[index].segments = segments
+        saveMeeting(id: meetingID)
+    }
+
+    /// Puts back what a session held in memory when the app stopped.
+    ///
+    /// The checkpointed segments replace whatever is on the meeting rather than appending to it: a
```
shanforge: **Minor** — this doc now says the opposite of what the function does. It states the checkpoint "replaces whatever is on the meeting rather than appending to it" because "appending would duplicate every line that had already been saved", directly above a body that never replaces and does append, and whose safety rests on precisely the id-keyed dedupe the comment denies is possible. Per CLAUDE.md a comment that no longer matches the code below it is worse than none. Rewrite it around the reach comparison and the union.

Logue/Services/RecordingSessionManager+Recovery.swift
```diff
@@ -0,0 +1,205 @@
...
+    ///
+    /// Returning false means we do not know what the library holds, and nothing may be deleted on
+    /// the strength of a meeting appearing to be absent.
+    private func waitForMeetingStore() async -> Bool {
+        let deadline = ContinuousClock.now + AppConstants.Delays.meetingStoreLoadTimeout
+        while ContinuousClock.now < deadline {
+            if MeetingStore.shared.isLoaded {
+                return true
+            }
+            try? await Task.sleep(for: AppConstants.Delays.meetingStoreLoadPoll)
```
shanforge: **Minor** — under cancellation this busy-spins the main actor for the full 30-second budget. `recoverInterruptedSessions()` is driven by a `.task` on the root view, so closing the window during launch cancels it; `try? await Task.sleep` then throws `CancellationError` immediately on every iteration and swallows it, and the `while ContinuousClock.now < deadline` loop re-enters at full speed until the wall clock catches up. Add `guard !Task.isCancelled else { return false }` inside the loop.

Logue/Services/RecordingSessionManager+Recovery.swift
```diff
@@ -0,0 +1,205 @@
...
+        logger.info("Recovered meeting \(meetingID)")
+    }
+
+    /// Rebuilds the capture sources from what the checkpoint wrote down.
+    ///
+    /// The placements are the part that matters. They are what says where each stretch of each file
+    /// belongs on the meeting, and they cannot be worked out afterwards from the files themselves.
+    private func captureSources(from checkpoint: RecordingCheckpoint) -> CaptureSources {
+        var sources = CaptureSources()
+        guard let directory = try? InProgressRecordingStore.directory(for: checkpoint.meetingID) else {
```
shanforge: **Minor** — unchanged from the last review. A silent `try?` on a directory-creating call, which CLAUDE.md forbids for anything but `Task.sleep`. The failure it swallows is exactly the one the `.task` comment in `Logue/App/LogueApp.swift` says this code has to tolerate — a container `SandboxContainerMigrator` just moved — and it degrades to an empty `CaptureSources`, so `persistRecordingAudio` returns `.none` and the meeting is recovered with its transcript and no audio, with nothing in the log saying why. `do/catch` and log the error.

Logue/Services/RecordingSessionManager+Diarization.swift
```diff
@@ -137,16 +137,34 @@ extension RecordingSessionManager {
...
+        // The transcript the user watched being written is the transcript they keep.
+        //
+        // Replacing it with the batch result was more accurate word for word, but it re-cut every
+        // sentence: a different number of blocks, text merged and split, and the reader's place
+        // lost — all arriving silently, a minute after they stopped recording. Speaker identity does
+        // not require it. Sortformer produces time ranges, not text, so the labels below apply to
+        // the live segments perfectly well, and the transcript never changes under anyone.
+        //
+        // `batchSegments` is still computed: it is what the speaker timeline is derived from, and
+        // `heardDuration` still bounds how much of the session that timeline may speak for.
```
shanforge: **Minor** — unchanged, and this commit made it matter more. `heardDuration` is bound seven lines above and never read (a live compiler warning), so this sentence describes a bound that does not exist. It is non-nil precisely on the disk route where the saved file is shorter than the session — the route this commit newly feeds into realignment by populating `lastBatchWords` — so a batch word near the truncated file's end can be placed by the uncapped nearest-segment fallback onto a live line the pass never heard. Either bound `live:` by the heard stretch, or delete the binding and this sentence together.

Logue/Services/MeetingStore.swift
```diff
@@ -414,7 +414,18 @@ final class MeetingStore: MeetingRepository, MeetingSegmentManager, MeetingSpeak
     /// when recording stops.
     func appendSegment(_ segment: TranscriptSegment, to meetingID: UUID, persistImmediately: Bool = true) {
         guard let index = meetingIndex(for: meetingID) else { return }
-        meetings[index].segments.append(segment)
+
+        // The transcriber finalises when it is confident, not when a sentence ends, so a thought
+        // arrives split across two lines. Joined here, as it arrives, the stored transcript is made
+        // of sentences and everything downstream inherits that rather than working around it.
+        if let last = meetings[index].segments.last,
+           TranscriptSentenceMerge.shouldMerge(previous: last, next: segment)
```
shanforge: **Minor** — unchanged from the last review. `startRecording` sets `timeOffset = lastEnd + 1.0` on a resumed recording, so the first new segment starts 1.0s after the previous session's last line — inside `TranscriptSentenceMerge.maximumGap` of 2s. When the earlier session left every `speakerLabel` nil (diarization found nothing), all four checks pass and session two's opening words are folded into session one's closing line, which keeps session one's `startTime`. `TranscriptReplacement`'s `startTime < sessionStart` filter, playback seek and bookmark anchoring then all treat those words as session one. Skip the merge when `next.startTime >= sessionStart`.

Logue/Engine/TranscriptRealignment.swift
```diff
@@ -0,0 +1,201 @@
...
+    /// A piece that begins a new word carries a leading space; anything else continues the word
+    /// before it. The joined word spans from its first piece's start to its last piece's end, so it
+    /// still lands in exactly one line.
+    static func words(fromTokens tokens: [TimedWord]) -> [TimedWord] {
+        var words: [TimedWord] = []
+        var pending: [TimedWord] = []
+
+        func flush() {
+            guard let first = pending.first, let last = pending.last else { return }
+            let text = pending.map(\.text).joined()
```
shanforge, line 53: **Minor** — unchanged from the last review. `words(fromTokens:)` accepts `\u{2581}` as a word-start marker at line 63 but nothing ever strips it: this join is verbatim, and `trimmingCharacters(in: .whitespacesAndNewlines)` in `realign` does not remove U+2581. If Parakeet emits the marker form the transcript gets literal `▁theCUBE` with no word separation; if it never does, the branch at line 63 is dead. `sentencePieceMarker` asserts only `joined.count == 1`, so it passes either way. Strip the marker here and assert the resulting string.

Logue/Engine/TimelineContributionLog.swift
```diff
@@ -0,0 +1,67 @@
...
+    /// A buffer that could not be converted never reaches the timeline at all, which is a different
+    /// failure from a silent one and worth telling apart.
+    mutating func recordDroppedConversion(from source: AudioSource, logger: Logger) {
+        let count = (droppedConversions[source] ?? 0) + 1
+        droppedConversions[source] = count
+        if count == 1 || count % 200 == 0 {
+            let name = String(describing: source)
+            logger.warning(
+                """
+                Timeline: dropped \(count, privacy: .public) buffer(s) from \(name, privacy: .public)                 — conversion produced nothing
```
shanforge: **Minor** — unchanged from the last review. The literal 17-space run before the em dash means this prints as `… buffer(s) from microphone                 — conversion produced nothing`. The sibling message at lines 39-42 uses `\` continuation correctly.

Logue/Views/Meeting/TranscriptSpeakerBlockView.swift
```diff
@@ -0,0 +1,398 @@
...
+    @State private var isHovered = false
+    /// The paragraph the pointer is in, if any.
+    @State private var hoveredGroupID: UUID?
+    /// The line whose bookmark picker is open.
+    ///
+    /// An identity rather than a flag: every marked line attaches a popover, and binding them all
+    /// to one boolean made SwiftUI anchor the picker to whichever button happened to render last —
+    /// so it opened at the bottom of the transcript instead of beside the button pressed.
+    @State private var bookmarkingSegmentID: UUID?
+    @State private var showBookmarkAdded = false
```
shanforge: **Minor** — unchanged from the last review. Written at lines 389 and 392 and read nowhere, so the confirmation `AppConstants.Delays.bookmarkConfirm` is named for never renders — either the state or the missing UI is a bug; as it stands it is dead. `isHovered` in this file is in the same position now that `hoveredGroupID` handles hover chrome.

shanforge (changes requested):
All four blocking findings are fixed at the cause, and I walked each one at this head rather than reading the diff: the session offset is applied and pinned by a test that would not compile at the reviewed commit, the store wait is bounded and leaves everything on disk when it times out, `filterWithFileVAD` now delegates to the same majority rule the in-memory pass uses, and the checkpoint is encrypted with a test asserting the bytes on disk. That is the important half.

The other half did not land. Four of seven majors are fixed; two were not touched at all; one — the block cache — only looks fixed, because the key gained a text signature but nothing new calls the function that evaluates it, so the transcript still stops updating mid-recording. Every one of the eight minors is unchanged. Three of the fixes also brought a new defect with them: the mute flag is never cleared between sessions, an unreadable checkpoint is still read as an absent one and deletes the audio, and the new `.recovering` state refuses a record press without saying anything.

## Fixes verified

**Blocking**

- Realignment ignores the session offset — **fixed**. `realign` takes `sessionStart` and shifts every word; the call site passes the `timeOffset` captured before reset. A resumed session's words can no longer land inside session one's lines. (One residual case survives — see Major below.)
- Recovery runs before the meeting store has loaded — **fixed**. `waitForMeetingStore()` polls `isLoaded` against a 30s `ContinuousClock` deadline and returns without touching anything on timeout. (The flag it waits on is weaker evidence than the fix assumes — see Major below.)
- VAD over-filter fix applied to one filter only — **fixed**. `filterWithFileVAD` maps its regions into `BatchTranscriptFilter.SpeechRegion` and returns `BatchTranscriptFilter.filter`, so the `kept.count * 2 >= segments.count` rule now governs the disk and recovery route too, and `distrustedVAD` is logged.
- Checkpoint written as plaintext — **fixed**. `EncryptionManager.encryptCodable` / `decryptCodable`, with `writtenCheckpointIsEncrypted` asserting the raw file does not contain the transcript.

**Major**

- Device drop un-mutes a muted mic — **fixed for the reported case**, with a new one introduced (`isMicMuted` is never cleared). The duplicated resume sequence was collapsed as asked.
- `restoreRecoveredSession` discards already-saved transcript — **fixed**. Reach comparison plus an id-keyed union; the saved copy wins on collisions.
- Recovery/record race — **fixed**. `.recovering` is held for the whole loop with a `defer` back to `.idle`, and `startRecording` refuses on it. No exhaustive `switch` over `RecordingState` exists, so the new case breaks no call site.
- Disk route never populates `lastBatchWords` — **fixed**. Built from `result.tokenTimings` exactly as the in-memory route does, and reset to `[]` on the no-timings path so a stale set cannot leak.
- Block cache misses text-only changes — **not fixed**. The key gained `segmentTextHash`; the trigger list did not.
- `TimelineAudioConverter` has no production caller — **not fixed**. File untouched.
- Cross-speaker merge guard is inert — **not fixed**. File untouched.

**Minor** — none of the eight was addressed: the silent `try?` in `captureSources`, the unused `heardDuration`, the resume-gap merge, the SentencePiece marker, the space run in `TimelineContributionLog`, the dead code, the recovery test gap, and the four comments to delete.

## Major

**None of the two tests this commit added ever runs.** `.github/workflows/ci.yml` invokes `xcodebuild test` with exactly two `-only-testing:` flags — `LogueTests/WhatsNewGateTests` and `LogueTests/AppVersionTests` — and neither `TranscriptRealignmentTests` nor `RecordingCheckpointTests` is among them. The file is byte-identical to `main` and this PR does not touch `.github/` at all. So the evidence that two of the four blocking fixes hold is green by never having run: revert the `sessionStart` shift or swap `encryptCodable` back for `JSONEncoder` tomorrow and CI stays green. The workflow's own comment three lines above the flags says "A guard nothing runs is not a guard — list a suite here when you add one that needs no model", and this PR adds twelve such suites. Add at least `TranscriptRealignmentTests`, `BatchTranscriptFilterTests`, `DeviceLossPolicyTests`, `TranscriptSentenceMergeTests` and `RecordingRecoveryTests`. Note before adding `RecordingCheckpointTests`: three of its cases now go through `EncryptionManager`, so it needs a usable Keychain on the runner — check that before wiring it up.

## Minor

**Dead code, unchanged from the last round** (these lines are outside the diff, so they cannot be commented on inline): `Logue/Services/ConferencingAppDetector.swift:4` still has no reference anywhere but its own declaration; `Logue/Services/MeetingStore.swift:212` `reassignSegmentSpeaker` has had no caller since drag-to-reassign was removed in this PR; `Logue/Views/Meeting/TranscriptTimelineView.swift:15` `SpeakerBlock.formattedStartTime` has none now the block header prints no time.

**Comments to delete, unchanged.** `Logue/Views/Meeting/TranscriptTimelineView.swift:69-71` still documents "Starts a new block when the speaker changes" six lines above the comment explaining that speaker splitting was removed and the only branch left is `if timeGap`. `Logue/Views/Meeting/TranscriptTimelineView.swift:368` is a tombstone for a type that moved to `TranscriptSpeakerBlockView.swift`. `Logue/Views/Meeting/TranscriptSpeakerBlockView.swift:57-61` still floats between declarations, describing a method 135 lines below.

**Recovery still has no test.** `LogueTests/RecordingRecoveryTests.swift` is unchanged — its five cases call `TranscriptReplacement.merged`, `RecordingCheckpoint.coveredDuration` and `CaptureSegmentTimeline`, none of which is new. Nothing constructs an in-progress directory with audio plus a checkpoint and runs `recoverInterruptedSessions()`, so the store wait, the `.recovering` hold, the meeting-not-found deletion and the new merge in `restoreRecoveredSession` — every decision that determines whether a crashed meeting keeps its transcript and its audio — are verified by nothing. The merge in particular is a blocking-fix decision that shipped with no test at all, and it is a pure function of two segment arrays, so it is the cheapest one to cover. `MicrophoneSpeechGate.admit` still has no test either.

## Compatibility with main

Clean, and unchanged from the last round. The fix commit adds no `Codable` field, renames no key, changes no enum raw value, and touches no `~/Logue` markdown, frontmatter or `_space.md` path. `RecordingCheckpoint`'s shape is unchanged — only its on-disk encoding — and the type is new on this branch, so no released build ever wrote one; a plaintext checkpoint left by an earlier build of *this branch* will fail to decrypt, which matters for dogfooding rather than for users, but it lands on the delete path (see the `RecordingCheckpoint.swift` comment). `RecordingState` gaining `.recovering` breaks no call site: every reader is an `== .x` comparison and there is no exhaustive `switch` over the enum anywhere in the tree. No new force unwrap, force cast, `[0]` on a FileManager array, `url.absoluteString` log, unwrapped user content in a prompt, or inference outside `complete()`/`completeStream()`.

CI at this head: Build pass, Lint Check pass — but see the note above on what those tests actually cover. No visual evidence has been supplied for the recovery UI or the live transcript, and neither is settleable by static review.

shanforge (commented):
Zero blocking and zero major findings, CI green, and the round on `ef6c62a1` was the first on this branch to introduce no new defect of its own — three lenses attacked the scoping fix and none could reach a failure. All sixteen minors are addressed in `8a70c5e`. This is a comment rather than an approval because there is still no screenshot or recording of a change that is almost entirely about what the user sees while recording and recovering.

## Two of the minors were worth more than their tier

**A second session's first words still extended the first session's last line.** Record an in-person meeting, stop, then press record again on the same meeting. `startRecording` sets the new offset to `lastEnd + 1.0`, which is inside `TranscriptSentenceMerge.maximumGap` of 2s, and during recording every segment carries `speakerLabel: nil` — so `previous.speakerLabel == next.speakerLabel` is `nil == nil` and passes. The system tap is not running in this scenario, so the `mayCarryTwoSpeakers` gate added last round does not fire either. The result was session two's opening words welded onto a line session one produced minutes or hours earlier, in one segment with one id that nothing downstream could split. `shouldMerge` now takes `sessionStart` and refuses to reach back across it. Verified by deleting the guard and watching the new case fail.

**`isDiarizing` and `diarizingMeetingID` were two fields holding one fact**, written at seven sites across three files with nothing keeping them in step — and `processDiarization` was already out of step: it set the flag with the meeting id sitting in its own parameters and never set the id. Since the scoped predicate is `diarizingMeetingID == meetingID`, that site silently turned every scoped spinner off. Now one stored field with `isDiarizing` derived from it, so the failure is not expressible. `MeetingWorkspaceView` — the last global reader, and the one the verifier correctly demoted from major after finding that `awaitPreviousPostRecordingTask()` makes it unreachable — is scoped with the others.

Two more that are user-visible:

- **Under cancellation the store-load wait span the main actor flat out.** `.task` on the root view cancels `recoverInterruptedSessions()` when the window closes during launch. `Task.sleep` then threw immediately on every iteration and `try?` swallowed it, leaving a loop with no suspension left in its body for the remainder of the 30-second budget — a beachball on quit, precisely when the store is slow, which is when this code runs at all.
- **A recovery that could not reach its working directory said nothing.** The silent `try?` degraded to empty capture sources: no audio attached, `reprocessRecoveredAudio` returning at its guard, and the meeting coming back with only its checkpointed transcript. `SandboxContainerMigrator` moving the container is both the case recovery exists to tolerate and the case that makes that lookup throw.
- **The SentencePiece word marker was accepted as a word start and never stripped.** U+2581 is not whitespace, so `trimmingCharacters(in: .whitespacesAndNewlines)` does not remove it: a realigned line would have read `▁the▁agent`, marker visible and no word separation, written into the transcript a minute after recording stopped. It is a boundary, so it now becomes a space.

The rest were comments that had outlived their code — `segmentTextHash`'s doc still described the length-based hash the fix had to remove and contradicted the inline comment two lines below it; `restoreRecoveredSession`'s doc claimed a wholesale replace and denied the id-keyed dedupe its body depends on; `+AudioStream`'s header claimed an ordering guarantee covering the file, which is written on the audio thread before the buffer ever reaches that consumer — plus three `Extension-visible:` markers that named only `+DeviceLoss` after the split, an unread `heardDuration` binding under a comment asserting a bound this path does not apply, a `showBookmarkAdded` flag written twice and read nowhere so bookmarking a transcript line acknowledged nothing, and a 17-space run that printed a log message with a ragged gap.

## Compatibility with main

Clean. No `Codable` field, frontmatter key or enum raw value changes, so meetings written by 1.0.x still decode. `shouldMerge`'s new `sessionStart` parameter is defaulted, so no existing caller changes behaviour; the one production caller now supplies `RecordingSessionManager.timeOffset`, which is the session's own start on the meeting timeline.

## Verification, and one thing worth fixing about it

CI at `8a70c5e`: Build pass 8m50s, Lint Check pass 24s. The Build job's test step passed — but it runs an `-only-testing:` allow-list, so what actually executed was **140 tests in 17 suites**. Run locally against the whole non-model target, this branch passes **1160 tests in 104 suites**, so nothing here breaks an existing suite. CI does not show that: a regression in any of the other eighty-seven suites would pass green.

That is worth closing before merge rather than after. `scripts/test-no-llm.sh` on the tasks branch is the same idea as a *deny*-list, and its CI step runs 1347 tests in 108 suites for the same wall-clock cost. Adopting it here would make this PR's green mean what it appears to mean. (`-skip-testing:` takes a class name, never a directory, which is the trap that script exists to avoid.)

## What is needed to approve

Only the click-through.

1. **A second session does not land in the first session's lines.** Record 2 min mic-only, stop, let processing finish, then record 1 min more on the same meeting. Correct: the second session's words appear as new lines after the first session's. Wrong: concatenated into the last line of the first, or the first session's lines rewritten. This is the defect fixed above and the one most worth a screenshot.
2. **Mute mid-recording.** Record, mute the mic 30 s, unmute, record 30 s more, stop. Correct: the transcript covers the whole duration with a gap where the mute was, not a truncated recording. Then start a new recording — the mic must show unmuted.
3. **Two remote speakers back to back**, with system audio captured and a short pause between them. Correct: two lines, not one carrying both.
4. **Recovery does not spinner over a finished meeting.** Start recording, force-quit mid-session, relaunch. While "Recovering meeting…" is showing, open a *different*, already-finished meeting. Correct: its Smart Minutes and speaker list render. Wrong: either shows a spinner.
5. **A queued auto-capture survives a wait but not a refusal.** (a) With a recovery in progress, open a calendar event set to auto-record — correct: it starts once recovery finishes. (b) Deny microphone permission, then open that meeting twice — correct: the second open does not start recording.
6. **The toolbar names what is holding things up.** During recovery and during stopping the record button is disabled and labelled accordingly, and the empty-state "Start Meeting" action is absent.
7. **Edge states.** Fresh install with no checkpoint file: no recovery attempt, no error. Seed data present: recovery must not run.

shanforge (approved):
Approving. Zero blocking and zero major findings after seven rounds, CI green, tests covering the new logic — and the visual-evidence gate is now met: I drove this branch's build by hand rather than reasoning about it, and the reviewer confirmed the click-through.

## What I exercised on a build of `8a70c5e`

Audio source was a two-host interview podcast played through system audio, so the capture path, the transcriber and the diarizer all had real speech to work on.

- **System-audio capture and live transcription.** Recording started, the level meter and system-audio indicator were live, and the podcast transcribed in real time — 68 segments over the first 3-minute session, auto-titled from its own content.
- **Diarization, scoped to the meeting it belongs to.** The pass ran as "Identifying speakers…" in that meeting's toolbar and resolved to per-line speaker labels (S1–S4) with an Attendees breakdown. The panels that this PR rescoped rendered against the meeting being viewed, not against a global flag.
- **The checkpoint is genuinely encrypted at rest.** This was a blocking finding two rounds ago, so I checked the artefact rather than the code: with a session interrupted mid-recording, `checkpoint.json` on disk is ciphertext — no readable JSON keys, no field names. The fix holds against real data.
- **Recovery of an interrupted session.** I killed the app mid-recording with ~1 minute of audio in flight (mic `.wav` and system `.caf` both on disk beside the checkpoint). On relaunch the session was rebuilt into a saved meeting with its transcript, a generated summary and action items, and the in-progress working directory was cleaned up afterwards.
- **A second session on an existing meeting.** Pressing record again on a meeting that already held 3:57 of transcript extended it correctly — the meeting grew to 6 minutes and 125 segments, with the new session's lines appended rather than replacing what was there.
- **A session that captured only silence added no transcript at all.** Six minutes of recording with the audio source paused produced zero new segments while the duration still advanced — the voice-activity gate behaving exactly as intended, and a useful negative control.

## What I did not isolate

I could not stage the one side-by-side moment cleanly: a *second* meeting rendering its own Smart Minutes while a *first* meeting's diarization is still in flight. On short recordings the pass completes faster than the meetings can be switched, and a longer attempt was interrupted. I am approving over it because the code path is settled from the other direction — all three formerly-global readers now take the meeting id, the last one was converted in this commit, and the previous round's verifier independently walked `awaitPreviousPostRecordingTask()` and could not construct a state where a live session sees a stale flag. Worth a glance next time someone has a long meeting processing.

## One incidental observation, filed rather than claimed

If macOS restores the app with **zero windows**, recovery never runs: it is driven by the root view's `.task`, so an interrupted recording sits untouched on disk until a window exists. I hit this only because a call closed the window mid-test, and forcing a window open recovered the session correctly. Not a regression in this PR — the same structure exists on `main` — but the gap between "the app is running" and "recovery happens" is real and cheap to close.

## Verification

CI at `8a70c5e`: Build pass, Lint Check pass. Locally this branch passes the whole non-model target — **1160 tests in 104 suites**. Note that CI itself runs an `-only-testing:` allow-list, so the green check covers **140 tests in 17 suites**; adopting `scripts/test-no-llm.sh` here, as the tasks branch does, would make that signal mean what it appears to mean. Not a merge blocker, but the next regression outside those 17 suites will pass green.

PR #96, opened 2026-09-14
feat: choose live transcript language independently of the Mac

### LINKED ISSUE #95: [Feature Request] Allow manual selection of audio transcription language / locale

Currently, the live/real-time transcription stream appears to be hardcoded or initialized to English. 

When attending non-English meetings (e.g., in German), the live transcript displays garbled phonetic noise in real time. While the post-processing engine (e.g., the final pass after recording finishes) transcribes German accurately, the real-time live view is unusable during the meeting itself. Since there is no option to configure the input language, users cannot follow live transcriptions in languages other than English.

### Proposed solution

Expose a setting to select the language/locale for the real-time transcription engine (e.g., `de-DE`, `en-US`, `fr-FR`). 

Configuring the live stream model to the correct locale will ensure that both the live transcript during the meeting and the final post-processed transcript remain accurate and readable.

### Alternatives considered

- Tested if changing macOS per-app language settings in System Settings affects the live engine locale, but it remains English.
- Relying solely on the post-processed transcript works for reviewing notes later, but completely defeats the purpose of following along with live meeting captions in real time.

### Privacy check

dannyneumann:
## Summary

Fixes live captions being forced to English (`en-US`) even when the meeting is in German (or another language). Apple's `SpeechTranscriber` is one locale per session; we were treating “no locale” as English and then falling back to English again when an explicit language failed to load.

- **Default:** Settings → General → Meetings → “Default recording language”. New meetings pick this up. This lives next to other meeting defaults, not under Models (those rows are LLMs, not ASR).
- **Per meeting:** the Start control is a `ControlGroup` — Start plus a language menu — so you can override the default before recording. Language is locked while recording.
- **Smart Minutes** (and title generation) follow the recording language, including German markdown headings.
- **Regenerate / Generate Title** stay disabled while `LLMEngineStatus.shared.isBusy` so overlapping inference cannot fire.

Closes bitwize-ai/Logue#95

## Test plan

- [ ] Settings → General → Meetings: set default to German, create a new meeting, confirm Start shows German
- [ ] On an existing meeting, open the Start language menu and switch to English, then start recording — captions should follow that override, not the General default
- [ ] Start a German meeting: live captions should be readable German, not phonetic English noise
- [ ] Stop recording and generate Smart Minutes: headings and body should be German
- [ ] While minutes are generating, Regenerate / Generate Title should be disabled
- [ ] Models tab is unchanged (on-device / connected models only)

Made with [Cursor](https://cursor.com)


dannyneumann:
Follow-up pushed: German live captions were blocked because Apple’s `transcription.de` asset stays in “Not Installing”. The mic now records immediately, the hanging download is skipped, and captions fall back to on-device dictation (`SFSpeechRecognizer`). Your voice and system audio should both appear in the transcript.

Logue/Engine/SpeechTranscriberEngine.swift
```diff
@@ -225,101 +251,318 @@ final class SpeechTranscriberEngine {
...
+        startLegacyTask(recognizer: recognizer, request: request)
+        logger.info("System speech recognizer started for \(locale.identifier)")
+    }
+
+    fileprivate func makeLegacyRequest(recognizer: SFSpeechRecognizer) -> SFSpeechAudioBufferRecognitionRequest {
+        let request = SFSpeechAudioBufferRecognitionRequest()
+        request.shouldReportPartialResults = true
+        request.taskHint = .dictation
+        if recognizer.supportsOnDeviceRecognition {
+            request.requiresOnDeviceRecognition = true
```
shanforge: **Blocking — privacy.** `requiresOnDeviceRecognition` is only set when `supportsOnDeviceRecognition` is true. Otherwise the request goes out with the default (`false`) and `SFSpeechRecognizer` streams the meeting audio to Apple's servers.

Failure scenario: German selected, the on-device dictation model for `de-DE` isn't downloaded (so `supportsOnDeviceRecognition == false`), and the SpeechAnalyzer path fails. That's exactly the case this fallback exists for. The whole meeting is then transcribed server-side with no notice to the user. That breaks the app's core promise that nothing leaves the laptop by default. It also makes `NSSpeechRecognitionUsageDescription` ("All processing happens on-device") untrue.

Fix: always set `requiresOnDeviceRecognition = true`. If on-device isn't supported, fail setup and let the existing "Live captions could not start" notice show.

Logue/Services/MeetingStore+AI.swift
```diff
@@ -192,13 +198,29 @@ extension MeetingStore {
...
+        guard let meeting = meetings.first(where: { $0.id == meetingID }),
+              let docID = meeting.summaryDocumentID,
+              let index = DocumentStore.shared.documents.firstIndex(where: { $0.id == docID })
+        else { return }
+        let markdown = meeting.smartMinutes != nil
+            ? meeting.smartMinutesMarkdown()
+            : (meeting.summary ?? "")
+        guard !markdown.isEmpty else { return }
+        var document = DocumentStore.shared.documents[index]
+        document.body = markdown
```
shanforge: **Blocking — overwrites user documents.** `summaryDocumentID` doesn't only point at an auto-created notes doc. `SaveSummaryToDocumentSheet` also sets it when the user picks **"Append the summary to an existing document"** (`updated.body += separator + markdown`). This line then replaces that document's *entire* body with the minutes.

Failure scenario: the user appends a summary to their "Project X" doc, which holds pages of their own writing. They record a second session on the same meeting. `PostRecordingPipeline` calls `generateAISummary` automatically (so does `ScheduledTaskManager`), and every word the user wrote in "Project X" is replaced. With plain-markdown storage on, the rewrite goes straight to their `.md` file. There's no undo and no Trash.

Fix: drop the automatic rewrite. If refreshing is wanted, make it an explicit user action that only touches a document Logue created and the user hasn't edited since.

Logue/Services/RecordingSessionManager.swift
```diff
@@ -417,20 +400,29 @@ final class RecordingSessionManager {
...
             diarizationManager = nil
             recordingState = .idle
             return .captureFailed
         }
 
+        do {
+            let language = TranscriptionLanguage.resolved(fromStored: meeting.transcriptionLanguage)
+            let locale = language.resolvedSpeechLocale()
+            recordingLocale = locale
+            try await engine.setup(locale: locale, allowLanguageFallback: language == .auto)
```
shanforge: **Blocking — two defects on this `await`.**

1. **Live timestamps shift early, and the first seconds are never captioned.** Setup now runs *after* `startMicrophoneRecording`. The buffer consumer drops audio while `speechEngine` is nil (`self?.speechEngine?.streamAudio`). The engine's `sessionStartDate = Date()` (wall-clock timing, `useAudioDrivenTiming` is `false`) is also stamped only once setup finishes. So every caption is timed from caption start, not from the start of the audio file. With the 8 s timeout plus the legacy fallback, a line spoken at 0:20 in the file is stored at about 0:12. The batch pass then pours words into lines by time (`TranscriptRealignment.realign`), so the text lands in the wrong lines. Playback, bookmarks and speaker alignment are off by the same amount. This breaks the rule that the saved audio file is the meeting's timeline.

2. **Stop during setup leaks a running engine.** Nothing re-checks `recordingState` after the await. If the user presses Stop in the ~8 s window, `teardownAudioPipeline` finds `speechEngine == nil` and finishes nothing. Setup then returns and assigns `speechEngine = engine` on an idle manager. That engine's `SFSpeechRecognizer` restarts itself on every final/error (see the comment on `restartLegacyRecognition`) and is never finished. The next `startRecording` overwrites the reference without calling `finish()`.

Fix: bring captions up without delaying capture, but time them from the capture clock (for example, the stream's frame count from session start), not from when setup finished. After the await, `guard recordingState == .recording, currentMeetingID == meetingID`, and `await engine.finish()` otherwise.

Logue/Engine/SpeechTranscriberEngine.swift
```diff
@@ -225,101 +251,318 @@ final class SpeechTranscriberEngine {
...
+        }
+
+        throw SpeechTranscriberError.assetInstallFailed(
+            locale: requested.identifier,
+            underlying: "No installed live-caption model for this language."
+        )
+    }
+
+    fileprivate func readySpeechTranscriber(requested: Locale) async -> SpeechTranscriber? {
+        let installed = await SpeechTranscriber.installedLocales
```
shanforge: **Blocking — the speech model is never downloaded any more.** On `main`, `ensureModel` called `AssetInventory.assetInstallationRequest(...)` → `downloadAndInstall()`. This PR removes every call to either API. The analyzer path now only uses locales already in `installedLocales`.

Failure scenario: a fresh install (or a newly selected language, English included) where the SpeechTranscriber asset hasn't been fetched yet. `readySpeechTranscriber`, `readyDictationTranscriber` and `englishCaption` all return nil, so setup throws. The language is then cached as failed for the process, and every meeting goes to `SFSpeechRecognizer`, which with the line-469 issue can mean server-side recognition. Before this PR, the same user got the on-device model downloaded. `downloadProgress` is now never set, and `assetInstallFailed`'s "Check the network and try Start again" points at a download that doesn't happen.

Fix: keep installing the asset for languages where it works. Skip it only for the specific asset that is stuck in "Not Installing", and do that without blocking capture.

Logue/Engine/SpeechTranscriberEngine.swift
```diff
@@ -225,101 +251,318 @@ final class SpeechTranscriberEngine {
...
+            }
+            return
+        }
+        if let error {
+            if isFinishing {
+                completeLegacyFinishWait()
+                return
+            }
+            logger.error("System speech recognizer error: \(error.localizedDescription, privacy: .public)")
+            restartLegacyRecognition()
```
shanforge: **Major — restart loop with no backoff.** Any error that isn't during finish restarts recognition immediately, and so does every `isFinal` (line 503). Errors that won't go away cause a hot loop: speech recognition not authorized (nothing on this path checks `SFSpeechRecognizer.authorizationStatus()` or requests it), or on-device recognition unavailable for the locale. The loop creates a new task, gets an immediate error, and hops to the main actor again, for the whole meeting (and, with the leak on `RecordingSessionManager.swift:418`, after it).

Fix: check authorization in setup and fail if it's missing. Restart only on the known "session limit reached" final, and use the project's exponential backoff (50 ms doubling, 400 ms cap) with a retry limit for errors.

Logue/Engine/SpeechTranscriberEngine.swift
```diff
@@ -75,115 +79,135 @@ final class SpeechTranscriberEngine {
...
-
-        analyzerFormat = await SpeechAnalyzer.bestAvailableAudioFormat(
-            compatibleWith: [transcriber]
-        )
-
-        guard analyzerFormat != nil else {
-            logger.error("No compatible audio format found")
-            throw SpeechTranscriberError.invalidAudioFormat
+        do {
+            try await withThrowingTaskGroup(of: Void.self) { group in
```
shanforge: **Major — this timeout can't cut anything short.** `withThrowingTaskGroup` waits for every child before it returns, even when the body throws. When the sleep child throws, the group cancels `setupAnalyzer` and then waits for it anyway. If the call that's stuck ignores cancellation (the "Not Installing" hang this was added for), the timeout doesn't fire until that call returns. `RecordingSessionManager.awaitPreviousPostRecordingTask` has a comment explaining this exact trap and why that code stopped racing inside a task group.

If `setupAnalyzer` does finish after cancellation, it has already assigned `liveCaption`, `analyzer` and `recognizerTask` and called `analyzer.start`. The catch then *also* sets up the legacy recognizer, so both are live.

Separately, line 127 caches the failure for the lifetime of the process. One slow first attempt (a cold asset daemon, say) turns SpeechAnalyzer off for that language until the app is relaunched.

Logue/Engine/SpeechTranscriberEngine.swift
```diff
@@ -225,101 +251,318 @@ final class SpeechTranscriberEngine {
...
         logger.info("Locale reserved: \(locale.identifier)")
     }
 }
 
+// MARK: - System speech recognizer (macOS dictation models)
+
+extension SpeechTranscriberEngine {
+    fileprivate func setupLegacyRecognizer(locale: Locale) throws {
+        let recognizer = SFSpeechRecognizer(locale: locale)
+            ?? SFSpeechRecognizer(locale: Locale(identifier: "de-DE"))
```
shanforge: **Major — hardcoded German fallback.** If `SFSpeechRecognizer(locale:)` returns nil for the requested locale, this silently transcribes in **German**, whatever language the user chose. That looks like debugging code left in. It also contradicts this PR's own rule, which `workingLocaleNoEnglishFallbackWhenChosen` tests: never silently switch a chosen language. `SFSpeechRecognizer()` (the Mac's language) also ignores `allowLanguageFallback`.

Fix: remove both fallbacks and throw `localeNotSupported`.

Logue/Engine/SpeechTranscriberEngine.swift
```diff
@@ -225,101 +251,318 @@ final class SpeechTranscriberEngine {
...
-    private func reserveLocale(_ locale: Locale) async throws {
+    fileprivate func reserveLocale(_ locale: Locale) async throws {
         let allocated = await AssetInventory.reservedLocales
         let bcp47 = locale.identifier(.bcp47)
 
         if allocated.contains(where: { $0.identifier(.bcp47) == bcp47 }) {
             return
         }
 
+        if allocated.count >= AssetInventory.maximumReservedLocales, let oldest = allocated.first {
```
shanforge: **Minor.** `AssetInventory.reservedLocales` isn't documented as ordered by age, so `allocated.first` isn't necessarily "the oldest". It could be a locale the user reserved for something else. `readySpeechTranscriber` and `readyDictationTranscriber` also reserve *before* checking `status == .installed`, and `englishCaption` loops over several fallbacks. One failed setup can therefore reserve (and evict) several locales it never uses. Check status first, and reserve only the locale you actually run.

Logue/Engine/SpeechTranscriberEngine.swift
```diff
@@ -75,115 +79,135 @@ final class SpeechTranscriberEngine {
...
+        }
+    }
+
     // MARK: - Stream Audio
 
     /// Feed a raw audio buffer from the microphone or system audio capture.
     /// The buffer is automatically converted to the format expected by SpeechAnalyzer.
     func streamAudio(_ buffer: AVAudioPCMBuffer) {
+        if let legacyRequest {
+            legacyRequest.append(buffer)
```
shanforge: **Question.** The analyzer path converts every buffer through `BufferConverter` to one format. This path appends microphone and system-tap buffers raw into a single `SFSpeechAudioBufferRecognitionRequest`. Have you tested an online meeting with system audio armed on this path? If the two sources arrive in different sample rates or channel counts, the request gets mixed formats.

Logue/Services/RecordingSessionManager+AudioStream.swift
```diff
@@ -11,31 +11,23 @@ import Foundation
...
-
-                guard captured.source == .microphone, let self else {
-                    engine?.streamAudio(captured.buffer)
-                    continue
-                }
-
-                for buffer in await admitToTranscriber(captured.buffer, diarizer: diarizer) {
-                    engine?.streamAudio(buffer)
-                }
+                self?.speechEngine?.streamAudio(captured.buffer)
```
shanforge: **Major — removes the voice-activity gate on a diagnosis that isn't shown.** The gate isn't dead weight. `TranscriptionGate`'s doc comment records a measured improvement: with silence withheld, "The quarterly numbers came in higher" is transcribed correctly, and with it streamed it came out as "that ours came in higher". `admitToTranscriber` already passes buffers straight through when the VAD model is unavailable.

The empty transcripts this PR describes are explained by the other two fixes (unflushed volatile text, and the batch pass having no lines to pour into). There's no test or log here showing the gate itself swallowed speech. Without a demonstrated failure, this is a quality regression for every English user. It also leaves `admitToTranscriber` and `MicrophoneSpeechGate` as dead code that `speechGate.reset()` still maintains.

If the gate is wrong, please fix it with a failing test in `TranscriptionGateTests`, not by bypassing it.

Logue/Services/RecordingSessionManager+Recovery.swift
```diff
@@ -240,4 +240,80 @@ extension RecordingSessionManager {
             speakerSegments: speakerSegments
         )
     }
+
+    /// Fills a meeting that has audio on disk but no transcript — the live captions
+    /// produced nothing, and the post-recording pass then had no lines to pour into.
+    func transcribeSavedRecordingIfEmpty(for meetingID: UUID) async {
```
shanforge: **Major — full ASR + diarization runs when a meeting is opened, and runs again on every open.** This is called from `MeetingWorkspaceView`'s `.task(id: meeting.id)`. So *selecting* any meeting that has an audio file and no segments loads the diarization models, transcribes the whole file, and then starts the post-recording LLM pipeline. The user never asks for any of it.

Failure scenarios:
- A recording that was genuinely silent, or where ASR finds nothing, never gets segments. The full pass runs again every time the meeting is opened.
- Existing meetings from earlier versions that happen to have empty transcripts get re-transcribed and summarized just because the user browsed to them.

Also:
- This is a near line-for-line copy of `reprocessRecoveredAudio` above. The copy drops that function's `savedAudio.describesSessionTimeline` guard, so it can transcribe a file that isn't the meeting's timeline, which is the recording invariant in `CLAUDE.md`.
- `diarizationStage = "Identifying speakers…"` is the wrong label for a transcription pass.

Fix: keep the explicit "Transcribe Recording" menu item and remove the automatic trigger. Route through the shared function instead of a copy.

Logue/Services/RecordingSessionManager+Recovery.swift
```diff
@@ -240,4 +240,80 @@ extension RecordingSessionManager {
...
+    /// if the stored URL no longer resolves (container move, unsigned local build).
+    private func savedAudioURL(for meeting: MeetingNote) -> URL? {
+        if let url = meeting.audioFileURL, FileManager.default.fileExists(atPath: url.path) {
+            return url
+        }
+        let support = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first
+        let recordings = support?
+            .appending(path: AppConstants.bundleID, directoryHint: .isDirectory)
+            .appending(path: "Recordings", directoryHint: .isDirectory)
+        for ext in ["m4a", "caf", "wav"] {
```
shanforge: **Minor — duplicated path logic.** `RecordingSessionManager+AudioFiles.swift` already builds this Application Support → `Recordings` path. Guessing `m4a`/`caf`/`wav` by UUID here is a second, drifting copy of where recordings live. It also re-derives the file afterwards, where `CLAUDE.md` says to trust what `persistRecordingAudio` reported. Reuse the existing helper.

Logue/Views/Meeting/MeetingWorkspaceView.swift
```diff
@@ -159,6 +160,9 @@ struct MeetingWorkspaceView: View {
             }
             .task(id: meeting.id) {
                 await takeAutoRecordRequest(for: meeting)
+                if !recorder.isRecording, store.pendingAutoRecord != meeting.id {
+                    await recorder.transcribeSavedRecordingIfEmpty(for: meeting.id)
```
shanforge: See the comment on `transcribeSavedRecordingIfEmpty`. Opening a meeting shouldn't start a transcription pass plus an LLM summary. Please remove this automatic trigger; the "Transcribe Recording" menu item covers the manual case.

Logue/Models/MeetingNote.swift
```diff
@@ -419,29 +481,153 @@ enum TranscriptionLanguage: String, CaseIterable, Identifiable {
         case .polish: Locale(identifier: "pl-PL")
         }
     }
+
+    /// Language stored on a meeting, or the app default when the meeting never chose one.
+    static func resolved(fromStored raw: String?, defaults: UserDefaults = .standard) -> TranscriptionLanguage {
```
shanforge: **Major — compatibility with existing meetings.** On `main`, meetings are created without `transcriptionLanguage` (only seed data sets it), so almost every existing meeting has `nil` here. Resolving `nil` against the *current* Settings default at read time means changing that default rewrites the past.

Failure scenario: a user has 200 English meetings from 1.1.0 and sets the default to German for a new German client. Regenerating any old meeting's minutes or title now asks the model for German, and `smartMinutesMarkdown()` exports those old English minutes under German headings.

Fix: `nil` should keep meaning what it meant when the meeting was recorded (Auto). Apply the Settings default only when a meeting is created, which `createMeeting` / `createVoiceNote` already do.

Logue/Models/MeetingNote.swift
```diff
@@ -419,29 +481,153 @@ enum TranscriptionLanguage: String, CaseIterable, Identifiable {
...
+            // tag (`de`). Apple's asset catalog names models `transcription.de`
+            // when the region is missing, and that download fails.
+            return match.locale ?? preferredLocale
+        }
+        return fallback
+    }
+
+    /// Locales to try for live ASR, in order: Apple's equivalent, then every
+    /// supported locale of the same language (`de-AT` after `de-DE`).
+    static func speechLocaleCandidates(
```
shanforge: **Minor — test-only production code.** `speechLocaleCandidates`, `isSpeechAssetUnavailable` and `workingLocale` have no production call sites; only `TranscriptionLanguageTests` calls them. The tests look like they cover the "don't silently switch a chosen language" rule, but the engine doesn't use `workingLocale`, and the real path (`setupLegacyRecognizer`) does switch languages (German fallback, line 447). Please remove the unused helpers and their tests, or route the engine through them.

Logue/Models/MeetingNote.swift
```diff
@@ -260,6 +260,40 @@ struct SmartMinutes: Codable {
...
+        duration: "Duration",
+        keyDecisions: "Key Decisions",
+        discussionPoints: "Discussion Points",
+        actionItems: "Action Items",
+        followUps: "Follow-ups",
+        attendees: "Attendees",
+        topics: "Topics"
+    )
+
+    static let german = MinutesMarkdownHeadings(
```
shanforge: **Minor.** Only German gets translated headings. The prompt asks for prose in all 13 languages, so a French or Spanish meeting exports French prose under English headings. Hand-rolling a translation table inside the model file doesn't scale to that. Either use a String Catalog lookup keyed on the meeting's locale, or leave the headings in English for every language.

Logue/Services/RecordingSessionManager+Diarization.swift
```diff
@@ -161,21 +161,32 @@ extension RecordingSessionManager {
         // Adopting them re-cut the transcript the user had been reading. So its words are poured
         // into the live segments instead: every line keeps its identity, its start and its end, and
         // only the text inside improves.
-        if let batchSegments, !diarizer.lastBatchWords.isEmpty,
+        //
+        // When live captions produced no lines, there is nothing to pour into. Keeping the empty
+        // live transcript then discards the batch result — which is how a recorded meeting ended
+        // with 71 ASR segments and zero text. Adopt the batch lines in that case.
+        if let batchSegments, !batchSegments.isEmpty,
```
shanforge: **Minor — stale comments.** The comment block just above still says this path "does not replace the transcript at all, so there is nothing for one to bound" and "The transcript the user watched being written is the transcript they keep". Neither is true once empty live captions adopt the batch lines. Please rewrite the block so it describes the two branches this now has, rather than stacking a third paragraph under two that contradict it.

Logue/Views/Meeting/MeetingWorkspaceView.swift
```diff
@@ -94,6 +94,7 @@ struct MeetingWorkspaceView: View {
                         volatileText: recorder.isRecording ? recorder.volatileText : "",
                         bookmarks: meeting.bookmarks,
                         isLive: recorder.isRecording,
+                        captionsPending: recorder.isRecording && recorder.speechEngine == nil,
```
shanforge: **Minor — layer boundary.** `speechEngine` was made internal with an "Extension-visible: +AudioStream" note. `CLAUDE.md` says extension-visible members must not be used outside the class and its extensions, and this view now reads it. Expose a `captionsPending` / `areCaptionsAttached` flag on the recorder instead.

Logue/Views/Meeting/MeetingWorkspaceView.swift
```diff
@@ -870,10 +923,16 @@ struct RecordingStatusView: View {
             AudioLevelMeter(level: audioLevel)
                 .frame(width: 64, height: 14)
 
+            Image(systemName: isMicActive ? "mic.fill" : "mic.slash.fill")
+                .font(.caption2)
+                .foregroundStyle(isMicActive ? AppThemeConstants.error : .secondary)
```
shanforge: **Minor — scope.** This change reverses a deliberate design decision: the deleted doc comment explained why the capsule has *no* mic glyph (two mic icons an inch apart read as a pair of toggles). It also paints the normal "mic on" state in `AppThemeConstants.error` red, here and on the mute button. The same goes for the new Regenerate button and the "Transcribe Recording" menu item: these UI changes aren't part of #95, and each deserves its own PR with screenshots.

LogueTests/SpeechTranscriberEngineTests.swift
```diff
@@ -0,0 +1,28 @@
...
+
+@Suite("Speech transcriber engine")
+@MainActor
+struct SpeechTranscriberEngineTests {
+    @Test("Finish keeps leftover live text as a final segment")
+    func finishFlushesVolatileText() async {
+        let engine = SpeechTranscriberEngine()
+        var received: [String] = []
+        engine.onFinalSegment = { received.append($0.text) }
+        engine.volatileText = "hello from the meeting"
```
shanforge: **Tests.** These only cover the volatile flush, by setting `volatileText` by hand. None of the risky new behaviour is tested:
- starting captions after the mic (timestamps relative to the file)
- Stop pressed during setup
- the legacy restart and error paths
- the timeout
- `refreshLinkedSummaryDocument` against an appended-to document
- `transcribeSavedRecordingIfEmpty` not re-running on a silent recording
- `resolved(fromStored: nil)` for pre-existing meetings

The first and the fourth-from-last are the regressions that matter most. Please add tests that fail on this head.

shanforge (changes requested):
## Review: request changes

Thanks for chasing the German captions problem. Two of the fixes are good and worth keeping: flushing the leftover volatile text when a session finishes, and adopting the batch lines when live captions produced nothing. The rest of the PR goes well beyond #95, though, and several parts regress users on `main`, including English users who never touch the new language picker.

### Blocking (4)
1. **Meeting audio can go to Apple's servers.** The `SFSpeechRecognizer` fallback only requires on-device recognition when it's supported (`SpeechTranscriberEngine.swift:469`).
2. **Users' documents get overwritten.** `refreshLinkedSummaryDocument` replaces the whole body of a document the user *appended* a summary to. It runs automatically after every recording (`MeetingStore+AI.swift:220`).
3. **Caption timing is wrong, and Stop can leak the engine.** Captions attach after capture starts: live timestamps shift early by the setup time, the first seconds are never captioned, and batch words are realigned into the wrong lines. Pressing Stop during setup leaks a recognizer that keeps running (`RecordingSessionManager.swift:418`).
4. **The speech model is no longer downloaded.** A fresh install gets no SpeechTranscriber asset and falls to the legacy recognizer (`SpeechTranscriberEngine.swift:367`).

### Major (6)
- Restart loop with no backoff and no authorization check.
- The setup timeout can't fire, because a task group waits for its children; the failure is also cached for the whole process.
- Hardcoded `de-DE` fallback.
- The measured voice-activity gate is removed without a demonstrated failure.
- Full ASR, diarization and an LLM summary run automatically just from *opening* a meeting, and again on every open.
- Existing meetings (`transcriptionLanguage == nil`) take on whatever the Settings default is now.

Details are inline, next to the code.

### Compatibility with `main`
- Every existing meeting has `transcriptionLanguage == nil`, so changing the new default changes their regenerated minutes, titles and export headings.
- Linked and appended-to documents (including `.md` files in plain-markdown storage) are rewritten without asking.
- English users lose the model download and the voice-activity gate.

Persisted `Codable` shapes are unchanged, so old data still decodes.

### Scope
This PR bundles #95 with an unrelated mic-indicator redesign, a Regenerate button, a Transcribe Recording command and a new fallback transcription engine. Please split it: the language picker plus prompt language as one PR, the recording-pipeline changes as another, the UI tweaks as a third. The recording path in particular needs to be reviewable on its own.

### CI
**No checks have run on this head** (`6d02080`). This is a fork PR, so the Build and Lint Check workflows are waiting for maintainer approval. The PR can't be approved without them.

### Evidence
The PR has no screenshots or screen recordings. Once the above is addressed, please attach:
- **Settings → General → Meetings** showing the "Default recording language" picker, plus the Start control's language menu open on a meeting.
- **A German meeting, 2 min:** record with the mic, show live captions appearing within the first 5 s, stop, and show the saved transcript's first line starting at about 0:00 (not shifted) when you click it to play back.
- **Stop within 3 s of Start** on a language with no installed asset, then start a new recording: Activity Monitor or the log should show only one recognizer running.
- **An existing 1.1.0 meeting** after changing the default to German: Regenerate should still produce English minutes.
- **A document** you appended a summary to, before and after a second recording on that meeting: your own text must still be there.
- **Airplane mode** (no network) with German selected and no on-device dictation model: captions should fail with the notice, not transcribe.

